PR #5, opened 2022-12-13
Adding packaging files and added pycache to gitignore

gareth-j:
I've added some packaging files so the repo can be installed using `pip install -e .`.
I've used `flit` to create the `pyproject.toml` so that should work OK. Hopefully the dependencies are correct.
Also the data folder might need adding to the manifest or toml file.


gareth-j:
Closed this as work done separately.

Issue #18, opened 2023-07-04
Post-processing output sometimes fails due to "Filter error"


brendan-m-murphy:
Sometimes the "Post-processing output" in `inversions_pymc.inferpymc_postprocessouts` fails due to a "Filter error":

> Post-processing output
Traceback (most recent call last):
  File "/user/home/bm13805/openghg_inversions/openghg_inversions/hbmcmc/run_hbmcmc.py", line 208, in <module>
    mcmc_function(**param)
  File "/user/home/bm13805/openghg_inversions/openghg_inversions/hbmcmc/hbmcmc.py", line 413, in fixedbasisMCMC
    mcmc.inferpymc_postprocessouts(xouts,bcouts, sigouts, convergence,
  File "/user/home/bm13805/openghg_inversions/openghg_inversions/hbmcmc/inversion_pymc.py", line 604, in inferpymc_postprocessouts
    outds.to_netcdf(output_filename, encoding=encoding, mode="w")
  File "/user/home/bm13805/.venv/lib/python3.10/site-packages/xarray/core/dataset.py", line 1946, in to_netcdf
    return to_netcdf(  # type: ignore  # mypy cannot resolve the overloads:(
  File "/user/home/bm13805/.venv/lib/python3.10/site-packages/xarray/backends/api.py", line 1272, in to_netcdf
    dump_to_store(
  File "/user/home/bm13805/.venv/lib/python3.10/site-packages/xarray/backends/api.py", line 1319, in dump_to_store
    store.store(variables, attrs, check_encoding, writer, unlimited_dims=unlimited_dims)
  File "/user/home/bm13805/.venv/lib/python3.10/site-packages/xarray/backends/common.py", line 278, in store
    self.set_variables(
  File "/user/home/bm13805/.venv/lib/python3.10/site-packages/xarray/backends/common.py", line 316, in set_variables
    target, source = self.prepare_variable(
  File "/user/home/bm13805/.venv/lib/python3.10/site-packages/xarray/backends/netCDF4_.py", line 495, in prepare_variable
    nc4_var = self.ds.createVariable(
  File "src/netCDF4/_netCDF4.pyx", line 2962, in netCDF4._netCDF4.Dataset.createVariable
  File "src/netCDF4/_netCDF4.pyx", line 4202, in netCDF4._netCDF4.Variable.__init__
  File "src/netCDF4/_netCDF4.pyx", line 2029, in netCDF4._netCDF4._ensure_nc_success
[...]

PR #19, opened 2023-07-05
Variable length data is no longer compressed

brendan-m-murphy:
Compressing variable length data when writing a dataset to netcdf can cause a "Filter Error".

To avoid this, I modified the encoding in `infterpymc_postprocessouts` to not apply `zlib` to data variables whose data type is variable length.

These data types are: "Object" and "Unicode". These are numpy datatypes with codes "O" and "U". I used a regex to parse the datatype string for each data variables, since these strings can start with a symbol to specify the byte-order.

This elimiates the FilterError from the inversion run on Iss18, and the netCDF output seems to be about 1/3 the size of the output if `encoding = None` (i.e. no additional compression from `zlib=True`) is used.

This closes Issue #18.


gareth-j (approved):
Looks good, nice method.

PR #20, opened 2023-08-15
Revert "Separate data merge function"

brendan-m-murphy:
Some of the changes in this merge mean the old .ini files don't work. After I add some tests to the project, I'll change the reverted code to accept the old .ini files as well as the new ones (if possible), and create a new PR.


aliceramsden (approved):
I'm happy to have these changes reverted, then they can be re-added once the tests have been updated.

Issue #22, opened 2023-08-15
Modify PR "separate merge function" to use old and new .ini format


brendan-m-murphy:
We reverted this PR temporarily to add some tests in: https://github.com/openghg/openghg_inversions/pull/16

Once the tests have been added, we should try to get the code from PR 16 to work with old .ini files, then re-raise the PR.

brendan-m-murphy:
The "separate merge functionality" code has been added to `eric_devel`, hence to `min_model_error`, so this issue can be closed.

PR #24, opened 2023-08-16
Adding multi gas model

aliceramsden:
### Additions:
This adds the multiple gas inverse model, including its own input data merge function, MCMC sampler and output processing functions. There is the option to run this using netCDF inputs (using the ACRG data merge functions) or an object store (with OpenGHG functions). I did test this a while ago using both methods with some basic CH4 inputs and this worked well, although haven't had the chance to do this again more recently.

The `multi_gas_model` directory contains a README and a PDF containing lots of detail behind the MCMC process for the multiple gas model including the equations for modelling mole fraction and isotope delta values. These documents also explains how to run the multiple gas model and all the available options for inputs. These may need updating now this code has been moved into this repository. I could potentially do this in a few weeks/months time if needed.

### Changes
I have pulled out `run_hbmcmc.py` from the hbmcmc directory and created an over-arching `run_inverse_model.py` which takes the model name and a .ini file as inputs on the command line. There's options for either `hbmcmc` or `multi_gas` and their corresponding .ini files, which have different inputs depending on the model type. There's currently no way round having different .ini formats, as the model inputs are so different. But templates for each type are included in each model's directory.

There's also some updates to how new quadtree basis function files are named, so they can be labelled with a date then searched for more easily during any model reruns.

### Notes:
I think this requires PR #22 (which was linked to #16) to be merged in first, so that the separate data merge function can be used by the multiple gas model.

There's some functionality missing from the multiple gas model (most importantly adding some tests!) so I'll add these things to the issues list.

Also, reviewing this will be a huge job, so I'm just assigning anyone who's worked on openghg_inversions for now, and I'm definitely not expecting this to be sorted any time soon!


aliceramsden:
Code in this branch will no longer work with the current setup of openghg and openghg_inversions, so I'll work on updating this later this year.

brendan-m-murphy:
Okay, sorry I never got to this. It might be worth adding in the option to run different models as a separate PR, and possibly try to use PyMC to do the sampling (although well done writing your own sampler).

Issue #8, opened 2023-03-01
Add some basic tests


gareth-j:

gareth-j:
I added some very simple tests for `load_json` with #9

brendan-m-murphy:
I'm going to add a test that runs a full inversion. Hopefully I can make this reasonably fast.

So far, the things I can think of testing with a full inversion are:
1. the inversion runs without errors
2. by storing the result of the inversion with the current set up (and fixing random seeds), checking if the results are equal (or very close to) the stored result. Since assessing if the results of an inversion are "correct" is pretty difficult, I want to be able to test if changes are altering the way the inversion works. (For initial tidying of the code, I don't want to change the actual results.)

If anyone has any other testable ideas, let me know. These first tests are just for checking that changes don't break the code. As I tidy things up, I should be able to test individual components.

I will probably just use CH4 data and one site, but I could add more types of data. It would be nice to create synthetic data, but I haven't thought about what that would look like.

Later, I'll make tests for performance. I think running pytest on Blue Pebble should work, so I should be able to run "profiling" tests there. Maybe it would be easier to just write some scripts for performance testing, since they need to be run on BP (or Blue Crystal). Anyway, the current goal is to get enough tests to be able to change the code without worrying that it's breaking.

gareth-j:
Done with #25

PR #25, opened 2023-09-15
Added full inversion test with all required data

brendan-m-murphy:
This is a test that runs `fixedbasisMCMC` for one iteration with 4 basis functions, over two days of data. It runs in about 20 seconds, so it should be useful for checking that minor changes to the code don't break existing functionality.

This test doesn't cover the `run_hbmcmc` script. The changes to `hbmcmc.py` are mostly due to black; there are a few changes to make it easier to choose where the quadtree basis is saved.

I added a test with more iterations, which is marked "slow". I set up pytest.ini to not run "slow" tests by default. The slow test uses a week of data, 50 basis functions, and 5000 iterations; it runs in about 7-8 minutes on my macbook.

In the future, I may add more "slow" tests that are marked "blue_pebble", which will only run on blue pebble, for tuning purposes. (Alternatively, this could be done with a separate script.)

For set-up, `conftest.py` creates an object in a temporary directory using some data files distributed with `openghg_inversions`. If this object store already exists and has data in it, then no data is standardised, which should speed up subsequent tests. (This is also a `test_conftest.py` file that checks if the right data is available.) Some data needed to run the inversion test isn't stored in the object store; this data is added to "bc_basis_functions" and "countries" by conftest when the tests are first run (if they don't exist, conftest creates these directories.)

To keep the `openghg_inversions` repo clean while running tests, I changed how `quadtreebasis` works: if "outputdir" is passed, then the basis is saved there, rather than in a temporary directory (which with the current code, would be created where the tests are run from). I also added some safe guards to the part of `hbmcmc.py` that deletes the temporary basis file. Saving the quadtree to the output folder is controlled by a flag in `fixedbasisMCMC`, so it doesn't change the behaviour of the code unless you ask it to.

Closes issue #8.


gareth-j (approved):
Looks good!

PR #39, opened 2024-01-25
Updates to min_model_error, to bring it up to date with changes in eric_devel etc.

aliceramsden:
This branch can be merged into min_model_error when everyone else is ready to start turning min_model_error into the new 'main' branch of the repo.

Updates on this branch include:

- A fix for reading in the correct prior fluxes, when creating the posterior country fluxes and saving everything after the inversion. The prior fluxes are now read directly from the merged data object, and the correct monthly/annual flux is sliced from the full flux object. This includes taking an average flux across a range of months, if the inversion is across mulitple months.

- A try/except loop which drops sites from the inversion if the data merge process doesn't work for that site (which normally happens if there's no obs).

- Adding in a print out of the number and % of obs that are removed by each filter, at each site.

 - Fixes for saving and reading in the merged data object, including modifying the site variable (and associated heights etc.) if these aren't found in the merged data object.

- Some minor bug fixes, including some in the basis function creation process and some variable names.

### DIFF
--- a/openghg_inversions/hbmcmc/inversion_pymc.py
+++ b/openghg_inversions/hbmcmc/inversion_pymc.py
@@ -19,7 +19,6 @@ from pathlib import Path
 from typing import Optional, Union
 
 from openghg.dataobjects import FluxData
-from openghg.retrieve import get_flux
 
 from openghg_inversions import convert
 from openghg_inversions import utils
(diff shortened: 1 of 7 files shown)


brendan-m-murphy:
Hi Alice, thanks for this, I'll work on reviewing it and merging into `min_model_error` tomorrow.

openghg_inversions/hbmcmc/inversion_pymc.py
```diff
@@ -523,27 +522,30 @@ def inferpymc_postprocessouts(
...
-        # Note, at the moment fluxapriori in the output is the mean apriori flux over the
-        # inversion period and so will not be identical to the original a priori flux, if
-        # it varies over the inversion period
-        emissions_flux = np.expand_dims(rerun_file.fluxapriori.values, 2)
-    elif flux:
-        emissions_flux = flux.data.flux.values
+        flux_array_all = np.expand_dims(rerun_file.fluxapriori.values, 2)
     else:
         if emissions_name == None:
             raise NameError("Emissions name not provided.")
```
brendan-m-murphy, line 528:
```suggestion
            raise ValueError("Emissions name not provided.")
```
A `NameError` is for when a variable name is used, but that variable hasn't been defined. We should use `ValueError` here because the value `None` is causing the error.

openghg_inversions/hbmcmc/inversion_pymc.py
```diff
@@ -523,27 +522,30 @@ def inferpymc_postprocessouts(
...
     if rerun_file is not None:
-        # Note, at the moment fluxapriori in the output is the mean apriori flux over the
-        # inversion period and so will not be identical to the original a priori flux, if
-        # it varies over the inversion period
-        emissions_flux = np.expand_dims(rerun_file.fluxapriori.values, 2)
-    elif flux:
-        emissions_flux = flux.data.flux.values
+        flux_array_all = np.expand_dims(rerun_file.fluxapriori.values, 2)
     else:
         if emissions_name == None:
```
brendan-m-murphy, line 527:
```suggestion
        if emissions_name is None:
```
`is None` is preferred to `== None` (I guess similar to comparing NaNs.)

brendan-m-murphy (changes requested):
Just a few changes. They're all suggested inline, except to use `except SearchError` you also need to add the import statement as described.

There are some things that the formatter picked up, or that should be separate issues (e.g. `get_data` should raise an error if `emissions_name` is `None`, since an error will happen in that case, but it might not be a very helpful error). So if anyone else is reviewing, keep in mind that the branch this PR is merging into still needs work.

PR #44, opened 2024-01-26
Updated CI and requirements.txt

brendan-m-murphy:
CI now installs requirements.txt so changes there will be used in CI.


gareth-j (approved):
Looks good, hopefully `pip install -e .` won't update OpenGHG to 0.7.0

PR #47, opened 2024-01-26
Fixed error in `get_country` class

brendan-m-murphy:
There is some general tidying in this commit, but the main change is line 188 in the new file, which converts the numpy array to strings.


gareth-j (approved):
Looks good.

PR #51, opened 2024-01-30
get_data Error handling more fine grained

brendan-m-murphy:
Added more try/except clauses to catch different possibilities from `get_obs_surface`, as well as printing more info about why a site was skipped.

### DIFF
--- a/openghg_inversions/get_data.py
+++ b/openghg_inversions/get_data.py
@@ -179,7 +179,7 @@ def data_processing_surface_notracer(
             )
         except SearchError:
             print(
-                f"\nNo footprint data found for {site} with inlet/height {fp_height[i]} and model {fp_model}.",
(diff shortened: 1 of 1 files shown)


EricSaboya (approved):
These look good! Happy to merge it in unless you can think of any additional exceptions that are needed for the observations block. 

Would it be helpful to include something in the footprints `except SearchError` specifying the model domain? Only thinking long-term when we'll be using the 6 km and 25 km EUROPE footprints? We might also need to include something in `get_fps` to specify when retrieving the HiTRes footprints for CO2 vs. short-lived vs. inert. I can always include this in when I merge the CO2 stuff

EricSaboya:
Happy to merge this PR, but please see review comment above

brendan-m-murphy:
I added the domain to the footprint error message. Otherwise everything is the same.

EricSaboya (approved):

PR #53, opened 2024-01-31
Release 0.0.3

brendan-m-murphy:
This PR contains some tidying and admin before release 0.0.3.

Note: after this is merged, we should tag the main branch with `v0.0.3`

Further note: there are some minor changed to `fixedbasisMCMC` because I realized that if `reload_merged` was true and the merged data couldn't be found, there would be an error from trying to find out what sites were dropped from `fp_all` (when `fp_all` hasn't been loaded yet).


gareth-j (approved):
Looks good!

PR #57, opened 2024-02-05
Update RHIME input configuration files

EricSaboya:
# Tidied up and updated the RHIME input files
PR to merge in updates made to the input files so they are more accessible and descriptive for new users of RHIME

## Example input configuration file
- Tidied up and updated file variable descriptions in the comments
- Changed data paths so they are not directed to e.g. wz22079 user directory 
- Changed example pdfs for hyper-parameters. Using Lognormal distributions in the example file is confusing. 
- Reduced no. of MCMC iterations and basis functions. Example input file is really to check the inversion runs and not designed for "science"

## Template configuration file
- Tidied up and updated file variable descriptions in the comments. Matches with the example file
- Formatted file 
- Changed example PDFs as above 
- Made sure all variable options available (as per previous `main' input file)


pearson01 (approved):
Looks good to me!

Issue #52, opened 2024-01-30
Test `get_data` and running inversions separately


brendan-m-murphy:
Now that there is a `get_data` function, we can test `get_data` vs. a fixed merged data file, and then use fixed merged data for the inversions tests.

The basis function wrapper could also be tested separately, and possible the post processing code.

pearson01:
Getting an error when trying to run an inversion with the example ini file. The `get_data` function can't seem to find obs data (TAC, ch4) which is preventing the `combined_scenario` from being created. The error message is:

`UnboundLocalError: local variable 'scenario_combined' referenced before assignment`

@brendan-m-murphy

EricSaboya:
The error @pearson01 found should now be fixed in the `update get obs` branch

brendan-m-murphy:
I'm guessing the fix was the part that throws an error if all of the sites are dropped?

I'll make sure to add some test cases where some/all data can't be found.

EricSaboya:
Exactly! I just added an if statement to exit the process if no site data are found before it gets to the line with `scenario_combined.mf.units`, which was what was raising this error

brendan-m-murphy:
This was closed by PR #73

PR #58, opened 2024-02-06
Get obs update

EricSaboya:
The following updates that primarily relate to get_obs.py have been made: 

**get_obs.py**
- input option for calibration scale (update propagates to `hbmcmc.py` and `.ini` files)
- input option for measurement data quality (update propagates to `hbmcmc.py` and `.ini` files)
- block for creating lists of `None` for args that specify just `None` (helps with for loop cycling)
- Added species option for retrieving footprints so CO2 footprints are specified 
- Streamlined the `model_scenario` part of the code so only 1 model scenario object is created instead of N
- added option for calculating `mf_mod` for different fluxes with specification of whether it is `mf_mod` or `mf_mod_high_res`
- Fixes issue #52 from Stephen P. if no site data found process now exits 

**hbmcmc.py**
- Changed default prior PDFs to something more sensible 
- added new args and added documentation to function 
- Tidied up docs and comments in file 

**inversion_pymc**
- Inversion does not run with `save_trace` block. Had to comment out in order for example .ini to run 

**ini files**
- updated example files so they actually run on BP 
- added new variables 
- tidied up commenting etc. 

NB. when using `fp_model=NAME` as an arg, fps are not retrieved. Changed default to None.

### DIFF
--- a/openghg_inversions/get_data.py
+++ b/openghg_inversions/get_data.py
@@ -36,7 +36,7 @@ def data_processing_surface_notracer(
     instrument=None,
     calibration_scale=None,
     met_model=None,
-    fp_model="NAME",
+    fp_model=None,
     fp_height=None,
     emissions_name=None,
     bc_input=None,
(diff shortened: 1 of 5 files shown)


aliceramsden (approved):
I've run some very short inversions using the inputs in the example .ini files and these are working well. Apart from that the 'localness' filter doesn't exist. For now, 'localness' could be changed to 'pblh' in the example .ini file as this filter seems to work (although probably needs more testing).

I've also used this branch to run a short inversion for a month with sites with some missing data and this is working well too, and still skipping these sites as expected.

Issue #45, opened 2024-01-26
Replace `pm.math` with pytensor.tensor`, use arviz instead of pymc for hdi


brendan-m-murphy:
PyMC seems to keep backwards compatibility with older versions, but e.g. `pm.math.sqrt` should be `pytensor.tensor.sqrt` and `hdi` should be calculated using `arviz`.

brendan-m-murphy:
Closed by PR #72, thanks @pearson01 !

PR #70, opened 2024-02-09
Iss45

pearson01:
Within openghg_inversions, pm.math has been replaced with pytensor.tensor and pm.stats has been replaced with arviz. This is to close issue #45


brendan-m-murphy:
This was replaced by another PR

PR #72, opened 2024-02-09
Adding pytensor.tensor and arviz to inversion_pymc.py

pearson01:
Same as PR #70 , but with resolved conflicts with ini files.


brendan-m-murphy (approved):
Looks magnificent to me!

Issue #62, opened 2024-02-08
Pass arguments in memory rather than using temporary files


brendan-m-murphy:
If no basis file is specified, we save the basis to a temporary file and then use that file in `fp_sensitivity` etc.

It would be better to treat loading a basis from file as one way of creating a basis array. Once we have a basis as a np array or xr DataArray, we should pass that as an argument to `fp_sensitivity` etc.

Issue #55, opened 2024-02-02
Implement "fixed outer regions" for basis functions


brendan-m-murphy:
To compare with other models, we'd like to be able to fix the basis functions along the border of the EUROPE domain. This means that the basis functions in this outer region can be made the same for each run.

brendan-m-murphy:
Closed by PR #87

PR #82, opened 2024-02-13
Iss62 refactor basis temp files

brendan-m-murphy:
This PR closes Issue #62 by removing the use of temporary files to pass basis functions from e.g. `quadtreebasisfunctions` to `fp_sensitivity`. Instead, the basis functions are returned as a dataset from `quadtreebasisfunctions` and then passed as an argument to `fp_sensitivity` and `bc_sensitivity`.

This change will make it easier to implement "fixed outer regions" for basis functions (Issue #55), because I can call the existing basis functions on the "inner regions".

I did some tidying and added a test to check that a quadtree basis function created on the fly is the same as one previously created with the same parameters and saved to netCDF. This should alert us to changes in the quadtree basis function.

Note: I'm not sure I've made the best choices of "basis case" vs. "basis function" vs. "basis algorithm".


EricSaboya (approved):
Looks good to me!

PR #85, opened 2024-02-14
Running inversions with to without boundary conditions

EricSaboya:
# Updated RHIME to run with or without boundary condition mole fractions
Main changes include an additional input argument "use_bc" set as True by default. Inversions ran without the BCs are done so from "inversion_pymc_nobc.py" (new file) which is near-identical to "inversion_pymc.py" except without the options to solve for BCs. These functions can always be merged into "inversion_pymc" at some later point if preferred. 

## Assumptions
1. The observations data specified in the input file are the "obs - baseline" mole fractions
2. The observations data follow the same format as "standard" obs data in openghg object stores 

## Changes 
1. Added "use_bc  = <bool>" Input argument to .ini file and elsewhere. 
2. Changed off_trace to OFFSETtrace to be consistent with similar variables. 

**get_data.py**
Added various if statement blocks to get or to not get BC data, (*that is the question*)

**hbmcmc.py**
Basis wrapper function has an if statement for calculating bc sensitivities 

If statement blocks to run with BCs --> inversion_pymc. Or without BCs --> inversion_pymc_nobc.

## Testing
Ran inversion using the five sites with HFC-134a data minus the InTEM baseline for January 2020. Inversion was ran with a small no. of iterations and failed the GR test but successfully produced outputs and sensible looking posterior mole fractions data. 

! IMPORTANT NOTE !
> The observations without baselines data in 'paris_baseline_store' have their uncertainties mislabelled and WILL NOT be included in the inversions as such. Using these data at present will assume zero uncertainty on the mole fractions data.


brendan-m-murphy (approved):
Looks good to me!

PR #86, opened 2024-02-15
co2_updates_eric_testing now running through correctly

aliceramsden:
A few minor changes required to get `co2_updates_eric` to run an inversion all the way through:

- Split up xtrace and Hx in the output dataset, because each sector can now have different numbers of basis functions, so Hx and xtrace are different shapes. I split these up using the size of the xout variable.
- Switched `np.concatenate` to more a basic sum, when adding together modelled `H.x`, as this line was failing when only using one sector. This works for single or multi sector runs now.
- Changed path for land/sea definition file to shared area.
- Other minor fixes for the output dataset creation.

Small checks made:

- I ran a couple of my old single-sector PARIS inversions using this branch, and the results were almost identical to those from the currently working main branch. So updates here shouldn't have impacted how the single-sector inversion works.

### DIFF
--- a/openghg_inversions/basis_functions.py
+++ b/openghg_inversions/basis_functions.py
@@ -261,7 +261,7 @@ def load_landsea_indices():
     land = 1
     sea = 0
     """
-    landsea_indices = xr.open_dataset("/user/home/wz22079/my_openghg/openghg_inversions/countries/country-EUROPE-UKMO-landsea-2023.nc")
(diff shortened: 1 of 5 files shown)


aliceramsden:
I've just realised I left most of the updates lines of code in, just commented out, in case there was anything I wanted to double check later. I can add another commit to this to remove those.

EricSaboya (approved):
These look great and the updates seem like a more sensible way of splitting Hx. Don't worry too much about leaving the commented code in - I do that too! We can always take them out in a separate PR

PR #87, opened 2024-02-19
Iss55 fix outer regions

brendan-m-murphy:
Added option to fix outer regions when computing basis functions.


hdelongueville (approved):
These changes sound good to me!

Issue #36, opened 2024-01-24
Review `hbmcmc.hbmcmc_post_process` (do we use it?)


brendan-m-murphy:
This code seems to be copied over from tdmcmc on the ACRG repo. I'm guess it isn't used anywhere presently, because there are import statements that would raise error if the module was imported.

If we're not using this, we should probably remove it. (Could tag it beforehand so it is easy to find the commit that has it in the future.)

PR #88, opened 2024-02-22
Update hbmcmc_post_process.py

qq23840:
C.F. Issue #36 about `hbmcmc_post_process.py`

The functions within `hbmcmc_post_process.py` are, I think, still quite useful, especially for someone still learning inversions and trying to quickly compare outputs. I've just fixed a few bugs and updated the documentation so that these can be used on the hbmcmc output datasets currently generated by `main` (I'm aware these may change soon).

Changes: 
-Changed how the `site_info.json` and `species_info.json` data are called to remove dependency on ACRG paths
-Fixed various functions so that they call the fluxmode variable rather than fluxmean (which doesn't exist)
-Thorough update of documentation for all functions
-Fixed bug in `set_cmap` where a very sparse dataset (i.e. >98% NaN) would not generate a valid array of values for the colormap
-General updates
-Also added a line in `basis_functions.py` to deal with NaNs in flux data which otherwise derail the quadtree algorithm (although this should be addressed by changes to the edgar parser in openghg, and probably in the new basis functions setup as well)
-Did NOT manage to tidy up the `calculate_DIC` function - I'm not quite sure how it works and I think a lot has changed since it was first written

Checks:
-Tests don't pass, but I think that's on my end? Can look into this if needed. I don't think any of the functionality I've changed/added are tested for, however
-Ran black on `hbmcmc_post_process.py`


openghg_inversions/hbmcmc/hbmcmc_post_process.py
```diff
@@ -789,7 +798,7 @@ def plot_abs_map(
         )
         for ds in ds_list
     ]
-    q_abs_list = [convert.mol2g(ds.fluxmean, species) for ds in ds_list]
+    q_abs_list = [convert.mol2g(ds.fluxmode, species) for ds in ds_list]
```
brendan-m-murphy, line 801: maybe `fluxmode` should be a parameter (seem comment below)

brendan-m-murphy (changes requested):
Looks good, definitely a big improvement! In the two places where you changed `fluxmean` to `fluxmode`, I think it makes sense to make this a parameter with default value `fluxmode`. I made a remark about the `Paths` class, but maybe just ignore that for now because it might impact other parts of the codebase.

qq23840:
fixed issues as highlighted by @brendan-m-murphy. Tests still failing but I think that's not an issue with these changes at all

brendan-m-murphy:
> fixed issues as highlighted by @brendan-m-murphy. Tests still failing but I think that's not an issue with these changes at all

The failing tests are due to some incompatible dependencies. I've fixed this on PR #87, so once that's merged into devel, merging devel into this branch should make the tests run. (But there aren't any tests related to the post_process module, so it doesn't really matter).

qq23840:
Great - I'll wait until that PR goes through, then merge into this branch

qq23840:
Sorted

brendan-m-murphy (approved):
Looks good to me!

Issue #89, opened 2024-02-28
Add tests for multiple versions of OpenGHG


gareth-j:
We should run the tests with multiple different versions of OpenGHG installed, maybe
- `devel`
- `current_tag`
- `previous_tag`

Issue #43, opened 2024-01-26
Upgrade to OpenGHG 0.8


brendan-m-murphy:
Currently 0.7 is causing the tests to fail (errors in the `conftest.py` during test set up), so those will need to be fixed.

OpenGHG 0.7 also upgrades from pandas 1.53 to latest version of pandas (2.x), so we need to check if inversions still works with the latest version of pandas.

rt17603:
@brendan-m-murphy Should we update this to Upgrade to v0.8.0 now or do we want that as a separate issue?

brendan-m-murphy:
Closed by PR #92

Issue #38, opened 2024-01-24
Replaced pickled merged data with netCDF


brendan-m-murphy:
Once the changes to `fp_all` in the co2 branch (https://github.com/openghg/openghg_inversions/tree/pymc_changes_for_co2) are completed, we should make `fp_all` an xarray `Dataset` rather than a dictionary, and convert the save/load merged data functions to save this `Dataset` to netCDF.

`.pickle` files aren't good for long term storage because changes in dependencies can make it impossible to use them (e.g. merged data files pickled using pandas 1.5.3 can't be opened in an environment using pandas 2.0)

brendan-m-murphy:
Closed by PR #92

PR #92, opened 2024-03-04
Updates to use Openghg v0.8 (zarr backend)

brendan-m-murphy:
This PR includes updates to make OpenGHG inversions compatible with OpenGHG versions 0.7.1 and 0.8 (using zarr backend). This closes issue #43.

Besides changes to the tests and a few minor changes, the way "merged data" is saved and loaded has been changed significantly. By default, if zarr is available, then merged data will be saved to a zarr store; otherwise, the merged data will be saved as a netCDF file.

Merged data can be loaded from zarr, netCDF, or a pickle file. The function `load_merged_data` in `get_data.py` searched for merged data in these formats, in this order, and uses the first compatible option found. (So if merged data is saved under the same name with ".zarr" and ".nc" extension, then ".zarr" will be used if zarr is available, otherwise the netCDF file will be used.)

There isn't an option to specify which format is used, but this could be added later.

This closes issue #38.

Further, the continuous integration (CI) workflow has been updated to use three version of OpenGHG: v0.7.1, v0.8, and the devel branch. This will help ensure some backwards compatibility, as well as warn us about conflicts with the devel branch of OpenGHG. (Currently the versions v0.7.1 and v0.8 are hard-coded, so these will need to be updated as new versions of OpenGHG are released.)

This closes issue #89.


openghg_inversions/get_data.py
```diff
@@ -283,8 +289,8 @@ def data_processing_surface_notracer(
 
                 scenario_combined = model_scenario.footprints_data_merge(recalculate=True)
 
-                for key in model_scenario_dict.keys():
-                    scenario_combined[key] = model_scenario_dict[key]
+                for k, v in model_scenario_dict.itmes():
```
gareth-j, line 292:
```suggestion
                for k, v in model_scenario_dict.items():
```

openghg_inversions/get_data.py
```diff
@@ -352,9 +358,350 @@ def data_processing_surface_notracer(
...
+    Alternatively, "pickle" can be specified.
+
+    Args:
+        fp_all: dictionary of merged data to save
+        merged_data_dir: path to directory where merged data will be saved
+        species: species of inversion
+        start_date: start date of inversion period
+        output_name: output name parameter used for inversion run
+        merged_data_name: name to use for saved data.
+
```
gareth-j, line 400: Missing `output_format` here

openghg_inversions/get_data.py
```diff
@@ -352,9 +358,350 @@ def data_processing_surface_notracer(
...
+    This function tries to automatically find a compatible format of merged data.
+    First, it checks for data in "zarr" format, then in netCDF, and finally in pickle.
+
+    Args:
+        merged_data_dir: path to directory where merged data will be saved
+        species: species of inversion
+        start_date: start date of inversion period
+        output_name: output name parameter used for inversion run
+        merged_data_name: name to use for saved data.
+
```
gareth-j, line 458: Missing `output_format` here

gareth-j (changes requested):
I didn't go over everything but just found a couple of little things.

gareth-j (approved):
Looks good!

qq23840:
Currently running a GSN-HFC-23 inversion with this branch. Have had to make a couple of tweaks locally to deal with a time-slicing issue in openghg itself that @gareth-j and I discussed yesterday (https://github.com/openghg/openghg/issues/970) but outside that, the inversion runs fine so all looks good here

PR #94, opened 2024-03-05
basis functions read in from fp_basis_case now work correctly again

aliceramsden:
The option to use a fixed basis function (specified by the `fp_basis_case` variable) wasn't working in the previous commit, because the code now looks for a `sector` coordinate in the basis function file. 

I've solved this issue (in a slightly fudge-y fix) by adding the `sector` coordinate to the simple basis function file. Then there is a matching basis function grid available for all sectors.

Also, changed the combine datasets method from `ffill` to `nearest` as `ffill` was adding nans at the start of the combined datasets.

I'll rename/move the co2_updates eric branch, once this merge is complete.


EricSaboya (approved):
Good spot with the fixed basis functions and I think the fix looks good for what we want now. Can always tweak it later. All looks good to me!

PR #98, opened 2024-04-10
Correction to prevent freezing when reindexing xarray

alexdanjou:
Correction to prevent freezing when reindexing xarray containing basis functions definition.

Typo corrections.


brendan-m-murphy:
I think this isn't necessary because of the fix in PR #87

Issue #99, opened 2024-04-11
specify the directory of country-EUROPE-UKMO-landsea-2023.nc in _weighted.py


hdelongueville:
The **line 14** in [openghg_inversions/openghg_inversions/basis/algorithms/**_weighted.py**, branch Iss55-fix-outer-regions] should specify the whole path of country-EUROPE-UKMO-landsea-2023.nc.

For now it is:

landsea_indices = xr.open_dataset("../countries/country-EUROPE-UKMO-landsea-2023.nc")

and should be changed to:

landsea_indices = xr.open_dataset("/group/chem/acrg/LPDM/countries/country-EUROPE-UKMO-landsea-2023.nc")

to avoid error like, for example,

FileNotFoundError: [Errno 2] No such file or directory: '/user/home/qq24644/countries/country-EUROPE-UKMO-landsea-2023.nc')

brendan-m-murphy:
Sorry I just noticed this. To work in different settings, we probably need this to point to a path inside of openghg inversions, and have the file there (e.g. the default country file location is `openghg_inversions/countries`), or allow a path to be passed in.

The first option, combined with distributing the file with openghg inversions might be simplest.

brendan-m-murphy:
This is related to issue #56.

brendan-m-murphy:
It looks like this is fixed by PR #101. I'll either split that fix out or just get that PR merged in soon.

Issue #97, opened 2024-04-09
Review how missing data is handled in inversions


brendan-m-murphy:
If no footprint or obs data is found for a site, that site is dropped. This works well for monthly inversions, but less so for annual inversions.

@joe-pitt had some runs where NaNs due to missing footprint data were propegated to `H` and caused a confusing pymc error.

Possible improvements could include:
- adding options to fill or drop NaN values
- modifying the code that drops sites when data isn't found so that it works for annual runs (might be more difficult?)

joe-pitt:
Quick update on this - if a month of fp data from the middle of the dataset is missing from the object store (e.g. you have April 2016 and June 2016, but you are missing May 2016) you get the confusing pymc error:
`pymc.exceptions.SamplingError: Initial evaluation of model at starting point failed!`

If there is a month of fp data missing at the end of the dataset (e.g. MHD obs run to November 2023, but the MHD footprints only run to October 2023) you get a different error:
```
File "/user/home/zh21490/openghg_inversions/openghg_inversions/hbmcmc/inversionsetup.py", line 100, in addaveragingerror
    fp_all[site.upper()].mf_repeatability.values ** 2
ValueError: operands could not be broadcast together with shapes (870,) (1,943) 
```

In this case there are no nans in fp_all, but the timeseries is truncated to the end of the fp data, which causes these subsequent issues. I think the first step would be to raise a more informative error when either of these cases occur, e.g. "Some observations could not be matched to footprints - check that all footprints are loaded in object store". Even better if it can tell you which footprints are missing!

brendan-m-murphy:
If the error happens during "add averaging error", then that might be because the obs data is loaded again from the object store for this part, so probably obs were dropped when the `ModelScenario`s are made, and then we try to add some averaging error based on the full obs time series.

joe-pitt:
Yes I think that is what is happening. So maybe it would run if averaging error was turned off, but by default I think we probably want it to fail, or at the very least give us a strong warning that some of the obs were removed due to missing fps

brendan-m-murphy:
For now, you might just want to turn off "addaveragingerror", because when it loads the obs, it specifies an averaging period, so the averaging error of that data (over the same period) is zero... so I don't think it does anything. I haven't got around to fixing that, but there is an issue open https://github.com/openghg/openghg_inversions/issues/42

joe-pitt:
OK cool thanks! Although in my case I found the missing fps so I'm running fine now

[...]

Issue #56, opened 2024-02-02
UKMO land/sea file for bucket basis function not publicly available


brendan-m-murphy:
The file `"../countries/country-EUROPE-UKMO-landsea-2023.nc"` used in the "bucket basis function" isn't distributed with inversions, and I couldn't find it on Blue Pebble.

We should either include this file with inversions, or modify the code so that the land/sea map can be provided as an argument.

brendan-m-murphy:
Closed by PR #101

Issue #30, opened 2024-01-19
Installation failing related to pymc: workaround


lukewestern:
When running `pip install -e openghg_inversions` I ran into an installation issue related to pymc.

My workaround was to install pymc when creating the new environment (note I used conda rather than a virtual environment here), and ensuring that numpy was version 1.25.2 (the latest numpy version causes conflicts, this happens at runtime):
`conda create -c conda-forge -n openghg_inv "pymc>5" "numpy=1.25.2"`
`pip install --upgrade pip setuptools wheel`
`pip install -e openghg_inversions`

brendan-m-murphy:
This was an issue for about a month, but you should be able to use numpy 1.26 with the latest version of pytensor and pymc: https://github.com/pymc-devs/pytensor/issues/441#event-10911975208. Maybe check to make sure you have the latest version of pytensor? (Although I don't see why you wouldn't get it from pymc...)

For reference, these are some of the packages in a python venv I created recently:

Package                   Version             Editable project location
------------------------- ------------------- ---------------------------------------------------
jax                       0.4.21
jaxlib                    0.4.21
numpy                     1.26.2
numpyro                   0.13.2
openghg                   0.6.2+204.ged8115f8
openghg-defs              0.0.2
openghg-inversions        0.0.2               /user/work/bm13805/.jax_venv/jax_openghg_inversions
opt-einsum                3.3.0
packaging                 23.2
pandas                    1.5.3
pymc                      5.10.1
pytensor                  2.18.3
scipy                     1.11.4
setuptools                69.0.2
wheel                     0.42.0
xarray                    2023.12.0
xarray-einstats           0.6.0

joe-pitt:
Not sure if this is related, but I also got an error related to pymc when running `pip install -e openghg_inversions` 
```
AttributeError: module 'pkgutil' has no attribute 'ImpImporter'. Did you mean: 'zipimporter'?
```
It was resolved by rolling back from python 3.12 to python 3.11 (as described here: https://stackoverflow.com/questions/77364550/attributeerror-module-pkgutil-has-no-attribute-impimporter-did-you-mean)

brendan-m-murphy:
I don't think openghg supports python 3.12 yet (I could be wrong). I suppose this should be in the install instructions, if that's the case.

PR #101, opened 2024-04-16
Numpyro sampler option, documentation, land/sea map fix, save trace fix

brendan-m-murphy:
This branch adds the option to use Numpyro's NUTS sampler in place of the pymc NUTS sampler. This seems to be 7-10x faster. This addresses issue #17 and the ACRG project [issue #30](https://github.com/ACRG-Bristol/projects/issues/30), although profiling needs to be done to assess the speed improvements.

It also adds some documentation by expanding the readme file and adding a short overview (https://github.com/openghg/openghg_inversions/tree/numpyro-fixed-outer-regions)

There are fixes to two other issues as well:
- Issue #99 and issue #56 are addressed by adding the land/sea split file to the repository and changing how this file is searched for. (Issue #99 is only partially addressed, because that asks for the option to specify a path to this file.)
- The `save_trace` option was removed in a previous PR; it is restored in this PR.
- NaN values are dropped before data is passed to pymc, partly addressing issue #97. (Still need to log warnings about missing data.)
- The option to compute the min model error on the fly was added. Also, min model error is saved in the attributes of the inversion output.

Also, a link to the DOI associated with releases was added to the readme file, and a test was added for the bucket/weighted basis function.


brendan-m-murphy:
@joe-pitt reports: error using weighted basis with fixed outer regions:

[Yesterday 21:23] Joe Pitt
  File "/user/home/zh21490/openghg_inversions/openghg_inversions/basis/_functions.py", line 201, in fixed_outer_regions_basis

    basis.loc[loc_dict] = (inner_region + 5).squeeze().values

[Yesterday 21:24] Joe Pitt
ValueError: could not broadcast input array from shape (183,128) into shape (182,127)

joe-pitt (approved):
It turns out the error Brendan mentioned only occurs for a certain flux file (or potentially a certain type of flux file), and it is actually giving the same behaviour on devel as on this branch. So it isn't an issue with these changes. I have successfully run a test ch4 inversion on this branch so I'm going to approve the PR.

PR #105, opened 2024-04-18
104 inversion does not know how to deal with a site where all data have been filtered

alexdanjou:
Addition of a check to see if all the data has been filtered for a site and removed the site if this is the case.

### DIFF
--- a/openghg_inversions/hbmcmc/hbmcmc.py
+++ b/openghg_inversions/hbmcmc/hbmcmc.py
@@ -378,15 +378,15 @@ def fixedbasisMCMC(
 
     # Apply named filters to the data
     fp_data = utils.filtering(fp_data, filters)
-    
+
     s_dropped = []
     for site in sites:
         # check if some datasets are empty due to filtering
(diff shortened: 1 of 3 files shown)


openghg_inversions/hbmcmc/hbmcmc.py
```diff
@@ -378,6 +378,16 @@ def fixedbasisMCMC(
 
     # Apply named filters to the data
     fp_data = utils.filtering(fp_data, filters)
+    
+    s_dropped = []
+    for site in sites:
+        # check if some datasets are empty due to filtering
+        if fp_data[site].time.values.shape[0]==0:
```
brendan-m-murphy, line 385:
```suggestion
        if fp_data[site].time.values.shape[0] == 0:
```

brendan-m-murphy (approved):
Looks good! Could you add an entry to CHANGELOG.md that summarizes the change? You should be able to commit that without needing another review. (Also, you can commit the suggestion I made without needed a new review). Thanks!

Issue #110, opened 2024-05-01
Use observations in error scaling


brendan-m-murphy:
To scale uncertainties in the likelihood to match large pollution events, we need to use the observations themselves, rather than the modelled obs, as currently done.

joe-pitt:
Closed by PR #111

PR #111, opened 2024-05-01
Added option to use obs to compute pollution event size

brendan-m-murphy:
Using the option `pollution_events_from_obs = True` uses the obs minus the modelled baseline as the size of the pollution events in the model-data mismatch error. (If `use_bc` is False, then just the obs are used as the pollution event size.)

This addresses issue #110

### DIFF
--- a/openghg_inversions/hbmcmc/inversion_pymc.py
+++ b/openghg_inversions/hbmcmc/inversion_pymc.py
@@ -137,6 +137,7 @@ def inferpymc(
     save_trace=False,
     use_bc: bool = True,
     reparameterise_log_normal: bool = False,
+    pollution_events_from_obs: bool = False,
 ):
     """
     Uses PyMC module for Bayesian inference for emissions field, boundary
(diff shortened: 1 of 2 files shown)


joe-pitt (commented):
In general this looks good to me. The only thing I'm not sure about is the case where use_bc=False. There may be some instances where we're happy to apply sig to the absolute mole fraction, but clearly there are other cases where this would not be sensible. Do we want to print some sort of warning here? I'm not sure whether it would be intuitive to people what would happen with use_bc_=False, pollution_events_from_obs=True

ag12733:
I think the issue is just lack of documentation. I don’t think it states that to use use_bc=False means that you need to pre-subtract the bc from the data so that y (obs) represents only the enhancements.

I think Joe’s point is because he didn’t know that obs are only enhancements so there are no fundamental issues, just lack of documentation I think!

________________________________

Dr. Anita Ganesan
Associate Professor
School of Geographical Sciences
University of Bristol
Bristol, BS8 1SS, United Kingdom
Tel: +44 117 455 1545<tel:+441174551545>
Email: ***@***.******@***.***>


From: Joe Pitt ***@***.***>
Date: Wednesday, 1 May 2024 at 21:00
To: openghg/openghg_inversions ***@***.***>
Cc: Anita Ganesan ***@***.***>, Review requested ***@***.***>
Subject: Re: [openghg/openghg_inversions] Added option to use obs to compute pollution event size (PR #111)

@joe-pitt commented on this pull request.

In general this looks good to me. The only thing I'm not sure about is the case where use_bc=False. There may be some instances where we're happy to apply sig to the absolute mole fraction, but clearly there are other cases where this would not be sensible. Do we want to print some sort of warning here? I'm not sure whether it would be intuitive to people what would happen with use_bc_=False, pollution_events_from_obs=True

joe-pitt (approved):
Yeah that make sense now - this seems good as it is

brendan-m-murphy:
There is some extra documentation about these extra parameters in the numpyro PR, I was planning to document these changes once this gets merged into devel, because I could pull these changes into the PR with the new documentation

Issue #113, opened 2024-05-08
Specify fp species in ini file


joe-pitt:
At the moment data_processing_surface_notracer can't select an fp for a specific species (except co2) from a store. So if you have inert fps and short lifetime fps in the same store, the call to get_footprint fails because it finds multiple matching entries.

joe-pitt:
Closed with RP #114

PR #114, opened 2024-05-08
Specify fp species in ini file

joe-pitt:
Adding the ability to specify fp_species, so that you can retrieve footprints from a store that contains inert footprints and species-specific footprints.

This addresses issue #113

### DIFF
--- a/openghg_inversions/get_data.py
+++ b/openghg_inversions/get_data.py
@@ -44,6 +44,7 @@ def data_processing_surface_notracer(
     met_model=None,
     fp_model=None,
     fp_height=None,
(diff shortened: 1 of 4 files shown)


hdelongueville (commented):
Maybe we should write somewhere that species = 'inert' is an option, because I don't find it really intuitive when you're not working with a specific species like for F-gases.

joe-pitt:
Good idea - I'll add that to the .ini file now

hdelongueville:
[like]  Helene De Longueville reacted to your message:
________________________________

Issue #116, opened 2024-05-09
min_model_error as a minimum value


joe-pitt:
The min_error calculated by residual_error_method includes time-invariant contributions from the obs error, representation error and model error. We also account for these terms in a time-variant way through other calculations (although addaveragingerror needs modifying - see Issue #42 ). Currently the min_error is added to the time-variant errors to yield the model-obs mismatch error, which means that errors are being double counted.

It would be better if this min_error was used as a floor below which the model-obs mismatch error (epsilon) never drops, and when the time-variant errors are larger than min_error we use those values (i.e. ignore min_error in those cases). That follow the approach taken in InTEM.

PR #117, opened 2024-05-09
Make min_error a minimum value and update README

joe-pitt:
Changed inferpymc so that it interprets min_error as a minimum value (lower bound) rather than adding it to the other errors for all points. This addresses Issue #116 .

Needs looking at by someone who understands pytensors before approval - I'm not certain that what I've done here is correct.

### DIFF
--- a/openghg_inversions/hbmcmc/inversion_pymc.py
+++ b/openghg_inversions/hbmcmc/inversion_pymc.py
@@ -285,7 +285,7 @@ def inferpymc(
             pollution_event = np.abs(pt.dot(hx, x))
 
         pollution_event_scaled_error = pollution_event * sig[sites, sigma_freq_index]
-        epsilon = pt.sqrt(error**2 + pollution_event_scaled_error**2 + min_error**2)
+        epsilon = pt.maximum(pt.sqrt(error**2 + pollution_event_scaled_error**2), min_error)
         y = pm.Normal("y", mu=mu, sigma=epsilon, observed=Y, shape=ny)
 
         step1 = pm.NUTS(vars=step1_vars)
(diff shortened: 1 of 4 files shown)


hdelongueville (approved):
I'm not an expert on pytensors but this change seems right to me.

openghg_inversions/hbmcmc/inversion_pymc.py
```diff
@@ -285,7 +285,7 @@ def inferpymc(
             pollution_event = np.abs(pt.dot(hx, x))
 
         pollution_event_scaled_error = pollution_event * sig[sites, sigma_freq_index]
-        epsilon = pt.sqrt(error**2 + pollution_event_scaled_error**2 + min_error**2)
+        epsilon = pt.maximum(pt.sqrt(error**2 + pollution_event_scaled_error**2), min_error)
```
brendan-m-murphy, line 288:
```suggestion
        epsilon = pt.max(pt.sqrt(error**2 + pollution_event_scaled_error**2), min_error)
```

brendan-m-murphy (changes requested):
I think the pytensor function is called `max`: https://pytensor.readthedocs.io/en/latest/library/tensor/basic.html#reductions although it's possible that `maximum` works too... their documentation isn't very good.

Otherwise, looks good!

openghg_inversions/hbmcmc/inversion_pymc.py
joe-pitt: Hmmm I think pt.maximum and pt.max do different things (https://pytensor.readthedocs.io/en/latest/library/tensor/basic.html#pytensor.tensor.maximum). I'm not certain that pt.maximum is right though, so you better check!

brendan-m-murphy:
It looks like `pt.maximum` is what we want: https://pytensor.readthedocs.io/en/latest/library/tensor/basic.html#mathematical We could also achieve this with the `switch` statement.

brendan-m-murphy (approved):

PR #121, opened 2024-05-14
Tidied up input files and added comments on variables

EricSaboya:
Cleaned up the example and blank template .ini files for RHIME with OpenGHG. Variable names and examples should be exactly the same as before. I've just included more comments and grouped the variables together in what is hopefully a more intuitive manner.

### DIFF
--- a/openghg_inversions/hbmcmc/config/openghg_hbmcmc_input_template_example.ini
+++ b/openghg_inversions/hbmcmc/config/openghg_hbmcmc_input_template_example.ini
@@ -1,187 +1,288 @@
 ; =======================================================================================
 ; Regional Hierarchical Inverse Modelling Environment Example Configuration File
 ; =======================================================================================
-; RHIME example configuration file for CH4. Please check through variable paths and 
-; names before running.
-; Required inputs are marked as such.
+; RHIME-with-OpenGHG example configuration file for CH4 
+; Please check through variable paths and names before running
+; This can be done  using openghg.retrieve.search() in Python
+; and specifying the relevant object stores and other variables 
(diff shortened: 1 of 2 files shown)


openghg_inversions/hbmcmc/config/openghg_hbmcmc_input_template_example.ini
```diff
@@ -1,187 +1,288 @@
 ; =======================================================================================
 ; Regional Hierarchical Inverse Modelling Environment Example Configuration File
 ; =======================================================================================
-; RHIME example configuration file for CH4. Please check through variable paths and 
-; names before running.
-; Required inputs are marked as such.
+; RHIME-with-OpenGHG example configuration file for CH4
```
brendan-m-murphy, line 4: If this file is the sample as the template file, but has sensible defaults, then I think maybe we should just make this the template. Having empty values is a bit confusing because the template isn't actually valid, and the values will be replaced when `run_hbmcmc.py` makes a copy.

We should double check this... but as long as the only CH4 specific info is in "required" categories, then that info won't be propagated through to the output from `run_hbmcmc.py`.

(Also, matching the function defaults will help make the output "correct".)

brendan-m-murphy (changes requested):
Thanks for the update! I would like to consider replacing the template with the "template_example" file. As long as this is done sensible, it will work the same for generating the output from `run_hbmcmc.py`, and it will mean one less file to keep updated. Plus, I think the empty values in the actual template are sort of confusing.

brendan-m-murphy:
Closed since it is far behind devel. We can revive this later if we need to.

PR #125, opened 2024-05-15
Pass met_model as a list to get_footprint

brendan-m-murphy:
* **Summary of changes**

`met_model` can be passed as list in an ini file, to cover the case where two footprints are distiguished only by met model. The list must have the same length as the number of sites. Values of `None` will be ignored, so for sites where the met model doesn't matter, `None` can be passed.

- [ ] Tests passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 # Version 0.2.0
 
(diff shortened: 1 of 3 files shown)


joe-pitt (approved):
Looks good - thanks for doing this!

PR #142, opened 2024-06-14
Added min model error options, incl "by site"

### LINKED ISSUE #136: Need to rethink min model error
In the early years of the HFO inversions, InTEM and ELRIS have almost zero emissions, while RHIME estimates substantial emissions for the whole time period, even though the obs show very few enhancements. The model error also looks wrong for RHIME in early years. Will elaborate on this further once I've investigated...

brendan-m-murphy:
Also added "robust" option to use median absolute deviation in place of standard deviation.

* **Summary of changes**

In an ini file use:
```
compute_min_error = True
min_error_options = {"by_site": True, "robust": True, "average_over": "by_site"} 
```

to use new options. (Either "by_site" or "robust" can be omitted; or `min_error_options` can be omitted entirely)

- [x] Closes #136 
- [x] Tests added and passing


joe-pitt:
Adding an option for a percentile-based calculation on min_model_error. This changes what is required in the ini - calculate_min_error now takes either "percentile" or "residual". Need to update docs/templates to reflect this

joe-pitt:
One thing that has just occurred to me - do we need to raise an error in the case when calculate_min_error is neither "percentile", "residual" or None? I guess at the moment if someone gets the spelling wrong it would silently skip the min error calculation

brendan-m-murphy:
@joe-pitt I've updated the docs with the new options and added some code to raise an error if the value of `compute_min_error` is something we don't recognise.

joe-pitt (approved):
Looks good! I think it does close Issue #136 , at least in that we now have various options and it would be easy to add more

PR #144, opened 2024-06-21
Obs error clean up

### LINKED ISSUE #42: `addaveragingerror` doesn't actually do anything due to functionality in `get_obs_surface`
TLDR: `get_obs_surface` already performs the calculations in `addaveragingerror` and resamples the data, so `addaveraging` error doesn't actually change anything.

If an `averaging` argument is passed to `openghg.retrieve.get_obs_surface`, then the following happens:

1. The obs data is resampled to the given resampling period:
```
ds_resampled = data.resample(time=average).mean(skipna=False, keep_attrs=True)
```
2. If `ch4_repeatability` (or `..._repeatability`) is present, then in the resampled data, `ch4_repeatability` is set to the variance of the mean of the original repeatability values:
```
if "repeatability" in var:
                ds_resampled[var] = (
                    np.sqrt((data[var] ** 2).resample(time=average).sum())
                    / data[var].resample(time=average).count()
                )

```
3. A variable called `f{species}_repeatability` (e.g. `ch4_repeatability`) is added, which contains the standard deviation of the observations over the resampling period:
```
ds_resampled[f"{species}_variability"] = (
      data[species].resample(time=average).std(skipna=False, keep_attrs=True)
)
```

Now, `addaveragingerror` in `openghg_inversions` does the following:
1. gets the obs again, with the same averaging period as specified in `fixedbasisMCMC` or `get_data.py`:
```
 get_obs = get_obs_surface(site=site,
                                  species=species,
                                  inlet=inlet[i],
                                  instrument=instrument[i],
                                  average=meas_period[i],
                                  start_date=start_date,
                                  end_date=end_date)
```
2. converts the obs values to a data frame called `sitedataerr` and pads the date range with NaNs: I'm not totally sure why this step is necessary
3. If `mf_variability` is present (which will always be the case if `meas_period[i]` is not `None`), it sets the `mf_variability` data variable of `fp_all[site]` to
```
np.sqrt(fp_all[site.upper()].mf_variability.values**2 + sitedataerr.resample(meas_period[i]).std(ddof=0).dropna().values.T**2))
```
Since the data was resampled to `meas_period[i]` by `get_obs_surface`, the second term in the square root is 0 (since the standard deviation of one value is zero), so `mf_variability` isn't changed.
4. The same is done for `mf_repeatability`: the `mf_repeatability` of `fp_all[site]` (if it is present) is set to
```
np.squeeze(np.sqrt(fp_all[site.upper()].mf_repeatability.values**2 + sitedataerr.resample(meas_period[i]).std(ddof=0).dropna().values.T**2))
```
but again, the second term in the `np.sqrt` is zero.

What this *should* be doing is adding the `mf_variability` to the `mf_repeatability` if `mf_repeatability` is present. This doesn't require fetching the data again.

brendan-m-murphy:
* **Summary of changes**

Cleaned up code that adds averaging error, using the `mf_variability` variable that is created by OpenGHG's `get_obs_surface` function.

A new variable `mf_error` is created:
- if "add averaging error" is true, then `mf_error` is the sqrt root of the sum of squares of repeatability and variability
- otherwise, `mf_error` is repeatability if it is present, otherwise, it is variability.

The output file now includes the variables:
- `Yerror_repeatability` containing the combined `mf_repeatability` for all sites
- `Yerror_variatability` containing the combined `mf_variatability` for all sites

These output variables are needed for the PARIS outputs

- [x] Closes #42 
- [x] Tests added and passing


joe-pitt (approved):
Looks good! I have successfully run inversions on the min_error_by_site_plus_robust branch after this branch was merged in at commit fe0cc36. I haven't run any inversions to test the changes to this branch since that point, but they look fine to me.

PR #145, opened 2024-06-24
Change name of hcfo-1233zde and add lifetime info to json

joe-pitt:
The species.json file previously had the wrong name for hcfo-1233zde and it didn't include the lifetime info. Fixed here.


brendan-m-murphy (approved):
Looks good!

PR #162, opened 2024-07-05
move basis function code into basis module

### LINKED ISSUE #157: Move basis related functions from `utils.py` to `basis` submodule
The functions:
- `basis`
- `basis_boundary_conditions`
- `fp_sensitivity` (and `fp_sensitivity_single_site` and `timeseries_HiTRes`)
- `bc_sensitivity`

Are only used by the basis functions wrapper, and their main function is to load basis functions or apply the basis functions to the flux/fp, or bc/fp.

So they should live with the other basis functions code.

qq23840:
Moved basis-related functions from `utils.py` to `basis._functions.py` and modified imports as required. 
These are `basis`, `basis_boundary_conditions`, `timeseries_HiTRes`, `fp_sensitivity`, `fp_sensitivity_single_site_basis_func` and `bc_sensitivity`

Also made a slight change in the way the TensorVariable type was imported in `inversion_pymc.py`, as it was causing tests to fail. 

- [x] Closes #157 
- [x] Tests added and passing (except three strange ones, think it's to do with my openghg branch)
- [ ] Documentation and tutorials updated/added NOT NEEDED
- [ ] Added any new requirements to `requirements.txt` NOT NEEDED


brendan-m-murphy (approved):
Thanks for the update! I merged in devel and moved some functions to `basis/_helpers.py`, but otherwise I didn't make any extra changes (besides formatting/typing)

PR #177, opened 2024-07-19
Fetch most recent flux

brendan-m-murphy:
* **Summary of changes**

If flux data isn't found for the given start and end dates, `data_processing_surface_notracer` will try to search again without the start date.

- [x] Tests added and passing ...kind of. The tests against OpenGHG 0.7 are failing, but I don't think it's worth fixing


hdelongueville (approved):
It seems to work for me!

PR #179, opened 2024-07-19
Iss176 local influence filter issue

### LINKED ISSUE #176: `local_influence` filter in the wrong category
Currently, passing `local_influence` as a filter throws an error as it wasn't included in a list of filters when that was refactored I think. It doesn't require a 'site' argument, but is called in a line which passes one. 

Should be a very easy fix

brendan-m-murphy:
* **Summary of changes**

Fixed the `filtering` function in `filters.py`, which was passing "site" as an argument for some filters, even though that argument was removed from all filters.

I also updated the tests for the filters to make sure that all filters are checked.

- [x] Closes #176 (but `local_influence` filter is possible very slow)
- [x] Tests added and passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # OpenGHG Inversions Change Log
 
 # Version 0.2.0
+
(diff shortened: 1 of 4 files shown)


qq23840:
Looks good to me

qq23840 (approved):

PR #182, opened 2024-07-24
More flexible filter instrument options

brendan-m-murphy:
* **Summary of changes**

Allow `filtering` argument in ini file to have missing sites. A warning will be logged and these sites will not have any filtering applied.

Allow inlet, instrument, fp_height, obs_data_level, and met_model to be single strings. In this case, that value is used for all sites.

- [ ] Tests added and passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 # Version 0.2.0
 
+- Made format for `filtering` in ini file allow for missing sites. Made `inlet`, `instrument`, `fp_height`, `obs_data_level`, and `met_model`
+  accept a single string in the ini file, which will be converted to a list of the correct length.  [#PR 182](https://github.com/openghg/openghg_inversions/pull/182)
+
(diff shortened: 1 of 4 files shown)


alexdanjou (approved):
Great!

hdelongueville (commented):
It seems that, in my the case, where 
1. MHD is in the ini file's sites list and filters list, but 
2. "No obs data found for MHD with inlet None and instrument None. Check these values. Continuing model run without MHD.", then
3.  the filtering is failing because it is not finding MHD in the dataset:
File "./openghg_inversions/openghg_inversions/filters.py", line 136, in filtering
    n_nofilter = datasets[site].time.values.shape[0]
                 ~~~~~~~~^^^^^^
KeyError: 'MHD'

Could it be that 'filters' also looks for the sites which should have been removed? 
If so, a solution would be to update filters in the dataset after checking for obs, or to add a line to check if site is in sites after line 128.

PR #186, opened 2024-07-31
Basis function fp x flux calculation fix

brendan-m-murphy:
* **Summary of changes**

xarray alignment was dropping times that were missing from some footprints, which causes problems if there are footprints with non-overlapping times. (Due to footprint_data_merge, these missing times can come from missing obs as well.)

This fixes that problem by explicitly aligning the footprints and filling missing times with 0's.

- [ ] Tests added and passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 # Version 0.2.0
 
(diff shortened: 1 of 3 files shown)


alexdanjou (approved):

alexdanjou:
All looks good to me, thanks for this needed correction!
One problem though : tests fails _FAILED tests/test_get_data.py::test_add_averaging_error - ValueError: cannot specify both keyword and positional arguments to .resample_ (not related to your changes)

PR #199, opened 2024-08-30
Iss198 weighted eastasia

### LINKED ISSUE #198: Weighted algorithm hard-coded for EUROPE domain only
Would like to apply the weighted algorithm to generate basis functions for the EASTASIA domain. Involves a little bit of refactoring the basis module, but happy to do this.

qq23840:
Update the basis module such that the 'weighted' algorithm can be used for the EASTASIA domain. This incorporates a land-sea split netCDF file. 

- [x] Closes #198 
- [x] Tests added and passing (no new tests, but tests updated to reflect change in code)


openghg_inversions/basis/algorithms/_weighted.py
hdelongueville, line 1: Wouldn't it be useful to add a warning message if there is no country-landsea file linked to the domain (i.e., domain other than EASTASIA and EUROPE)? 
e.g.
if domain == 'EASTASIA':
        landsea_indices = xr.open_dataset(Path(__file__).parent / "country-land-sea_EASTASIA.nc")
elif domain == 'EUROPE':
        landsea_indices = xr.open_dataset(Path(__file__).parent / "country-EUROPE-UKMO-landsea-2023.nc")
else:
        ++ add a warning message here like "No landsea mask file related to {domain}. The file country-EUROPE-UKMO-landsea-2023.nc is used instead." ++
        landsea_indices = xr.open_dataset(Path(__file__).parent / "country-EUROPE-UKMO-landsea-2023.nc")

qq23840: Yes sounds sensible - I'll add it with a logger warning

hdelongueville (approved):
Looks good to me!

brendan-m-murphy (changes requested):
Around line 370 of `basis/_functions.py`, you need to pass "domain" to the inner region basis function.

I guess we should have a test for the fixed out regions basis function, if we don't already. (I think this should be an easy modification of the existing basis functions tests, but if it isn't, I can try to make a new test.)

brendan-m-murphy (approved):
Look good, thanks!

PR #206, opened 2024-09-10
allowed outer region to include EASTASIA and added region file

### LINKED ISSUE #201: Extend InTEM outer regions to EASTASIA
Would like to fix the outer regions for EASTASIA when creating basis function a la EUROPE / `intem_region_definition.nc`

qq23840:
Adds an EASTASIA outer region file and modifies the outer region basis function code to read in any outer region file. 

Currently a pretty arbitrary outer region - I will check with Alistair/Alice whether there is an 'official' set of outer regions that InTEM uses, for consistency (hence the draft PR). 

- [x] Closes #201 
- [ ] Tests added and passing


brendan-m-murphy:
Looks good. You could parameterise the test you added by using the decorator `@pytest.mark.parametrize("domain", ["EUROPE", "EAST_ASIA"])` above the test and making `domain` an argument of the test function. Let me know when you want a review.

qq23840:
Have added the official regions from @aliceramsden. To add a test I'd have to generate a load of files, as all the test data is only set up for ch4 /  EUROPE at the moment. Maybe a code retreat task rather than an urgent priority?

brendan-m-murphy:
> Have added the official regions from @aliceramsden. To add a test I'd have to generate a load of files, as all the test data is only set up for ch4 / EUROPE at the moment. Maybe a code retreat task rather than an urgent priority?

Ah good point. Yea we can leave it for now. It would be good to be able to test the basis functions without running all of the code in `get_data.py` (so we could use synthetic data for the flux * fp, say).

brendan-m-murphy (approved):
Looks good!

PR #211, opened 2024-10-04
Added pm.Data to track data used in model

brendan-m-murphy:
* **Summary of changes**
 

Using `pm.Data` for external data used in the model will make sure that data is added to the `InferenceData` returned from sampling, and will apply dimensions and coordinates to that data.

This will make it easier to construct the outputs from the inversion.

I also added a check to make sure `min_error` is the correct size.

- [ ] Tests added and passing


alexdanjou (changes requested):
Failed because `min_error` can be a 0-dim array. Would suggest to change l.287 to 
`if isinstance(min_error, float) or min_error.shape==():` (or check why `min_error` arrives as a 0-d array)

alexdanjou (approved):

PR #213, opened 2024-10-08
Update getting_started.md

alexdanjou:
Update bash example for BluePebble in getting_started.md (as discussed during meeting).
Needed as the uploaded modules where out of date.

- [ ] Tests added and passing
- [X] Documentation and tutorials updated/added

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -164,7 +164,8 @@ This script assumes that you have already created a conda env called `pymc_env` 
 
 
 # Set up Python environment
-module load lang/python/anaconda
+module --force purge
+module load git/2.45.1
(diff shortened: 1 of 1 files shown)


brendan-m-murphy (commented):
Does this still need `module load languages/python` after the `module --force purge`? Or do we not need that because we have to install conda ourselves now? (I haven't used SLURM much since they changed to Rocky linux.)

alexdanjou:
I don't think so (at least I don't need it anymore to launch a job).
I think that the reason you say (on the fact that we don't need it as we now install anaconca ourselves) is the right one.

hdelongueville (approved):

PR #224, opened 2024-11-13
Changed recursive optimization to loop

brendan-m-murphy:
* **Summary of changes**

Changed `optimize_nregions` from recursive function to loop, so that it can run longer if necessary.
Plus added extra loop to increase tolerance if initial tolerance is too strict.

This was in response to situations where the weighted basis optimisation was exceeding the max recursion depth.

Caching was added to the function that reads the land/sea definition file, so that the result is kept in memory. 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # OpenGHG Inversions Change Log
 
+- Changed optimization in weighted basis function from recursion to loop. [#PR 224](https://github.com/openghg/openghg_inversions/pull/224)
+
 - Added MHD obs and footprint to test data. [#PR 209](https://github.com/openghg/openghg_inversions/pull/209)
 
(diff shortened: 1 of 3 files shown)


brendan-m-murphy:
@EricSaboya yea that's a good question. In the case that caused this problem, there was a lot of bounding back and forth between * .995 and * 1.005, and it kind of seemed like it just couldn't get close enough to the target, so relaxing the tolerance helped. 

Maybe going to 2000 or 3000 iterations would be better though.. since (1.005)**1000 is about 150, so going to 2000 iterations could get the threshold over 20,000, and the lower bound could go closer to 1/10,000. 

It seems like 1,000 might be the recursion limit: https://stackoverflow.com/questions/3323001/what-is-the-maximum-recursion-depth-and-how-to-increase-it, although I think I might have chosen 1,000 iterations arbitrarily.

I guess the range of thresholds necessary sort of depends on the size of the values in the array as well.

It is probably also possible to use some optimization function like quadtree uses.

PR #239, opened 2025-03-13
Iss233 flask data fix

### LINKED ISSUE #233: How does RHIME treat flask data?
I need to run RHIME using in situ obs from "CMN", "JFJ", "MHD", "TAC", "TOB", "ZEP", as well as flask data from "TOB", "CBW".

Is there a way to differentiate the different type of data? 
We also need to think about the resampling for data with coarser frequency (e.g. weekly flask sampling).
Is this the moment to include "platform" in RHIME?

This issue is related to #232 where RHIME doesn't like calling a site twice.

brendan-m-murphy:
* **Summary of changes**

Add option to pass `platform`, which will be used by `ModelScenario`. 

For instance, in a ini file:
```
sites = ["TAC", "TOB"]
platform = [None, "flask"]
```

To test that this is working, you can add
```
output_format = "merged_data"
```
to your ini file. This will just return `fp_all`, and won't save anything to disk, but it will stop the run before actually doing the inversions.

If you also add `save_merged_data = True`, you can see the merged data after by using `load_merged_data` in a notebook (e.g. `from openghg_inversions.inversion_data import load_merged_data`).

* **Please check if the PR fulfils these requirements**

- [ ] Closes #233 
- [ ] Tests added and passing


hdelongueville:
I've tested this PR and it seems to work. The job completed. I've saved the merged data (here hfc143a/merged_data/hfc143a_2021-01-01_PARIS_merged-data.zarr.zip) but I am not sure what should I look at.

PR #242, opened 2025-03-18
Adapting postprocessing code to work for other domains

### LINKED ISSUE #241: PARIS-style post-processing only works for EUROPE domain
We routinely run EASTASIA inversions, and I'm starting to run on the WESTUSA domain as well. It would be good to be able to run these using the PARIS post-processing code, as we're looking to compare with Alistair's InTEM results at least.

qq23840:
This PR adapts the new postprocessing code to work for the EASTASIA domain as well, and implements tests to do this. 

- [X] Closes #241 
- [X] Tests added and passing (locally, although not on the CI workflow against OpenGHG 0.12). 


openghg_inversions/postprocessing/make_outputs.py
```diff
@@ -226,12 +227,12 @@ def make_country_outputs(
 
     """
     if country_regions == "paris":
-        country_regions = paris_regions_dict
+        country_regions = paris_regions_dict[domain.lower()]
```
brendan-m-murphy, line 230: You can use `inv_out.domain` to get the domain (this is why most functions that take a `InversionOutput` argument don't also take a "domain" or "species" output).

openghg_inversions/postprocessing/countries.py
```diff
@@ -197,6 +202,7 @@ def __init__(
         country_selections: list[str] | None = None,
         country_code: Literal["alpha2", "alpha3"] | None = None,
         country_regions: dict[str, list[str]] | str | Path | None = None,
+        domain: str | None = None
```
brendan-m-murphy, line 205: If we don't need to use "domain" in any of the methods for "Countries", then it doesn't need to be in the init. Usually domain info comes from the inversions outputs. (I guess the country file is usually associated with a particular domain though.)

tests/test_postprocessing.py
```diff
@@ -68,14 +82,39 @@ def test_basic_outputs(inv_out, europe_country_file):
...
+    flux_outs, conc_outs = make_paris_outputs(inv_out, country_file=europe_country_file, obs_avg_period="1h", domain="europe")
 
     flux_outs.to_netcdf(tmpdir / "flux.nc")
     conc_outs.to_netcdf(tmpdir / "conc.nc")
 
+def test_make_paris_outputs_EASTASIA(inv_out_EASTASIA, eastasia_country_file, tmpdir):
+    """Check that we can create and save PARIS outputs for EASTASIA domain"""
+    flux_outs, conc_outs = make_paris_outputs(inv_out_EASTASIA, country_file=eastasia_country_file, obs_avg_period="1h", domain="eastasia")
+
+    flux_outs.to_netcdf(tmpdir / "flux.nc")
```
brendan-m-murphy, line 116: we probably don't need to check that writing to disk works for both EUROPE and EASTASIA. I just wanted one check since this step was causing Joe problems at some point. But keeping the first part that creates the outputs would still be useful.

brendan-m-murphy (changes requested):
Looks good!

I'm not sure we need to add `domain` to `Countries`, since it is available from `InversionOutput`, but usually (always?) country files are associated with a particular domain, so it kind of makes sense to have it with the country file too.

I'd probably remove the `domain` argument where possible, but it isn't doing any harm, so up to you.

Also just a nitpick, but ideally the test function names would be all lower case, so "eastasia" instead of "EASTASIA". Not a big deal though if you want to leave it.

Thanks!

brendan-m-murphy (approved):
Looks good! Thanks for taking care of this.

PR #249, opened 2025-04-03
Added fix to convert BC units in fp_all

brendan-m-murphy:
This was removed (by mistake) when we refactored `get_data.py`

I added a test to make sure the modelled obs are somewhat close to the obs, to check if the baseline has the right order of magnitude.

* **Summary of changes**

Added fix to convert BC units in `fp_all`.
This was removed (by mistake) when we refactored `get_data.py`

I added a test to make sure the modelled obs are somewhat close to the
obs, to check if the baseline has the right order of magnitude.

- [x] Tests added and passing

### DIFF
--- a/openghg_inversions/inversion_data/get_data.py
+++ b/openghg_inversions/inversion_data/get_data.py
@@ -18,6 +18,7 @@ import xarray as xr
 
 from openghg.retrieve import get_bc
 from openghg.types import SearchError
+from openghg.util import extract_float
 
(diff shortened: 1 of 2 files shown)


alexdanjou (approved):
Looks good to me. Can wait to rerun my inversions with this :D

pearson01:
All good to me!

pearson01 (approved):

PR #253, opened 2025-04-04
Remove met_model = met_model or 'not_set'

### LINKED ISSUE #252: get_footprint_to_match replaces None met_model with not_set
When fp_height = 'auto', get_footprint_to_match is used and the following line is causing issues:

`met_model = met_model or "not_set"  # replace None with 'not_set'`

It is returning an empty data when met_model is something else than 'not_set', like 'ukv'.

Is there a reason why this line is used in the first place? Why not keeping met_model to None?
A simple solution would be to remove this line.

hdelongueville:
* **Summary of changes**
Allow for None met_model to be different that 'not_set' when using fp_height = 'auto'. 
The original line was probably not useful in the first place. 
If people wants to specify 'not_set', they can specify it in the ini file.

- [x] Closes #252 
- [ ] Tests added and passing

### DIFF
--- a/openghg_inversions/inversion_data/getters.py
+++ b/openghg_inversions/inversion_data/getters.py
@@ -198,8 +198,6 @@ def get_footprint_to_match(
     start_date = start_date or obs._start_date
     end_date = end_date or obs._end_date
 
-    met_model = met_model or "not_set"  # replace None with 'not_set'
-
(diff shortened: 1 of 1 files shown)


brendan-m-murphy (approved):
Look good to me! Thanks for finding/fixing this problem!

PR #267, opened 2025-04-11
Updates to make compatible with devel openghg

alexdanjou:
* **Summary of changes**
Change add_obs_error and vars to check for nans in fixedbasisMCMC to agree with PR #1298 of openghg

Add inlet_height to the keep_variables for get_obs_data to agree with PR #1300 and resolve issue #1299 of openghg

- [ ] Tests added and passing

### DIFF
--- a/openghg_inversions/inversion_data/get_data.py
+++ b/openghg_inversions/inversion_data/get_data.py
@@ -84,7 +84,8 @@ def add_obs_error(sites: list[str], fp_all: dict, add_averaging_error: bool = Tr
                 )
 
         elif add_averaging_error:
-            ds["mf_error"] = np.sqrt(ds["mf_repeatability"] ** 2 + ds["mf_variability"] ** 2)
+            ds["mf_error"] = np.sqrt(ds["mf_repeatability"].fillna(0) ** 2 + ds["mf_variability"].fillna(0) ** 2)
+            ds = ds.where(~(np.isnan(ds["mf_repeatability"]) & np.isnan(ds["mf_variability"])))
         else:
             ds["mf_error"] = ds["mf_repeatability"]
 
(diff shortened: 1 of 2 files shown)


openghg_inversions/inversion_data/get_data.py
```diff
@@ -84,7 +84,8 @@ def add_obs_error(sites: list[str], fp_all: dict, add_averaging_error: bool = Tr
                 )
 
         elif add_averaging_error:
-            ds["mf_error"] = np.sqrt(ds["mf_repeatability"] ** 2 + ds["mf_variability"] ** 2)
+            ds["mf_error"] = np.sqrt(ds["mf_repeatability"].fillna(0) ** 2 + ds["mf_variability"].fillna(0) ** 2)
```
brendan-m-murphy, line 87: maybe add a comment before this line to explain the strategy (e.g. "fill with zeros so that if one of repeatability and variability is not NaN, then mf_error will not be NaN. Then drop values where repeatability and variability are both NaN")

brendan-m-murphy (changes requested):
Maybe just add a comment to explain the idea for the `add_averaging_error` case.

It looks like we're not dropping values where `mf_error` is zero. Should we open an issue to remind us to come up with some strategy for this? (I think this case is covered in the PyMC code, so it shouldn't cause a divide by zero problem, it's just if we want to get rid of data with no uncertainty.)

alexdanjou:
> Maybe just add a comment to explain the idea for the `add_averaging_error` case.
I've commented each line to explain the `fillna(0)` and the `where(...)`.
 
> It looks like we're not dropping values where `mf_error` is zero. Should we open an issue to remind us to come up with some strategy for this? (I think this case is covered in the PyMC code, so it shouldn't cause a divide by zero problem, it's just if we want to get rid of data with no uncertainty.)
I agree that just printing a warning maybe a bit light and I think we would want to get rid of those data. I'll open an issue so that the other can give their thoughts about it

brendan-m-murphy (approved):
Looks good! And good thing you added the comments because I realised I misinterpreted what was happening... we don't use `where(..., drop=True)` so those values in `mf_error` are set to NaN, not dropped (like I thought).

PR #272, opened 2025-04-15
set flux time to start_date when using an anterior flux prior

### LINKED ISSUE #258: Wrong timestamp in output
When I try to do monthly inversions for n2o for 2023 with`output_format = 'paris'`, my flux outputs have a timestamp equal to `"2022-07-07"`. This is possibly due to the fact that no flux data is available for 2023 and the one from 2022 are read.

Solution proposed b Brendan:
1) change the time in the flux data if we load old data and use it for a new time (or instruct the paris outputs to do this, which might be better since in some places the flux used is saved, and maybe we don't want to alter that)
2) need to set "time_period" in the flux attributes to "monthly" or "3M" (I think...) for 3 monthly)

hdelongueville:
* **Summary of changes**
Change flux time to start_date when an anterior flux prior is used.

- [x] Closes #258 
- [ ] Tests added and passing

### DIFF
--- a/openghg_inversions/inversion_data/getters.py
+++ b/openghg_inversions/inversion_data/getters.py
@@ -89,8 +89,9 @@ def get_flux_data(
             except SearchError as e:
                 raise SearchError(f"No flux data found before {start_date}") from e
             else:
-                flux_data.data = flux_data.data.isel(time=[-1])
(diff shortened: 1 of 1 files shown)


brendan-m-murphy (approved):
Looks good! Have you tried running it? We might have the right data to do a test fairly easily. If it works we can skip the test for now.

PR #282, opened 2025-05-30
Added offset to PARIS conc outputs

### LINKED ISSUE #280: Add offset to PARIS post-processing
It doesn't look like we output offsets (if the model uses them).

We need to
1. add the offset to the BC for paris outputs
2. report the offset/bias term in paris outputs

brendan-m-murphy:
* **Summary of changes**

The offset is added to the modelled bc before calculating statistics, and the offset is included as Yapost_bias, Yapriori_bias

The function `make_concentration_outputs` in `make_outputs.py` now has an option to combine offset with BC, and will output offset if available (regardless of combining offset with BC)

- [x] Closes #280 
- [x] Tests added and passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # OpenGHG Inversions Change Log
 
+# Unreleased
+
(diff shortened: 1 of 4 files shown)


hdelongueville (approved):
Works well! Just to be sure: will the paris formatting still work without using the offset?

PR #306, opened 2025-08-14
Removed threshold for filling missing obs. error

brendan-m-murphy:
* **Summary of changes**

The threshold was previously 10%, but we would like to fill any value where the obs. error is 0.

Any data where the obs. themselves are NaN will be dropped at a later stage (before the data is passed to PyMC).

- [ ] Tests added and passing

### DIFF
--- a/openghg_inversions/inversion_data/get_data.py
+++ b/openghg_inversions/inversion_data/get_data.py
@@ -101,7 +101,14 @@ def add_obs_error(sites: list[str], fp_all: dict, add_averaging_error: bool = Tr
 
         if err0.any():
             percent0 = 100 * err0.mean()
-            logger.warning("`mf_error` is zero/nan for %.2f percent of times at site %s.", percent0, site)
+            logger.warning(
+                (
+                    "`mf_error` is zero/nan for %.2f percent of times at site %s;"
+                    "filling with max(median(mf_error), std(mf))."
+                ),
+                percent0,
+                site,
+            )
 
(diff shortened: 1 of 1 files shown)


openghg_inversions/inversion_data/get_data.py
```diff
@@ -103,18 +103,17 @@ def add_obs_error(sites: list[str], fp_all: dict, add_averaging_error: bool = Tr
...
+            mf_err_da = ds["mf_error"].as_numpy()  # load into memory to avoid Dask issues
+            fill_value = np.nanmax(
+                [
+                    mf_err_da.where(mf_err_da != 0).dropna(dim="time").median(),
+                    ds["mf"].std(dim="time"),
+                ]
+            )
+            ds["mf_error"] = mf_err_da.where(mf_err_da != 0, fill_value)
+            logger.warning(
+                "More than 10% of `mf_error` is zero/nan, it is thus filled with the max of median `mf_error` and stdev of `mf`."
```
hdelongueville, line 115: Should we change the warning message and combined it with warning from line 104? 
e.g. logger.warning(
    "`mf_error` is zero or NaN for %.2f%% of records at site %s; "
    "replaced with max(median(`mf_error`), std(`mf`)).",
    percent0,
    site
)

brendan-m-murphy: Yea that looks better, I've combined them now.

PR #309, opened 2025-08-19
Change way we infer ouput frequency

alexdanjou:
* **Summary of changes**
When the ini file has start and end date one year apart (annual freq) and the flux has a monthly freq, the freq use for the inversion is set to annual and the flux is resampled at an annual freq (xr.Dataset.resample(time="YS")).

- [ ] Tests added and passing

### DIFF
--- a/openghg_inversions/inversion_data/getters.py
+++ b/openghg_inversions/inversion_data/getters.py
@@ -122,9 +122,10 @@ def get_flux_data(
             or (inferred_time_period_str == "other")
         ):
             flux_data.data.flux.attrs["time_period"] = existing_time_period_str
-        elif "month" in existing_time_period_str.lower():
-            logger.warning("Monthly flux detected, but inversion period is {time_period.days} days. Setting flux time_period to 'monthly'.")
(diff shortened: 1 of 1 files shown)


brendan-m-murphy (commented):
I'm a bit concerned that the postprocessing outputs won't be completely correct: with a monthly flux, the flux is forward filled when it is combined with the footprints, but if we resample here, then the flux used to convert back from the basis functions to the full grid will be wrong.

I suppose we can just open an issue for this and deal with it later if you don't think it will matter much (e.g. if the monthly flux is constant across the year). I didn't think of this before... but it won't be a super easy fix because I think I'd need to change the flux postprocessing (and we would still need some way to specify a yearly inversion period...)

brendan-m-murphy:
Closing as superseded by [#387](https://github.com/openghg/openghg_inversions/pull/387), which handles flux/inversion interval overlap while retaining monthly flux slices, and [#540](https://github.com/openghg/openghg_inversions/pull/540), which preserves authoritative source-period metadata through retrieval and postprocessing. Together these address the use case while avoiding the incorrect reconstruction risk from annually resampling a monthly prior.

PR #327, opened 2025-10-24
adapt for column data

### LINKED ISSUE #323: Adding `"mf_prior"` and `"mf_prior_factor"` variables to inversion outputs of column data
For now, the output of the inversion is the transform obs with which we do the inversions. In order to compare the measurement made with column data with other type of measurement, or column data made using a different number of levels, we need to add back `"mf_prior"` and `"mf_prior_factor"` variables to the obs, the prior and the posterior for the column sites.

Where would it be the most sensible to do it? 
In `make_inv_out_for_fixed_basis_mcmc` seems like reasonnable choice to me. What are your thoughts @brendan-m-murphy ?

alexdanjou:
* **Summary of changes**

This PR updates the inversions workflow to have prior_factor and prior_level_factor to be in the output. Also updates the column workflow of inversions with some added checks related to platform and met_model etc.

- [x] Closes #323 (Replace xxxx with the Github issue number)
- [x] Tests added and passing

### DIFF
--- a/openghg_inversions/basis/algorithms/_quadtree.py
+++ b/openghg_inversions/basis/algorithms/_quadtree.py
@@ -148,10 +148,13 @@ def get_quadtree_basis(fps: np.ndarray, nbasis: int, seed: int | None = None) ->
 
     cost = 1e6
     pwr = 0
+
+    fps = fps/np.sum(fps) # normalize to improve optimization stability
+
     search_max = 10 * np.sum(fps)
     while cost > 3.0:
         optim = scipy.optimize.dual_annealing(
-            qtoptim, np.expand_dims([0, search_max / 10**pwr], axis=0), seed=seed
+            qtoptim, np.expand_dims([0, search_max / 2**pwr], axis=0), seed=seed
         )
         cost = np.sqrt(optim.fun)
         pwr += 1
--- a/openghg_inversions/inversion_data/get_data.py
+++ b/openghg_inversions/inversion_data/get_data.py
@@ -218,10 +218,10 @@ def data_processing_surface_notracer(
         instrument:
             Specific instrument for the site
             (length must match number of sites)
-      max_level:
-        Maximum atmospheric level to extract. Only needed if using 
-        satellite data. Must be an int 
-        calibration_scale:
+        max_level:
+            Maximum atmospheric level to extract. Only needed if using 
+            satellite data. Must be an int 
+            calibration_scale:
             Convert measurements to defined calibration scale
         met_model:
             Meteorological model used in the LPDM. List must be same length as number of sites.
@@ -283,6 +283,7 @@ def data_processing_surface_notracer(
     met_model = convert_to_list(met_model, nsites, "met_model")
     averaging_period = convert_to_list(averaging_period, nsites, "averaging_period")
     platform = convert_to_list(platform, nsites, "platform")
+    max_level = convert_to_list(max_level, nsites, "max_level")
 
     fp_all = {}
     fp_all[".species"] = species.upper()
@@ -349,7 +350,7 @@ def data_processing_surface_notracer(
             average=averaging_period[i],
             instrument=instrument[i],
             calibration_scale=calibration_scale,
-            max_level=max_level,
+            max_level=max_level[i],
             stores=obs_store,
             keep_variables=keep_variables,
         )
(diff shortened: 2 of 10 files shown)


openghg_inversions/basis/algorithms/_quadtree.py
```diff
@@ -148,10 +148,13 @@ def qtoptim(x):
 
     cost = 1e6
     pwr = 0
+
```
brendan-m-murphy, line 151: The changes to quadtree should be removed, they are also present in Ben's open PR.

openghg_inversions/inversion_data/get_data.py
```diff
@@ -143,7 +143,7 @@ def convert_to_list(
         length.
 
     """
-    if x is None or isinstance(x, str):
+    if x is None or isinstance(x, str) or isinstance(x, int):
```
brendan-m-murphy:
```suggestion
    if x is None or isinstance(x, str | int):
```

openghg_inversions/postprocessing/inversion_output.py
```diff
@@ -240,6 +242,9 @@ def __post_init__(self) -> None:
         # format obs data and errors
         self.obs = self.nmeasure_to_site_time(self.obs.rename("y_obs"))
         self.obs_err = self.nmeasure_to_site_time(self.obs_err.rename("y_obs_error"))
+        if self.obs_prior_factor is not None and self.obs_prior_upper_level_factor is not None:
+            self.obs_prior_factor= self.nmeasure_to_site_time(self.obs_prior_factor.rename("y_obs_prior_factor"))
```
brendan-m-murphy: There are missing spaces. Format with black w/ max-line-length 110, that should fix these.

openghg_inversions/postprocessing/inversion_output.py
```diff
@@ -548,7 +566,8 @@ def from_datatree(cls: type[Self], dt: xr.DataTree) -> Self:
 
         """
         obs_and_errs_ds = dt.obs_and_errors.to_dataset().drop_vars(["site", "time"])
-        obs_and_errs = tuple(obs_and_errs_ds.values())
+        obs_and_errs = dict(obs_and_errs_ds)
+        obs_and_errs = {k.replace("y_", "").replace("obs_error","obs_err"): v for k, v in obs_and_errs.items()}
```
brendan-m-murphy: black formatting (missing space)

tests/test_full_inversion.py
```diff
@@ -36,7 +36,7 @@ def satellite_mcmc_args(tmp_path,satellite_ch4_data_args, southamerica_country_f
             "basis_algorithm": "quadtree",
             "basis_output_path": str(tmp_path),
             "nbasis": 4,
-            "nit": 1,
+            "nit": 100,
```
brendan-m-murphy:
```suggestion
            "nit": 1,
```
we don't need this above 1, assuming the error you were facing was due to your environment

brendan-m-murphy (changes requested):
Some formatting needs to be fixed, and I'd like to remove the changes to quadtree, since these have been done elsewhere. (Really should be a separate PR there too... but let's just change it in one place.)

Has this been run on real data for e.g. 2000 iterations? It would be good to see that it actually works, although we could always do a later fixup PR.

openghg_inversions/basis/algorithms/_quadtree.py
SutarPrasad: if this is going ahead can't these go in and Ben's PR will update these?

openghg_inversions/postprocessing/inversion_output.py
SutarPrasad: done

openghg_inversions/basis/algorithms/_quadtree.py
brendan-m-murphy: Yea I guess it actually looks like the merge will be okay between the two. The issue is partly that line 154 needs to be fixed, because np.sum(fps) is just 1, so we should remove that. But yea fine to leave this... I think I'll need to make that change to 154 separately because it will disappear from Ben's diff probably.

brendan-m-murphy (approved):
Okay looks good!

PR #347, opened 2026-01-28
add SAUSSIE and WESTUSA regions

joe-pitt:
* **Summary of changes**
Add a couple of extra regions for the DCE runs

Tests are failing locally for me but the same one (test_quadtree_basis_function) fails for me on devel so let's see what happens here...


joe-pitt:
This should be a straightforward one. @brendan-m-murphy is it normal that the openghg 0.16.0 test fails now?

brendan-m-murphy (commented):
Looks good, the only snag is that @qq23840 has changed this same part of the code in a different PR, in a different way: https://github.com/openghg/openghg_inversions/pull/333/changes#diff-d7146852b9f358b6baedccc4d1dbc5bd676b9612d6df39b0a2a4a7a8083076a6

I guess it might be slightly easier to merge Ben's PR, then add the new domains using the set-up he's introduced.

brendan-m-murphy:
@copilot can you open a PR to merge into this PR. Your PR should resolve the conflicts. Try to adapt the approach in this PR to use the updates from PR #333, which made it easier to add new domains. Try to keep it minimal.

brendan-m-murphy (approved):
Looks good to me! The only weird thing is the name for the Europe outer regions, which requires a separate case; maybe we can fix that later.

PR #359, opened 2026-03-02
[TEMP]Inner domain adaptation

### LINKED ISSUE #388: Adapt Inversions to handle inner domain
Currently we are getting 6km inner domain data for europe and we would like to handle the inversions for the same.

Steps recognised:

1. Ingest standard_domain data for footprints and inner_domain data separately for model scenario.
2. created a fp_data_merge separately for both of them
3. The scenarios are stored in data tree where inner_domain_merged is children of standard_domain_merged.
4. We will also need to consider handling datatree in further functions like
5. `basis_functions_wrapper`, `fp_sensitivity`, `make_inv_inputs`, `fixedbasisMCMC`
6. Ensure the H matrix is calculated for standard-domain and inner_domain individually. We would need to consider masking the values in the standard_domain data which are corresponding to inner domain values. Thus is two avoid duplicating basis function applied on the data.
7. basis function calculated need to have separate data arrays for both in discussion. this requires changes to basis_functions_wrapper

SutarPrasad:
* **Summary of changes**

- [ ] Tests added and passing


openghg_inversions/inversion_data/scenario.py
```diff
@@ -42,4 +101,24 @@ def merged_scenario_data(
...
+        inner_domain_merged = inner_scenario.footprints_data_merge(
+            calc_fp_x_flux=True,
+            calc_bc_sensitivity=False,
+            cache=False,
+            
+        )
+        scenario_combined = scenario_combined.copy()
+
+        # 6km fp_x_flux is added as a separate variable to the combined dataset, and can be merged with the EUROPE fp_x_flux.
+        scenario_combined["fp_x_flux_inner"] = inner_domain_merged["fp_x_flux"]
```
brendan-m-murphy: I think this is still going to force alignment of lat and lon. The inner fp_x_flux needs to be separate from the normal fp_x_flux until after basis functions are applied.

openghg_inversions/basis/_helpers.py
```diff
@@ -63,18 +63,90 @@ def fp_sensitivity(fp_and_data: dict, basis_func: xr.DataArray | dict[str, xr.Da
     if "time" in basis_func.dims and basis_func.sizes["time"] <= 1:
         basis_func = basis_func.squeeze("time")
 
-    fp_and_data[".basis"] = basis_func
+        fp_and_data[".basis"] = basis_func
```
brendan-m-murphy: I don't think this should be indented.

openghg_inversions/basis/_helpers.py
```diff
@@ -63,18 +63,90 @@ def fp_sensitivity(fp_and_data: dict, basis_func: xr.DataArray | dict[str, xr.Da
...
+            sensitivity = apply_fp_basis_functions(
+                fp_x_flux=entry[fp_x_flux_name],
+                basis_func=basis_func,
+            )
+            fp_and_data[site]["H"] = sensitivity
 
     return fp_and_data
 
 
+def combine_inner_outer_fp_x_flux(
```
brendan-m-murphy, line 215: Something like this might be useful for plotting, but we would probably want to interpolate the outer grid lat and lon to match the inner. Anyway, this kind of computation is not what we want for forward modelling.

openghg_inversions/basis/_helpers.py
```diff
@@ -63,18 +63,90 @@ def fp_sensitivity(fp_and_data: dict, basis_func: xr.DataArray | dict[str, xr.Da
...
-            fp_x_flux=fp_and_data[site][fp_x_flux_name],
-            basis_func=basis_func,
-        )
-        fp_and_data[site]["H"] = sensitivity
+        entry = fp_and_data[site]
+
+        # extract root fp_x_flux and mask inner-domain cells to zero if inner domain exists. This ensures that the outer sensitivity H only reflects the outer domain fluxes.
+        if isinstance(entry, xr.DataTree):
+            root_ds = entry.ds
+            fp_x_flux_outer = root_ds[fp_x_flux_name]
```
brendan-m-murphy, line 143: maybe just call this fp_x_flux or fp_x_flux_for_H? It is only outer if the next if statement is true.

openghg_inversions/basis/_helpers.py
SutarPrasad: function not used

SutarPrasad:
Closed with https://github.com/openghg/openghg_inversions/pull/704

Issue #181, opened 2024-07-19
Time dimension for flux data not properly handled.


alexdanjou:
Model scenario from openghg loads all the data available when loading flux. 
No time selection is made, this is not the expected behaviour. We expect to have only the data between the start and end dates.
See for example _mean_fp_times_mean_flux from basis._functions.py.

While resolving this, attention should be paid to small bugs in postprocessing with monthly inversions : the handling of flux data for monthly inversions when the flux data are at on yearly period must be reviewed in inferpymc_postprocessouts from hbmcmc.inversion_pymc.py.
The resolution of issue 126 may resolved this second point.

PR #387, opened 2026-04-02
Iss386 flux timestamps

### LINKED ISSUE #386: Incorrect flux timestamps in monthly/3-monthly inversions
An issue that both @supriyamantri and I have experienced recently is that when running inversions for shorter periods than a year (monthly in my case, 3-monthly in Supriya's), the timestamps in the flux output file are incorrect. For some reason, the dates are sometimes coming out as the start of the calendar year or 3-month period, rather than the middle of the time period.

qq23840:
First attempt at fixing the bug where monthly inversions weren't being assigned the correct timestamps in the flux files. 

Previously, timestamp for the flux file was defined as the start date for the prior flux file plus 2 weeks (for monthly inversions) or 6 months (for yearly inversions). 

This wasn't able to cope with other inversion periods such as 6 months (or indeed 2 years). 

The fix implemented here is to define the flux time as the midpoint of the inversion period (taken as halfway between the start and end dates). 

The problem here is that this won't necessarily give a 'nice' timestamp. For example, for a month with 31 days this will give a timestamp halfway through the 16th day, and for a year this will give midday on the 2nd of July, rather than midnight on the 1st as currently implemented. 

An alternative solution would solve this issue would be to expand (via hard-coding) the `infer_flux_frequency` function to be able to deal with other inversion periods. Then the old 'offset' functionality could be used (e.g. L298 of `make_paris_outputs`), although we would still need to use the `inv_out.start_date` rather than `ds.time.values` to give the correct date from which to start the offset. @brendan-m-murphy and @supriyamantri, what do you think?

(A) [currently implemented] - exact midpoint. Pros - works for any inversion period. Cons - doesn't give 'nice' timestamp (and may not align with InTEM or ELRIS as a result). 

(B) [not implemented] - hard-code inversion intervals including 3-monthly and fix bug to ensure correct start date. 

- [X] Closes #386 
- [x] Tests added and passing

### DIFF
--- a/openghg_inversions/postprocessing/inversion_output.py
+++ b/openghg_inversions/postprocessing/inversion_output.py
@@ -382,7 +382,7 @@ class InversionOutput:
     @property
     def period_midpoint(self) -> pd.Timestamp:
         """Midpoint of inversion period."""
-        return self.start_time + (self.start_time - self.end_time) / 2
+        return self.start_time + (self.end_time - self.start_time) / 2
 
     def get_total_err(self, take_mean: bool = True) -> xr.DataArray:
         """Return the posterior model-data mismatch error.
--- a/openghg_inversions/postprocessing/make_paris_outputs.py
+++ b/openghg_inversions/postprocessing/make_paris_outputs.py
@@ -304,7 +304,7 @@ def paris_flux_output(
             offset = pd.to_timedelta(flux_frequency) / 2
 
         def time_func(ds):
-            return ds.assign_coords(time=(pd.to_datetime(ds.time.values) + offset))
+            return ds.assign_coords(time=[inv_out.period_midpoint])
     else:
 
         def time_func(ds):
--- a/openghg_inversions/postprocessing/make_paris_outputs.py
+++ b/openghg_inversions/postprocessing/make_paris_outputs.py
@@ -236,6 +236,38 @@ def _flux_frequency_to_offset(flux_frequency: str) -> pd.DateOffset | pd.Timedel
         return pd.to_timedelta(flux_frequency)
 
 
+def _flux_interval_midpoints(
+    flux_times: list[pd.Timestamp],
+    flux_period: pd.DateOffset | pd.Timedelta,
+    inv_start: pd.Timestamp,
+    inv_end: pd.Timestamp,
+) -> list[pd.Timestamp]:
+    """Compute output timestamps as midpoints of each flux interval clipped to the inversion period.
+
+    For each flux interval [ft, ft + flux_period], the output timestamp is the
+    midpoint of the overlap with [inv_start, inv_end]. This correctly handles
+    cases where the flux period and inversion period differ in length, e.g.:
+
+    - 3-monthly inversion on yearly fluxes: the yearly flux interval is clipped
+      to the inversion period, so the midpoint is within the inversion period.
+    - 2-yearly inversion on yearly fluxes: each yearly flux interval lies fully
+      within the inversion period, so the midpoint is mid-year as expected.
+
+    Args:
+        flux_times: Start timestamps of each flux interval.
+        flux_period: Duration of a single flux interval.
+        inv_start: Start of the inversion period.
+        inv_end: End of the inversion period.
+
+    Returns:
+        List of midpoint timestamps, one per flux interval.
+    """
+    return [
+        max(ft, inv_start) + (min(ft + flux_period, inv_end) - max(ft, inv_start)) / 2
+        for ft in flux_times
+    ]
+
+
 def paris_flux_output(
     inv_out: InversionOutput,
     country_file: str | Path | None = None,
@@ -312,10 +344,7 @@ def paris_flux_output(
 
         def time_func(ds):
             flux_times = pd.to_datetime(ds.time.values)
-            midpoints = [
-                max(ft, inv_start) + (min(ft + flux_period, inv_end) - max(ft, inv_start)) / 2
-                for ft in flux_times
-            ]
+            midpoints = _flux_interval_midpoints(flux_times, flux_period, inv_start, inv_end)
             return ds.assign_coords(time=midpoints)
     else:
 
--- a/openghg_inversions/postprocessing/make_paris_outputs.py
+++ b/openghg_inversions/postprocessing/make_paris_outputs.py
@@ -226,6 +226,16 @@ def paris_concentration_outputs(
     return result
 
 
+def _flux_frequency_to_offset(flux_frequency: str) -> pd.DateOffset | pd.Timedelta:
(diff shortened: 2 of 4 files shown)


qq23840:
@brendan-m-murphy - sorry, I let this slip last week. Is this an appropriate approach? I mentioned two approaches above, and this one may not be compatible with comparisons in fluxie etc.

openghg_inversions/postprocessing/make_paris_outputs.py
```diff
@@ -304,7 +304,7 @@ def renamer(name: str) -> str:
             offset = pd.to_timedelta(flux_frequency) / 2
 
         def time_func(ds):
-            return ds.assign_coords(time=(pd.to_datetime(ds.time.values) + offset))
+            return ds.assign_coords(time=[inv_out.period_midpoint])
```
brendan-m-murphy: I think this will assign only one value to time; the previous approach was to shift all of the time values by an offset.

openghg_inversions/postprocessing/inversion_output.py
```diff
@@ -382,7 +382,7 @@ def end_time(self) -> pd.Timestamp:
     @property
     def period_midpoint(self) -> pd.Timestamp:
         """Midpoint of inversion period."""
-        return self.start_time + (self.start_time - self.end_time) / 2
+        return self.start_time + (self.end_time - self.start_time) / 2
```
brendan-m-murphy, line 399: This looks correct; I can't believe the old version wasn't causing problems. Maybe it wasn't actually used?

brendan-m-murphy:
@qq23840 There is a `.floor` function for datetimes in pandas, so you could round down to the nearest day.

I think using the exact midpoint will fail for any inversion that has a flux with multiple time points (I think). If you have e.g. a monthly flux, with a year inversion period, then even though you are only optimising for one set of scaling factors, there will be 12 flux time points. In this case, what should the offset be? I would say 2 weeks, since the flux frequency is monthly.

I guess the issue is if you use an annual flux, but only run for JJA, then the start date of the flux is not what you want. In this case, your solution gives the right timestamp.

I think maybe the issue is more precisely: if your inversion period does not fall neatly on the flux time points, then an offset applied to the flux times will be incorrect (most likely).

Monthly (and hourly) fluxes exist for CH4 and CO2, so we need a solution that works in those cases.

Maybe the flux times can be split into periods like:

inversion start, 2nd flux time, 3rd flux time, ..., inversion end

and then we calculate the mid point of each period (e.g. inversion start to 2nd flux time, 2nd flux time to 3rd flux time, ...).

It's possibly that "mid point of inversion period" is just not that well defined here...

qq23840:
OK, that makes sense - I hadn't quite considered the subtlety of yearly inversion over monthly-varying prior fluxes. 

Is an easier bet to just revert to the previous behaviour and add in a few more options for inversion frequencies? E.g. 3monthly, 6monthly, 2yearly, 3yearly? For now?

qq23840:
> OK, that makes sense - I hadn't quite considered the subtlety of yearly inversion over monthly-varying prior fluxes.
> 
> Is an easier bet to just revert to the previous behaviour and add in a few more options for inversion frequencies? E.g. 3monthly, 6monthly, 2yearly, 3yearly? For now?

I've found a more elegant solution, as suggested by copilot. We want the flux timestamp to be the middle of the overlap between the inversion period and the flux time period, in all cases. For example:

- yearly fluxes, 3-month inversion. Overlap is those three months, midpoint of the second month is appropriate. 
- monthly fluxes, yearly inversion. The inversion overlaps with the entirety of each month, so the midpoint of each month is appropriate. 

I've implemented this, and run for my monthly inversions that were failing before. @supriyamantri is testing on her issue over the weekend as well.

supriyamantri:
I’ve now tested this on 3 monthly inversions, including cases that were failing before, and the timestamps are correct across all runs.

supriyamantri (approved):

brendan-m-murphy:
> * yearly fluxes, 3-month inversion. Overlap is those three months, midpoint of the second month is appropriate.
> * monthly fluxes, yearly inversion. The inversion overlaps with the entirety of each month, so the midpoint of each month is appropriate.

Can you ask copilot to make a unit test for these cases? You might need to extract the "time func" from inside the paris flux function into a stand-alone function so it can be tested. I think for the 3-month case, we want to test e.g. start_date is the same as the yearly flux time time, start_date is not the same (this was the problematic case) and probably also start dates not aligned, but end_date is the next year.

For monthly, we probably want to test e.g. start/end dates aligned/not aligned with month starts.

I think we would either need to say for each case, what the expected outcome is, or let copilot make the tests and then check carefully to decide if the "expected" results are really what we want.

brendan-m-murphy:
Here is a related issue from @alexdanjou: https://github.com/openghg/openghg_inversions/issues/181

He says:

> While resolving this, attention should be paid to small bugs in postprocessing with monthly inversions : the handling of flux data for monthly inversions when the flux data are at on yearly period must be reviewed in inferpymc_postprocessouts from hbmcmc.inversion_pymc.py.

tests/test_postprocessing.py
```diff
@@ -46,6 +50,63 @@ def inv_out_eastasia(raw_data_path):
...
+            pd.Timestamp("2019-02-01"),
+            pd.Timestamp("2019-03-01"),
+            [pd.Timestamp("2019-02-01") + (pd.Timestamp("2019-03-01") - pd.Timestamp("2019-02-01")) / 2],
+        ),
+        # 3-monthly inversion, yearly flux: yearly interval clipped to Jan-Apr,
+        # so midpoint is mid-Feb, not mid-year (Jul)
+        (
+            [pd.Timestamp("2019-01-01")],
+            pd.DateOffset(years=1),
+            pd.Timestamp("2019-01-01"),
```
brendan-m-murphy, line 69: I would add a case where the start date is not the same as the flux time, since that was the case that raised this issue. The expected output is still the middle of the three month period, regardless of when it starts.

tests/test_postprocessing.py
```diff
@@ -46,6 +50,63 @@ def inv_out_eastasia(raw_data_path):
     return InversionOutput.load(raw_data_path / "inversion_output_EASTASIA.nc")
 
 
+@pytest.mark.parametrize(
+    "flux_times, flux_period, inv_start, inv_end, expected",
+    [
+        # Monthly inversion, monthly flux: overlap = full month, midpoint = mid-month
```
brendan-m-murphy, line 56: Maybe add a case where the inversion period is several months. The expected output would still be the middle of each month.

brendan-m-murphy (changes requested):
Looks good! I suggested two more test cases.

brendan-m-murphy (approved):
Thanks, looks good! I'm glad we have solid testing here.

Issue #360, opened 2026-03-05
Refactor to use new `BasisFunctions` class


brendan-m-murphy:
# BasisFunctions PR review and refactor sketch

PR #358 added a `BasisFunctions` object that encapsulates basis functions related code:
1. creating a "dummy matrix" with a region coordinate
2. applying this matrix to `fp_x_flux` to create H matrices
3. converting from the "region" dimension back to lat/lon

It also added functionality to support a dictionary of bucket basis functions, for use with multi-source/sector inversion.

Currently, this code is not used in the main pipeline, but it is tested to show that it is equivalent to the helper functions currently used to apply basis functions. My plan for refactoring the current code to use the new `BasisFunctions` class is detailed below; it will be broken into small-ish PRs, with a focus on maintaining parity with existing code.
 
### 1) Retain BasisFunctions object at runtime
 
After `basis_functions_wrapper` executes, retain a `BasisFunctions` object in memory (e.g. under a side-channel
entry on the run context) so downstream code can access:
 
- `operator.sensitivity(...)` for H construction parity,
- `operator.basis_matrix` for flux/country trace calculations,
- `operator.interpolate(...)` for state->grid reconstruction.
 
### 2) Basis save/read compatibility policy (temporary dual format)
 
Because `basis_functions_wrapper` can save basis functions today, and the new operator serialization is DataTree,
the migration should explicitly support both for a deprecation window.
 
Recommended temporary contract:
 
- **Legacy write (default)**: keep existing flat dataset save behavior unchanged.
- **New write (opt-in)**: add a new option to save DataTree serialization (for operator-aware experiments).
- **Read path**:
- if DataTree metadata is present, use it to reconstruct `BasisFunctions`/`BasisOperator`;
- else fall back to legacy flat basis dataset loader.
 
This gives reproducibility for experiments that reuse fixed basis functions across model variants without forcing
an immediate format switch.
 
---
 
### PR-1: Wrapper plumbing + dual-format IO scaffold (no fixedbasisMCMC behavior change)
 
Scope:
 
- Add opt-in return/payload from `basis_functions_wrapper` (non-breaking default).
- Attach basis objects in-memory on `fp_data` side channel.
- Introduce temporary dual write/read support:
- legacy flat format remains default,
- DataTree write/read added behind explicit option/flag.
 
Why this as a standalone PR:
 
- It is self-contained and low-risk.
- It can be validated with focused unit tests even if broader fixedbasisMCMC reviews are light.
 
[...]

PR #390, opened 2026-04-14
Clean up `fixedbasisMCMC` output handling

### LINKED ISSUE #376: fixedbasisMCMC output cleanup
This is PR-2 from the overview issue #360.

PR-2 should focus on cleaning up and separating output code so each part is easier to understand and change.

### PR-2 objective

Refactor `fixedbasisMCMC` output handling so the code paths are clearer, easier to test, and easier to review,
without changing inversion math.

Output handling in `fixedbasisMCMC` should be easier to understand, but existing outputs should not be changed (this should be tested).

## PR-2 in-scope

1. **Separate output responsibilities into explicit stages/functions**
   - Stage A: inversion-run core outputs (trace, sensitivities, metadata).
   - Stage B: basis artifact handling (legacy/datatree path selection + metadata).
   - Stage C: postprocessing invocation and file writing.

2. **Define a small output-contract object/dict** passed between stages
   - e.g. run metadata, paths, enabled output modes, selected basis format.

3. **Standardize where basis-format decisions are made**
   - one place in control flow for `legacy` vs `datatree` decisions.

4. **Keep numerical behavior stable**
   - no intended changes to sampler inputs, `Hx`, `Hbc`, or prior/likelihood calculations.

## PR-2 out-of-scope

- Replacing legacy H computation with `BasisFunctions.operator.sensitivity`.
- Removing legacy basis storage or reconstruction code.
- Making the new operator-based postprocessing path the default (PR-4/5).

## PR-2 tests (structure-preserving)

- Regression test: expected output variables/attrs still present.
- Smoke test: (at least) one fixedbasisMCMC run still writes expected outputs. (If more tests write outputs, they should all continue to pass.)
- No-change checks for key array shapes and coordinate names used by downstream code.

brendan-m-murphy:
* **Summary of changes**

This is an attempt to clean up the output handling in `fixedbasisMCMC`. This PR extracts most of the output logic from `fixedbasisMCMC`, but doesn't really simplify it.

To simplify the output logic we need to:
- remove the legacy `hbmcmc` format; this format can now be created using the new post-processing code
- simplify `make_inv_out_for_fixed_basis_mcmc`, and possibly make `InversionOutput` simpler too
- split some debugging options from the main script (e.g. if you just want to make/load/save merged data, that probably doesn't need to go through `fixedbasisMCMC`)

I think those changes need to happen in a separate PR. I will update the umbrella issue #360 with a new plan.

Another issue that needs to be addressed before completing #360 is how basis functions are routed to the postprocessing code. Currently the flat basis functions are passed via `fp_all`.

- [x] Closes #376
- [x] Tests added and passing
- Documentation and tutorials updated/added
- Added any new requirements to `requirements.txt`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - Stage B of PyMC model refactor. Updated `inferpymc` to accept current/legacy inputs as well as xarray `Dataset`. [PR #380](https://github.com/openghg/openghg_inversions/pull/380)
 - Stage C of PyMC model refactor. Added function for building PyMC "model components". The model building code from Stage B is still used by default, but the new code can be selected by adding `model_builder="components"` to the .ini file. [PR #382](https://github.com/openghg/openghg_inversions/pull/382)
 - Stage D of PyMC model refactor. Removed temporary scaffolding to preserve legacy model building code. `inferpymc` now only accepts inversion inputs as `xr.Dataset`, and `fixedbasisMCMC` has been updated to reflect this. [PR #389](https://github.com/openghg/openghg_inversions/pull/389)
+- Clean up of `fixedbasisMCMC` output handling. [PR #390](https://github.com/openghg/openghg_inversions/pull/390)
 
 # Version 0.6.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@
 - Stage B of PyMC model refactor. Updated `inferpymc` to accept current/legacy inputs as well as xarray `Dataset`. [PR #380](https://github.com/openghg/openghg_inversions/pull/380)
 - Stage C of PyMC model refactor. Added function for building PyMC "model components". The model building code from Stage B is still used by default, but the new code can be selected by adding `model_builder="components"` to the .ini file. [PR #382](https://github.com/openghg/openghg_inversions/pull/382)
 - Stage D of PyMC model refactor. Removed temporary scaffolding to preserve legacy model building code. `inferpymc` now only accepts inversion inputs as `xr.Dataset`, and `fixedbasisMCMC` has been updated to reflect this. [PR #389](https://github.com/openghg/openghg_inversions/pull/389)
-- Clean up of `fixedbasisMCMC` output handling. [PR #390](https://github.com/openghg/openghg_inversions/pull/390)
+- Neutral refactor of `fixedbasisMCMC` output handling to make the end-of-run logic clearer, whichis now split into explicit stages for artefact creation, `InversionOutput` construction, and output mode dispatch. [PR #390](https://github.com/openghg/openghg_inversions/pull/390)
 
 # Version 0.6.0
 
--- a/openghg_inversions/hbmcmc/hbmcmc.py
+++ b/openghg_inversions/hbmcmc/hbmcmc.py
@@ -45,7 +45,10 @@ from openghg_inversions.inversion_data import (
     load_merged_data,
 )
 from openghg_inversions.filters import filtering
-from openghg_inversions.postprocessing.inversion_output import InversionOutput, make_inv_out_for_fixed_basis_mcmc
+from openghg_inversions.postprocessing.inversion_output import (
+    InversionOutput,
+    make_inv_out_for_fixed_basis_mcmc,
+)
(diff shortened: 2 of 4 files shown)


openghg_inversions/hbmcmc/hbmcmc.py
```diff
@@ -208,6 +210,355 @@ def _inv_inputs_from_rerun_arrays(
...
+
+def _resolve_inversion_output_path(
+    save_inversion_output: str | Path | bool, outputpath: str, outputname: str, start_date: str
+) -> Path | None:
+    """Resolve the output path for a saved inversion-output file."""
+    if not save_inversion_output:
+        return None
+    if isinstance(save_inversion_output, str | Path):
+        return Path(save_inversion_output)
+    return Path(outputpath) / (outputname + f"{start_date}_inversion_output.nc")
```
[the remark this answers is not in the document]
brendan-m-murphy: That is incorrect: support for union types in `isinstance` was added in Python 3.10 https://peps.python.org/pep-0604/ https://docs.python.org/3/library/stdtypes.html#types-union

openghg_inversions/hbmcmc/hbmcmc.py
```diff
@@ -208,6 +210,355 @@ def _inv_inputs_from_rerun_arrays(
...
+    if paris_postprocessing is True:
+        output_format = "paris"
+        warnings.warn(
+            "The `paris_postprocessing` argument will be deprecated. Use `output_format = 'paris'` instead."
+        )
+
+    resolved_output_format = output_format.lower()
+
+    if resolved_output_format == "hbmcmc" and is_sat_column:
+        raise ValueError(
```
SutarPrasad, line 256: Do we want this to be extended to site based column data?

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 - Stage B of PyMC model refactor. Updated `inferpymc` to accept current/legacy inputs as well as xarray `Dataset`. [PR #380](https://github.com/openghg/openghg_inversions/pull/380)
 - Stage C of PyMC model refactor. Added function for building PyMC "model components". The model building code from Stage B is still used by default, but the new code can be selected by adding `model_builder="components"` to the .ini file. [PR #382](https://github.com/openghg/openghg_inversions/pull/382)
 - Stage D of PyMC model refactor. Removed temporary scaffolding to preserve legacy model building code. `inferpymc` now only accepts inversion inputs as `xr.Dataset`, and `fixedbasisMCMC` has been updated to reflect this. [PR #389](https://github.com/openghg/openghg_inversions/pull/389)
+- Clean up of `fixedbasisMCMC` output handling. [PR #390](https://github.com/openghg/openghg_inversions/pull/390)
```
SutarPrasad: needs extended version giving better idea about the change

SutarPrasad (commented):
looks good, should be ready after discussing the comment and adding liitle bit more of information to changelog.

openghg_inversions/hbmcmc/hbmcmc.py
brendan-m-murphy: Possibly. I don't think there was logic to check that before. We do:

```
if inlet is not None:
         is_sat_column = any([i == "column" for i in inlet])
     else:
         is_sat_column = False
```

but there is no site column check. The ValueError raised on line 256 was previously inside `fixedbasisMCMC`.

CHANGELOG.md
brendan-m-murphy: Okay that is done.

openghg_inversions/hbmcmc/hbmcmc.py
brendan-m-murphy: I think maybe `is_sat_column` could just be renamed to `-is_column` and the message could be updated. To know if it is satellite specifically we'd have to check `platform` right?

SutarPrasad: yes renaming and message change should suffice, this essentially means hbmcmc will not be supported for any column data. Do we have any inputs about this from Anita or Matt? I guess everything is now worked under "paris" so should be ok?

SutarPrasad (approved):

PR #395, opened 2026-04-21
Add option to drop missing vars from gather-concat

### LINKED ISSUE #394: Concat-gather bug
# Notes

In `make_inv_inputs`, the call to `concat_gather_datasets` raises a ValueError because one of the sites has an `inlet_height` data variable, which is missing from another site.

The issue is most likely that we call `concat_gather_datasets` before we select the data variables we need for the inversion. If that is true, we can just swap the steps to fix this.

brendan-m-murphy:
* **Summary of changes**

The `gather_concat_datasets` function that is used to combined scenarios from multiple sites previously raised an error if one site had a data variable that was missing from another site.

This PR adds an option to drop missing variables in this case.

Tests were added to check the new functionality.

This closes #394.

- [x] Closes #384
- [x] Tests added and passing
- Documentation and tutorials updated/added
- Added any new requirements to `requirements.txt`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ## Code changes
 
 
+- Made concat-gather handling of mismatched site data variables order-independent, added an opt-in drop policy used by `make_inv_inputs`, and added lightweight regression tests for issue #394. [#394](https://github.com/openghg/openghg_inversions/issues/394)
 - Fix bug which was assigninig the wrong times to inversion flux outputs in non-standard cases, such as 3-monthly inversions. [#PR 387](https://github.com/openghg/openghg_inversions/pull/387)
 - Fix small bug where postprocessing was failing if country codes in file didn't match exactly those in `paris_regions_dict`. [#PR 377](https://github.com/openghg/openghg_inversions/pull/377)
 - More flexibility for new inversion domains. [#PR 333](https://github.com/openghg/openghg_inversions/pull/333)
--- a/openghg_inversions/array_ops.py
+++ b/openghg_inversions/array_ops.py
@@ -471,6 +471,7 @@ def _resolve_shared_data_vars(
             stacklevel=2,
         )
 
+    # Keep only shared vars, but preserve the first dataset's variable order.
     first_dataset = next(iter(ds_dict.values()))
     return [dv for dv in first_dataset.data_vars if dv in intersection]
 
--- a/openghg_inversions/array_ops.py
+++ b/openghg_inversions/array_ops.py
@@ -42,7 +42,7 @@ an xarray issue, which now is written in terms of ``force_align``.
 from __future__ import annotations
 
 from collections.abc import Hashable, Iterable, Mapping, Sequence
-from typing import Any, overload, TypeVar
+from typing import Any, Literal, overload, TypeVar
 import warnings
 
 from dask.array.core import Array as DaskArray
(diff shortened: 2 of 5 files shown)


supriyamantri:
I tested the fix, but it’s still not working—I’m getting the same error and interestingly only for years 2020, 2021 and 2023. I’ve attached the tracer, the .ini file, the error message, and the output file of 2020 for the reference.
[untitled folder.zip](https://github.com/user-attachments/files/27007351/untitled.folder.zip)

brendan-m-murphy:
> I tested the fix, but it’s still not working—I’m getting the same error and interestingly only for years 2020, 2021 and 2023. I’ve attached the tracer, the .ini file, the error message, and the output file of 2020 for the reference. [untitled folder.zip](https://github.com/user-attachments/files/27007351/untitled.folder.zip)

Were you getting the error message for other years before? Can you just paste the traceback here?

supriyamantri:
> > I tested the fix, but it’s still not working—I’m getting the same error and interestingly only for years 2020, 2021 and 2023. I’ve attached the tracer, the .ini file, the error message, and the output file of 2020 for the reference. [untitled folder.zip](https://github.com/user-attachments/files/27007351/untitled.folder.zip)
> 
> Were you getting the error message for other years before? Can you just paste the traceback here?

Previously, I was running the inversion using the old obs store
/group/chem/acrg/object_stores/paris/obs_paris_2025_07_storee, and the inversion was working fine for all years there.
I’ve now switched to the new obs store
acrg/object_stores/paris/obs_nid_2026_store, and with this one I consistently get errors for the years 2020, 2021, and 2023. For these years, the tracer seems to be empty (see attached screenshot Screenshot 2026-04-23 at 11 51 24).

brendan-m-murphy:
Sorry I don't mean the trace, I mean the error traceback, which is probably in the .err or .out log files from slurm, or would just be printed in a notebook or if you ran the script on the login node.

It is a sequence of steps in the code that lead to the error that you get. There is more info and some examples here: https://realpython.com/python-traceback/ one example they show looks like:

```bash
$ python example.py
Traceback (most recent call last):
  File "/path/to/example.py", line 4, in <module>
    greet('Chad')
  File "/path/to/example.py", line 2, in greet
    print('Hello, ' + someon)
NameError: name 'someon' is not defined
```

supriyamantri:
> Sorry I don't mean the trace, I mean the error traceback, which is probably in the .err or .out log files from slurm, or would just be printed in a notebook or if you ran the script on the login node.
> 
> It is a sequence of steps in the code that lead to the error that you get. There is more info and some examples here: https://realpython.com/python-traceback/ one example they show looks like:
> 
> ```shell
> $ python example.py
> Traceback (most recent call last):
>   File "/path/to/example.py", line 4, in <module>
>     greet('Chad')
>   File "/path/to/example.py", line 2, in greet
>     print('Hello, ' + someon)
> NameError: name 'someon' is not defined
> ```

Ahh, my bad - thought you are asking about trace. Here is the error message screenshot 
Screenshot 2026-04-23 at 12 14 13

brendan-m-murphy:
I think you need to checkout the branch that this PR is on. The line numbers in your traceback don't match the line numbers in the code on this PR. From the openghg inversions repo, you can do `git fetch origin` and `git switch concat-gather-bug` (I think, maybe try `git checkout` instead of `git switch` if that doesn't work).

supriyamantri:
You’re absolutely right, thanks for pointing that out. I was using git switch to change branches, but it turned out that concat-gather-bug wasn’t actually the branch being used when I ran the inversion.
Once I explicitly ran git checkout concat-gather-bug and reran the inversion, everything worked as expected. Thanks!

supriyamantri (approved):

PR #418, opened 2026-04-29
fix persisting bug about flux timestamps

### LINKED ISSUE #417: bug in inversion output timestamps not fixed
Unfortunately, I've spotted a bug from where we 'fixed' the inversion timestamps issue a couple of weeks ago (#387) . Essentially, although we were calculating the correct timestamps to output, the actual flux data wasn't filtered to match those timestamps, and so there were still erroneous timestamps in the case that a multi-year annual flux was being used in a single-year inversion. 

I've attempted a fix and will PR soon.

qq23840:
Updated the flux postprocessing code to actually filter the data according to the correct timestamps, which wasn't happening before. I don't think this will have materially impacted too many inversions, but it's worth checking for erroneous timestamps in anything you've run in the last week or so @supriyamantri 

- [x] Closes #417 
- [x] Tests added and passing

### DIFF
--- a/openghg_inversions/postprocessing/make_paris_outputs.py
+++ b/openghg_inversions/postprocessing/make_paris_outputs.py
@@ -260,12 +260,17 @@ def _flux_interval_midpoints(
         inv_end: End of the inversion period.
 
     Returns:
-        List of midpoint timestamps, one per flux interval.
+        List of midpoint timestamps, one per flux interval (only for periods overlapping the inversion period).
     """
-    return [
-        max(ft, inv_start) + (min(ft + flux_period, inv_end) - max(ft, inv_start)) / 2
-        for ft in flux_times
-    ]
+    midpoints = []
+    for ft in flux_times:
+        overlap_start = max(ft, inv_start)
+        overlap_end = min(ft + flux_period, inv_end)
+        # Only include midpoint if there is valid overlap
+        if overlap_end > overlap_start:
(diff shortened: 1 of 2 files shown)


openghg_inversions/postprocessing/make_paris_outputs.py
```diff
@@ -345,7 +350,15 @@ def renamer(name: str) -> str:
         def time_func(ds):
             flux_times = pd.to_datetime(ds.time.values)
             midpoints = _flux_interval_midpoints(flux_times, flux_period, inv_start, inv_end)
-            return ds.assign_coords(time=midpoints)
+            # Find which time indices have valid overlap (those included in midpoints)
```
brendan-m-murphy: Maybe the function that returns midpoints can also return the indices. Or maybe you can use sel instead of isel, maybe with method "bfill" and limit=1? Duplicating the loop with minor variations seems bad because those two loops need to stay in sync if we make future changes. Another option is to just make the helper function that creates "midpoints" just do the selection and coordinate assignment and then just do time_func = midpoint_helper_function (maybe with a better name). Python has "first class functions" so functions are just objects you can assign like anything else.

brendan-m-murphy (commented):
Looks good but maybe we can remove some duplicate code to make maintenance easier.

brendan-m-murphy (approved):
Looks good, thanks!

PR #422, opened 2026-05-05
Fix shared sigma indexing when site indicators are already registered

brendan-m-murphy:
## Summary
- Avoid reusing the existing `site_indicator` model data when `sigma_per_site=False`
- Register a separate zero-valued site index for shared sigma so prior predictive sampling does not index out of bounds
- Add a regression test that covers `add_offset=True` with shared sigma and prior predictive sampling

Closes #421 

## Testing
- `pytest tests/models/test_components.py -k shared_sigma`
- `pytest tests/models/test_components.py`
- `pytest tests/test_inferpymc.py`
- `pytest`
- `black --check` and `ruff check` passed for the touched files

### DIFF
--- a/openghg_inversions/models/components.py
+++ b/openghg_inversions/models/components.py
@@ -252,7 +252,8 @@ def add_sigma_component(
         )
 
     site_data = site_indicator if per_site else xr.zeros_like(site_indicator)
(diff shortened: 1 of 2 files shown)


supriyamantri (approved):
Tested and the inversion runs without any issues.

PR #433, opened 2026-05-28
Split fixedbasis and RHIME preparation contracts

brendan-m-murphy:
## Summary
- Delete `PreparedInversionData` and replace the mixed preparation contract with `FixedBasisPreparedData` and frozen `RhimePreparedInputs`.
- Split fixedbasis preparation from RHIME preparation; RHIME now builds `H` directly via `BasisFunctions.sensitivity()` without exposing fixedbasis side channels.
- Keep legacy `.basis` / `.flux` behavior confined to fixedbasis and output/postprocessing boundaries, and simplify review-found cruft after the initial implementation.

Closes #431

## Tests
- `uv run pytest tests/test_rhime.py tests/test_basis_functions.py tests/test_postprocessing.py`
- `uv run ruff check openghg_inversions/inversion_data/preparation.py openghg_inversions/rhime.py openghg_inversions/basis/_wrapper.py openghg_inversions/basis/__init__.py openghg_inversions/hbmcmc/hbmcmc.py tests/test_rhime.py tests/test_postprocessing.py`
- `uv run pyright openghg_inversions/basis/_wrapper.py openghg_inversions/inversion_data/preparation.py openghg_inversions/rhime.py openghg_inversions/hbmcmc/hbmcmc.py`

### DIFF
--- a/openghg_inversions/hbmcmc/hbmcmc.py
+++ b/openghg_inversions/hbmcmc/hbmcmc.py
@@ -26,7 +26,6 @@ the users OpenGHG config file (default location: ~/.openghg/openghg.conf).
 
 import logging
 from dataclasses import dataclass, field
-from importlib import import_module
 from pathlib import Path
 import time
 from typing import Literal, cast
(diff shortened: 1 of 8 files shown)


openghg_inversions/hbmcmc/hbmcmc.py
```diff
@@ -34,12 +35,15 @@
 import numpy as np
 import xarray as xr
 
-from openghg.util import split_function_inputs
+try:
```
brendan-m-murphy: I don't think this is necessary. If this is cover for an old version of openghg we should adjust the lower bound pin on openghg because this has been exported by util for quite a while.

openghg_inversions/rhime.py
```diff
@@ -4,22 +4,27 @@
...
 import time
 import warnings
 
 import arviz as az
 import numpy as np
 import pymc as pm
 import xarray as xr
-from openghg.util import split_function_inputs
+
+try:
```
brendan-m-murphy: Same as above, just leave the original import unless it is truly necessary, but I would rather pin a lower bound for openghg because we generally need more recent releases of openghg

brendan-m-murphy (commented):
Generally looks good, I don't like the complicated import from openghg.util though. We probably should pin a more recent version of openghg if necessary.

PR #472, opened 2026-06-09
Codex/405 paris sector vars

### LINKED ISSUE #405: Create sector-aware RHIME outputs and PARIS-compatible total outputs
## Goal

Convert modern multi-sector `InferenceData` results into outputs that preserve sectors while still
producing production PARIS outputs.

## Tasks

- Reconstruct sector-specific posterior flux traces or summaries outside the PyMC graph.
- Add total posterior flux by summing sector posterior flux when all sectors share the same basis.
- Preserve sector-specific modelled-observation contributions where practical.
- Route total outputs through PARIS output creation.
- If PARIS cannot carry a `sector` coordinate, emit a PARIS-compliant total output plus a separate
  sector diagnostic NetCDF.

## Acceptance criteria

- Output contains sector posterior flux and total posterior flux.
- Total flux equals the sum of sectors for a synthetic fixture.
- PARIS flux and concentration outputs work for total outputs.
- Sector diagnostics are saved or returned in a documented format.

brendan-m-murphy:
## Summary

- Adds latest-PARIS multisector per-sector flux variables alongside the required total variables, including suffix-safe expansion of the CDL sector placeholders.
- Projects retained basis-region traces directly to country totals for each sector, preserving lazy Dask/sparse execution until statistics or NetCDF serialization require dense chunks.
- Adds total, within-sector, and between-sector posterior covariance diagnostics. The total covariance is emitted for both single-sector and multisector outputs and uses the same population estimator as the reported standard deviations.
- Keeps the latest EUROPE template's canonical 22-country list as the public default. Explicit `country_selections=None` opts into all countries in the domain file, and explicit names or codes are normalized to alpha-3 labels before projection.
- Preserves CF `units` and `calendar` attributes on bounds variables through a shared NetCDF writer used by RHIME and fixed-basis PARIS outputs.
- Adds a marked end-to-end integration test that runs two equal flux inputs through the real multisector data preparation, PyMC sampling, postprocessing, NetCDF write, and reopen path. It validates totals against sectors and compares dimensions, variables, dtypes, fill values, coordinates, and required attributes with the flux v03 CDL template.
- Keeps output-specific tests under `tests/postprocessing` and updates the issue #405 planning note with the implemented decisions and remaining concentration-output design work.

Closes #405.

## Checks

- [x] `tox -p` (OpenGHG current, previous minor, and `devel`, plus Ruff)
- [x] `.venv/bin/pytest tests/postprocessing/test_make_outputs.py tests/postprocessing/test_make_paris_outputs.py tests/test_rhime.py tests/test_postprocessing.py -q` (145 passed)
- [x] `.venv/bin/pytest -m slow tests/integration/test_multisector_paris_pipeline.py -q` (1 passed)
- [x] Ruff check and format check on all changed Python files
- [x] `git diff --check`

## Review

Subagent passes covered pipeline mapping, scientific-array correctness, test construction, docstrings, and final PR review. Their findings on country defaults and normalization, direct lazy country projection, single-sector covariance, bounds metadata, and fixed template dimensions are addressed in this branch. The final re-review found no remaining blockers.

## Notes

The PARIS CDL schema intentionally uses duplicate country and sector dimension names for covariance variables. Xarray warns when reopening those variables, so the integration test also validates their raw NetCDF dimensions directly.

## Template checklist

- [x] Closes #405
- [x] Tests added and passing
- [x] Documentation updated
- [x] Changelog entry not required for this unreleased feature branch
- [x] No new requirements

### DIFF
--- a/openghg_inversions/postprocessing/make_paris_outputs.py
+++ b/openghg_inversions/postprocessing/make_paris_outputs.py
@@ -866,52 +866,35 @@ def _latest_paris_countries(country_file: str | Path | None, domain: str) -> Cou
     return countries
 
 
-def _multisector_country_trace_kg(inv_out: InversionOutput, countries: Countries, species: str) -> xr.Dataset:
-    """Return multisector total country flux traces in kg/yr from reconstructed total flux."""
-    total_flux_trace = make_multisector_flux_trace_outputs(
-        inv_out,
-        report_flux_on_inversion_grid=False,
-    )[["flux_total_prior", "flux_total_posterior"]].rename(
-        {
-            "flux_total_prior": "country_prior",
-            "flux_total_posterior": "country_posterior",
-        }
-    )
-    country_weights = countries.matrix.as_numpy() * countries.area_grid.as_numpy()
-    country_weights = align_sparse_lat_lon(country_weights, total_flux_trace["country_posterior"])
-    country_trace = sparse_xr_dot(country_weights, total_flux_trace, dim=["lat", "lon"])
-    country_trace = country_trace * 365 * 24 * 3600 * convert.molar_mass(species) * 1e-3
-    return country_trace.reindex(country=list(PARIS_LATEST_COUNTRIES))
-
-
-def _multisector_sector_country_trace_kg(
+def _multisector_country_trace_kg(
     inv_out: InversionOutput,
     countries: Countries,
     species: str,
     sector_name_by_suffix: Mapping[str, str],
+    flux_trace: xr.Dataset | None = None,
 ) -> xr.Dataset:
-    """Return multisector per-sector country flux traces in kg/yr."""
-    flux_trace = make_multisector_flux_trace_outputs(
-        inv_out,
-        report_flux_on_inversion_grid=False,
-    )
-    rename: dict[str, str] = {}
-    keep_vars = []
+    """Reconstruct once and lazily reduce total and sector flux traces to countries."""
+    if flux_trace is None:
+        flux_trace = make_multisector_flux_trace_outputs(
+            inv_out,
+            report_flux_on_inversion_grid=False,
+            materialize=False,
+        )
+    rename = {
+        "flux_total_prior": "country_prior",
+        "flux_total_posterior": "country_posterior",
+    }
     for variable_suffix, sector_name in sector_name_by_suffix.items():
         for when in ("prior", "posterior"):
             source_name = f"flux_{variable_suffix}_{when}"
             if source_name in flux_trace:
-                keep_vars.append(source_name)
                 rename[source_name] = f"country_{sector_name}_{when}"
 
-    if not keep_vars:
-        return xr.Dataset()
-
-    sector_flux_trace = flux_trace[keep_vars].rename(rename)
-    country_weights = countries.matrix.as_numpy() * countries.area_grid.as_numpy()
-    reference_flux = next(iter(sector_flux_trace.data_vars.values()))
+    selected_flux_trace = flux_trace[list(rename)].rename(rename)
+    country_weights = countries.matrix * countries.area_grid
+    reference_flux = selected_flux_trace["country_posterior"]
     country_weights = align_sparse_lat_lon(country_weights, reference_flux)
-    country_trace = sparse_xr_dot(country_weights, sector_flux_trace, dim=["lat", "lon"])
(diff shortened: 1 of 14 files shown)


openghg_inversions/postprocessing/make_paris_outputs.py
```diff
@@ -780,20 +866,35 @@ def _latest_paris_countries(country_file: str | Path | None, domain: str) -> Cou
...
-        }
-    )
-    country_weights = countries.matrix.as_numpy() * countries.area_grid.as_numpy()
-    country_weights = align_sparse_lat_lon(country_weights, total_flux_trace["country_posterior"])
-    country_trace = sparse_xr_dot(country_weights, total_flux_trace, dim=["lat", "lon"])
+def _multisector_country_trace_kg(
+    inv_out: InversionOutput,
+    countries: Countries,
+    species: str,
+    sector_name_by_suffix: Mapping[str, str],
```
brendan-m-murphy, line 890: Use a full google-style docstring since the form of this mapping is not obvious

openghg_inversions/postprocessing/make_paris_outputs.py
```diff
@@ -780,20 +866,35 @@ def _latest_paris_countries(country_file: str | Path | None, domain: str) -> Cou
...
+            if source_name in flux_trace:
+                rename[source_name] = f"country_{sector_name}_{when}"
+
+    selected_flux_trace = flux_trace[list(rename)].rename(rename)
+    country_weights = countries.matrix * countries.area_grid
+    reference_flux = selected_flux_trace["country_posterior"]
+    country_weights = align_sparse_lat_lon(country_weights, reference_flux)
+    country_trace = sparse_xr_dot(country_weights, selected_flux_trace, dim=["lat", "lon"])
     country_trace = country_trace * 365 * 24 * 3600 * convert.molar_mass(species) * 1e-3
     return country_trace.reindex(country=list(PARIS_LATEST_COUNTRIES))
```
brendan-m-murphy: Is this reindexing specific to the "EUROPE" domain? There were recent efforts to avoid making the PARIS outputs Europe-specific. (The PARIS project is EU+UK specific, I think, but the format is used on other projects that do not use the Europe domain.)

openghg_inversions/postprocessing/make_paris_outputs.py
```diff
@@ -780,20 +866,35 @@ def _latest_paris_countries(country_file: str | Path | None, domain: str) -> Cou
...
+        for when in ("prior", "posterior"):
+            source_name = f"flux_{variable_suffix}_{when}"
+            if source_name in flux_trace:
+                rename[source_name] = f"country_{sector_name}_{when}"
+
+    selected_flux_trace = flux_trace[list(rename)].rename(rename)
+    country_weights = countries.matrix * countries.area_grid
+    reference_flux = selected_flux_trace["country_posterior"]
+    country_weights = align_sparse_lat_lon(country_weights, reference_flux)
+    country_trace = sparse_xr_dot(country_weights, selected_flux_trace, dim=["lat", "lon"])
```
brendan-m-murphy: Are we certain this computation is lazy? Ideally we would not materialise the full posterior flux traces; instead, we want to create a map from basis functions/regions to countries, then map the scaling factor tracers.

openghg_inversions/postprocessing/make_paris_outputs.py
```diff
@@ -846,12 +964,82 @@ def _country_posterior_covariance_kg(
...
+    values = posterior.transpose("flux_time", "country", "draw").values
+    if values.shape[2] == 0:
         return np.full((values.shape[0], values.shape[1], values.shape[1]), np.nan, dtype="float32")
 
     centered = values - values.mean(axis=2, keepdims=True)
-    return np.einsum("tcd,ted->tce", centered, centered) / (values.shape[2] - 1)
+    return np.einsum("tcd,ted->tce", centered, centered) / values.shape[2]
+
+
+def _sector_country_posterior_covariances_kg(
```
brendan-m-murphy, line 1030: Do we have something equivalent for the single-sector outputs? I hadn't considered including this in the past, but our basis regions can cross country boundaries, so even with independent scaling factors, there can be covariance here.

brendan-m-murphy (commented):
I cannot test this on real data currently, but the code I read seems to be doing the right thing, and there is an end-to-end test that passes locally (skipped on CI).

PR #491, opened 2026-07-05
Add grouped basis layout metadata core

brendan-m-murphy:
* **Summary of changes**

Adds the first grouped basis layout implementation slice for OGI-023:
- adds an internal `BasisLayout` / `BasisPartition` helper that combines disjoint partition-local labels into one flat positive-label map plus raw-label keyed metadata;
- teaches `BucketBasisOperator` to attach `basis_group`, `basis_partition`, and `region_in_partition` after its final state-coordinate label policy is applied;
- persists that optional state metadata through operator DataTree and `BasisFunctions.save/load` roundtrips;
- keeps legacy `fixed_outer_regions_basis`, `run_hbmcmc.py`, and the existing `quadtree` / `weighted` option surface unchanged.

This tracks #457 / #449 as a focused OGI-023 implementation slice. It intentionally does not close #457; disconnected-part reporting, geography-aware cleanup, and RHIME/config routing remain follow-up work.

- [ ] Closes #xxxx (not closing #457; partial OGI-023 implementation slice)
- [x] Tests added and passing
- [ ] Documentation and tutorials updated/added (internal docstrings only; no user-facing option added)

Validation:
- `uv run pytest tests/test_basis_layout.py`
- `uv run pytest tests/test_basis_functions.py -k "datatree or fixed_outer"`
- `uv run pytest tests/test_constrained_basis_algorithm.py -k "intersect or tuple or constrained"`
- `uv run ruff check openghg_inversions/basis/layout.py openghg_inversions/basis/operators.py tests/test_basis_layout.py`
- `tox -p`

### DIFF
--- a/openghg_inversions/basis/layout.py
+++ b/openghg_inversions/basis/layout.py
@@ -7,6 +7,10 @@ algorithms decide how to split cells inside a mask or region class, while
 ``BasisLayout`` records how those partition-local outputs are assembled into the
 single state axis consumed by ``BucketBasisOperator``.
 
+``BasisLayout`` does not infer a catch-all remainder region. Callers that need
+one should pass it as an explicit ``BasisPartition`` so the semantic group and
+local region label are visible in the resulting state metadata.
+
 The current implementation is eager and in-memory. It is intended for small
 metadata/layout assembly steps after region masks and basis labels have already
 been loaded or generated. File loading, lazy execution, large-grid chunking, and
@@ -26,6 +30,7 @@ BASIS_LABEL_DIM = "basis_label"
 BASIS_GROUP_COORD = "basis_group"
 BASIS_PARTITION_COORD = "basis_partition"
 REGION_IN_PARTITION_COORD = "region_in_partition"
+STATE_METADATA_COORDS = (BASIS_GROUP_COORD, BASIS_PARTITION_COORD, REGION_IN_PARTITION_COORD)
 
 
 @dataclass(frozen=True, slots=True)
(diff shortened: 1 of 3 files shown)


openghg_inversions/basis/layout.py
```diff
@@ -0,0 +1,220 @@
...
+    """Combine disjoint basis partitions into one flat label map.
+
+    This first implementation is intentionally eager and in-memory: partition
+    label arrays are materialized as NumPy arrays while building the combined
+    map. File loading, lazy execution, and large-grid chunking policy remain
+    outside this small metadata/layout boundary.
+
+    Attributes:
+        partitions: Ordered partition definitions to combine. Partitions must
+            cover disjoint grid cells and the combined layout must cover the
```
brendan-m-murphy, line 249: Would it make sense to have a "remainder" category here? Or should that be determined upstream?

openghg_inversions/basis/operators.py
```diff
@@ -465,6 +525,129 @@ def _apply_region_labels_policy(self, mat: xr.DataArray) -> xr.DataArray:
         # ordered by sorted unique labels. If that assumption changes, we must reindex.
         return mat.assign_coords({self.meta.state_dim: coord})
 
+    def _assign_state_metadata(
```
brendan-m-murphy: Are these state/metadata related methods best as part of the basis operator? There seems to be a lot of coordination with module level constants from layout.py; perhaps a class defined in layout.py would be better? Alternatively, such a class could live in this module, like `BasisMeta`. And actually, there is some overlap with `BasisMeta`: the `state_metadata` enhances what `BasisMeta` provides.

brendan-m-murphy (commented):
I think the state metadata helper methods in `BucketBasisOperator` might be partially overlapping with the role of `BasisMeta`. If possible, expand `BasisMeta` (or subclass it) to include expanded state vector metadata.

PR #499, opened 2026-07-08
Sanitize non-finite flux before retained output reconstruction

brendan-m-murphy:
* **Summary of changes**

Closes #495 by treating retained flux as another surface of the same non-finite flux root cause that PR #474 fixed for generated-basis weights.

- Adds `sanitize_flux_nonfinite()` for lazy zero-fill sanitation of `NaN`, `+Inf`, and `-Inf` flux values, with optional count-mode auditing.
- Records machine-readable sanitation metadata through `FluxNonFiniteMetadata`, serialized into one namespaced JSON attr: `openghg_inversions:non_finite_flux`.
- Keeps that metadata contract local and canonical. It does not guess future OpenGHG attribute names; openghg/openghg#1680 tracks definition of an upstream contract.
- Applies sanitation during flux ingestion/preparation, merged-data recovery, retained `BasisFunctions` flux construction, and operator-backed postprocessing reconstruction so old unsanitized retained artifacts get a backstop.
- Avoids repeated scans and duplicate dask graphs/history entries when valid local metadata is already present.
- Preserves exact count metadata from single-source ingestion. Multi-source stacking/combining refreshes aggregate metadata because alignment can introduce non-finite cells and source-specific counts are not aggregate counts.
- Does not fabricate an exact count when count mode is requested after a prior lazy fill; the original non-finite values are no longer recoverable.
- Propagates sanitation metadata through multisector flux outputs and PARIS flux outputs after reductions/merges.
- Keeps PR #474's generated-basis weight sanitation as the algorithm guard now that #474 is merged into `devel`.
- Fixes byte-like country labels exposed by the OCO2 PARIS run so `numpy.bytes_` labels do not fail country-code conversion.

Closes #496.
Closes #497.
Closes #498.

Follow-up work is tracked in:

- openghg/openghg#1680 for the authoritative missing-value policy metadata;
- openghg/openghg#1682 for shared xarray JSON/history utilities; and
- #500 for consolidating sanitation after the inversion data-preparation refactor.

* **OCO2 HPC-CI evidence**

Final pixi run succeeded:

- Run directory: `/group/chem/acrg/test_inversions/runs/oco2-satellite_20260708_011822Z`
- Output directory: `/group/chem/acrg/test_inversions/runs/oco2-satellite_20260708_011822Z/outputs`
- SLURM job: `18052524`
- `seff 18052524`: `COMPLETED (exit code 0)`, wall time `00:28:18`, memory utilized `8.67 GB`
- Generated basis completed: `rhime.prepare_inputs.basis_build seconds=29.138565 nbasis=32`
- Sampler completed: `rhime.sampler_total seconds=160.345810 divergences=0`
- Trace save completed: `rhime.output.trace_save seconds=2.823421 path=...trace.nc`
- PARIS postprocess completed: `rhime.output.paris_postprocess seconds=417.754412`
- PARIS NetCDF writes completed: concentration `seconds=0.096311`, flux `seconds=67.048875`
- Run ended with `exit_code=0` and `status=OK`

The earlier venv reproduction had already cleared the issue #495 sparse/dense failure path but failed later with `RuntimeError: NetCDF: HDF error` during NetCDF write. Re-running under pixi completed successfully, consistent with the suspected netCDF4/h5netcdf/HDF wheel conflict rather than a retained-flux sparse/dense regression.

- [x] Closes #495, closes #496, closes #497, and closes #498
- [x] Tests added and passing

Full local pre-PR validation:

```bash
tox
```

Results:

- `py310-openghgCur`: `502 passed, 8 skipped, 6 xfailed`
- `py310-openghgPrev`: `502 passed, 8 skipped, 6 xfailed`
- `py310-openghgDev`: `502 passed, 8 skipped, 6 xfailed`
- `lint`: passed

Additional focused checks:

```bash
uv run ruff check <all changed Python files>
uv run pyright openghg_inversions/flux_sanitization.py openghg_inversions/basis/basis_functions.py openghg_inversions/postprocessing/_basis_products.py openghg_inversions/postprocessing/make_outputs.py openghg_inversions/postprocessing/make_paris_outputs.py
```

Result: both passed.

Lazy dask-mask verification was also run in the current, previous, and development OpenGHG tox environments:

```bash
pytest tests/test_flux_sanitization.py -q
```

Each environment passed all 8 tests. The regression test verifies that lazy sanitation executes no tasks while constructing the graph, remains dask-backed, preserves chunks, and computes the expected zero-filled values.

### DIFF
--- a/openghg_inversions/flux_sanitization.py
+++ b/openghg_inversions/flux_sanitization.py
@@ -1,9 +1,23 @@
-"""Helpers for applying a consistent non-finite flux policy."""
+"""Helpers for applying a consistent non-finite flux policy.
+
+OpenGHG flux data can contain ``NaN`` or infinite values even though those
+values do not have a scientific interpretation for flux weighting. This module
+normalises those values to zero early enough to protect generated and retained
+basis workflows.
+
+The sanitation policy is stored in one namespaced xarray attribute as compact
+JSON. xarray attrs are the metadata channel that survives NetCDF-like
+serialisation, but they should contain simple serialisable values rather than
+Python objects. Keeping the metadata behind a small dataclass gives callers a
+typed interface without spreading many string attribute names through the code.
+"""
 
 from __future__ import annotations
 
 from collections.abc import Mapping
+from dataclasses import dataclass
 from datetime import datetime, timezone
+import json
 from typing import Any, Literal
 import warnings
 
@@ -12,14 +26,9 @@ import xarray as xr
 
 FluxNonFiniteCheck = Literal["lazy", "count"]
 
-NONFINITE_POLICY_ATTR = "openghg_inversions:non_finite_policy"
-NONFINITE_FILL_VALUE_ATTR = "openghg_inversions:non_finite_fill_value"
-NONFINITE_CHECKED_ATTR = "openghg_inversions:non_finite_checked"
-NONFINITE_COUNT_ATTR = "openghg_inversions:non_finite_count"
-NONFINITE_TOTAL_ATTR = "openghg_inversions:non_finite_total"
-NONFINITE_FRACTION_ATTR = "openghg_inversions:non_finite_fraction"
-NONFINITE_CONTEXT_ATTR = "openghg_inversions:non_finite_context"
-NONFINITE_SOURCE_ATTR = "openghg_inversions:non_finite_source"
+# xarray persists attrs to NetCDF/Zarr metadata, so this key is namespaced to
+# avoid collisions with CF/OpenGHG attrs while keeping the value JSON-serialisable.
+NONFINITE_METADATA_ATTR = "openghg_inversions:non_finite_flux"
 
 NONFINITE_POLICY_ZERO_FILL = "zero_fill"
 NONFINITE_CHECKED_NOT_COUNTED = "not_counted"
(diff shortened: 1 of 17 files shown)


openghg_inversions/flux_sanitization.py
```diff
@@ -0,0 +1,253 @@
...
+from datetime import datetime, timezone
+from typing import Any, Literal
+import warnings
+
+import numpy as np
+import xarray as xr
+
+FluxNonFiniteCheck = Literal["lazy", "count"]
+
+NONFINITE_POLICY_ATTR = "openghg_inversions:non_finite_policy"
```
brendan-m-murphy: These needs some explanation: they're namespaced attributes for netCDF-like data.

openghg_inversions/flux_sanitization.py
```diff
@@ -0,0 +1,253 @@
...
+
+def sanitize_fp_all_fluxes(
+    fp_all: dict,
+    *,
+    context: str,
+    check: FluxNonFiniteCheck = "lazy",
+    trust_attrs: bool = True,
+    warn: bool = False,
+) -> dict:
+    """Sanitize every ``fp_all['.flux']`` entry in place and return ``fp_all``."""
```
brendan-m-murphy: All functions (except very short private functions) need full google-style docstrings.

brendan-m-murphy (commented):
In retrospect, using attributes to track the "non-finite" status might be a bad idea. It looks like it will be hard to maintain and understand. Consider alternatives.

openghg_inversions/flux_sanitization.py
```diff
@@ -30,31 +39,201 @@
...
-_LOCAL_ATTRS = (
-    NONFINITE_POLICY_ATTR,
-    NONFINITE_FILL_VALUE_ATTR,
-    NONFINITE_CHECKED_ATTR,
-    NONFINITE_COUNT_ATTR,
-    NONFINITE_TOTAL_ATTR,
-    NONFINITE_FRACTION_ATTR,
-    NONFINITE_CONTEXT_ATTR,
-    NONFINITE_SOURCE_ATTR,
+_LEGACY_NONFINITE_POLICY_ATTR = "openghg_inversions:non_finite_policy"
```
brendan-m-murphy: The "legacy" attributes were introduced in previous commits in this branch, so we do not need to retain them.

openghg_inversions/postprocessing/_basis_products.py
```diff
@@ -127,6 +130,25 @@ def _operator_region_flux_mean(basis_functions: BasisFunctions, flux: xr.DataArr
     return ((basis * flux).sum(grid_dims) / basis.sum(grid_dims)).fillna(0.0)
 
 
+def _sanitize_postprocessing_flux(flux: xr.DataArray, *, context: str) -> xr.DataArray:
```
brendan-m-murphy: This helper could probably be inlined unless there is reason to believe it will be changed in the future.

openghg_inversions/flux_sanitization.py
```diff
@@ -0,0 +1,486 @@
...
+            data["total"] = int(self.total or 0)
+            data["fraction"] = float(self.fraction or 0.0)
+        return data
+
+    def to_json(self) -> str:
+        """Return compact JSON suitable for storing in xarray attrs."""
+        return json.dumps(self.to_json_dict(), sort_keys=True, separators=(",", ":"))
+
+
+def _optional_str(value: object) -> str | None:
```
brendan-m-murphy: These helpers and some of the to/from json code in the class above them seem to be more general "JSON serialization helpers". Perhaps they can be isolated as such. (Or this can be part of a larger serialization clean-up... this kind of code usually lives upstream in `openghg`.)

openghg_inversions/flux_sanitization.py
```diff
@@ -0,0 +1,486 @@
...
+
+    for attr in _UPSTREAM_POLICY_ATTRS:
+        upstream_policy = attrs.get(attr)
+        if upstream_policy is not None and _normalise_policy(upstream_policy) in _ZERO_FILL_POLICY_VALUES:
+            return True
+
+    return False
+
+
+def _history_entry(
```
brendan-m-murphy, line 165: Appending to the history attribute is also more general than this situation.

openghg_inversions/flux_sanitization.py
```diff
@@ -0,0 +1,486 @@
...
+            source=str(source),
+            check=check,
+            trust_attrs=trust_attrs,
+            warn=warn,
+        )
+
+    return fp_all
+
+
+def copy_flux_nonfinite_attrs(
```
brendan-m-murphy: Use generic types since the output type matches the type of `target`.

openghg_inversions/postprocessing/_basis_products.py
```diff
@@ -127,6 +130,25 @@ def _operator_region_flux_mean(basis_functions: BasisFunctions, flux: xr.DataArr
...
+    """Apply the late fallback non-finite policy for old retained artifacts."""
+    return sanitize_flux_nonfinite(
+        flux,
+        context=context,
+        warn=True,
+    )
+
+
+def _with_flux_nonfinite_metadata(ds: xr.Dataset, flux: xr.DataArray) -> xr.Dataset:
```
brendan-m-murphy: These two helpers can be removed once the function they wrap has generic types

openghg_inversions/flux_sanitization.py
```diff
@@ -0,0 +1,391 @@
...
+    count: int | None = None
+    total: int | None = None
+    fraction: float | None = None
+    checked = NONFINITE_CHECKED_NOT_COUNTED
+    if check == "count":
+        total = int(flux.size)
+        count, total, fraction = _nonfinite_count(finite, total)
+        checked = NONFINITE_CHECKED_COMPUTED
+
+    sanitized = flux.where(finite, 0.0)
```
brendan-m-murphy, line 336: Is this tested in the lazy case? selection/where with lazy boolean masks sometimes raises an error.

SutarPrasad (approved):

PR #514, opened 2026-07-20
Serialize RHIME prepared inputs

brendan-m-murphy:
## Summary of changes

- Extract shared DataTree save/load, InferenceData conversion, and MultiIndex
  persistence helpers from postprocessing-private ownership.
- Add strict schema-v1 `RhimePreparedInputs.to_datatree`, `from_datatree`,
  `save`, and `load` APIs for NetCDF and Zarr.
- Embed canonical inversion inputs plus self-contained `BasisFunctions`,
  including retained prior flux required for posterior reconstruction.
- Preserve non-lexicographic multisource order explicitly across both storage
  formats and keep backward loading support for older basis artifacts.
- Validate the outer schema, canonical `nmeasure=(site,time)` MultiIndex,
  nested basis schema/version, typed metadata, and required nodes.
- Verify a reloaded prepared artifact executes through the parent #510 runner.

This is a narrow prepared-input portion of #415 built on the shared-helper work
in #442 and tracked by #509. It is stacked on draft PR #510. It does not attempt
to serialize a live PyMC model, trace, complete sampler/output specs, or private
workflow policy.

## Validation

- Prepared serialization and basis-operator tests: `42 passed`.
- Prepared runner/output/trace compatibility: `23 passed, 93 deselected`.
- Multisource BasisFunctions round trips: `2 passed, 71 deselected`.
- Ruff check and format check: passed.
- Scoped Pyright: `0 errors`, `0 warnings`.
- Independent review and scientific-data rereview: no remaining blockers.
- Full `uv run tox -p --parallel-no-spinner`: passed for current, previous,
  and development OpenGHG environments, plus lint, after rebuilding a stale
  shared test fixture.

## Please check if the PR fulfills these requirements

- [x] Progress towards #415 or #509 — this PR implements one narrow checklist slice. Closes #442
- [x] Tests added and passing.
- [x] Documentation and tutorials updated/added.
- [x] Added an entry to `CHANGELOG.md`.
- [x] No new requirements are needed.

Issue #414, opened 2026-04-29
Generalise component data preparation and deterministic reconstruction


brendan-m-murphy:
## Goal

Generalise the M1 sigma seam into a maintainable component architecture.

## Tasks

- Define a small protocol for component data preparation, PyMC graph construction, and reconstruction
  of derived variables.
- Apply it to linear components, offsets, boundary conditions, and sigma where useful.
- Ensure data-building methods can be used by non-PyMC optimisation/sampling workflows.
- Keep PyMC-specific code out of pure data-building helpers.
- Document when a derived variable should be saved as a PyMC deterministic versus reconstructed later.

## Acceptance criteria

- At least two components support independent data preparation and reconstruction.
- Derived variables needed for outputs can be reconstructed from `InferenceData` plus model data.
- Non-PyMC workflows can reuse the prepared component data.

brendan-m-murphy:
**CO₂-family follow-up from PR #632**

Please include the coherent affine forward contract in the staged component migration, outside PR #632:

```text
linear contribution z is composed visibly by the recipe
fixed affine contribution b is supplied by coherent reduction
modelled concentration = b + z
```

Add one minimal `add_coherent_affine_component(b, z, *, output_name=...)` that registers the labelled fixed contribution and always creates the named `pm.Deterministic`. It must not own state creation, sensitivity application, source/ratio multiplication, channel stacking, or broad validation. CO₂ can pass its single `H_alpha @ x` signal; CO₂/O₂ can pass the visible concatenation of its channel signals. The docs should show `mu_prior + H_alpha(x - m_alpha) = (mu_prior - H_alpha m_alpha) + H_alpha x` and identify `b` as the coherent affine intercept.

Broaden the existing standard/multisector/CO₂ migration checks accordingly:

- use components to own registration of data that is inseparable from their scientific operation rather than leaving bare `add_model_data` calls scattered through builders;
- keep observations and fixed errors owned consistently by the complete error/likelihood component;
- keep named deterministics in each PyMC graph, but test sampling with selected observation-aligned deterministics omitted via `pm.sample(var_names=...)`;
- test `pm.compute_deterministics(...)` against stored reference values with the live model and with the model rebuilt from the #415 bundle;
- retain the backend-neutral labelled reconstruction fallback for postprocessing without PyMC;
- measure trace-size reduction before changing defaults, and preserve compact public state deterministics.

[...]

Issue #35, opened 2024-01-24
Change import statements from `pymc3` to `pymc` (where relevant)


brendan-m-murphy:
@qq23840 noticed that `hbmcmc.hbmcmc_post_process` has `import pymc3 as pm`. This should be updated to `import pymc as pm` since our requirements.txt doesn't pin the version of PyMC, hence we are on the latest version by default.

Look for other instances and update.

brendan-m-murphy:
This has been addressed.

PR #515, opened 2026-07-20
Prepare RHIME inputs from source-neutral xarray data

brendan-m-murphy:
## Summary of changes

- Add public, source-neutral `prepare_rhime_inputs_from_xarray` for ordered
  per-site Dataset mappings and root DataTrees with one site-local Dataset in
  each direct child.
- Accept canonical `H` or cached `fp_x_flux` /
  `fp_x_flux_sectoral`; project caches through the retained basis before the
  existing concat-gather path combines unequal site/time axes, and omit the
  gridded caches from canonical inputs.
- Preserve gathered source-specific/ragged basis state, retained flux, source
  order, sampled `H_bc`, extension variables, release metadata, and averaging
  periods through `RhimePreparedInputs` serialization and prepared execution.
- Validate site-local layouts, time coordinates, concentration units,
  source/state compatibility, source labels and ordering, finiteness, retained
  sites, and release-coordinate semantics at the public boundary.
- Match the existing RHIME preparation default with
  `min_error_per_site=False`.
- Explicitly reject direct Dataset, dense `Dataset(site, time)`, pre-stacked
  `nmeasure`, and deterministic `fixed_baseline` inputs. Dense/stacked
  generalization remains deferred; fixed-baseline semantics are tracked in
  #550 as a reusable Baseline component.
- Leave legacy HBMCMC and legacy output behavior unchanged. Existing sampled
  `H_bc` and modern/PARIS output paths continue to work.
- Separate verification-games-style custom input assembly from OpenGHG
  Inversions adaptation, persistence, and execution in the usage guide.

This is the verification-games MVP slice of #509 and advances the cached
`fp_x_flux` path in #492. Prepared-input persistence is supplied by merged
#514; serialization is available but is not invoked implicitly by this
adapter. OpenGHG retrieval remains an adapter, not a requirement.

## Validation

- Focused adapter/input/model/output suites: `294 passed, 1 deselected`.
- NetCDF and Zarr unequal-source-region round-trip and prepared execution:
  passed, including monthly transformed sampled `H_bc`.
- Ruff check and format check: passed.
- Scoped Pyright: `0 errors`.
- Final public-docstring audit: no remaining blockers.
- Required local `tox -p --parallel-no-spinner`: current OpenGHG tests and
  lint passed.
- GitHub Actions: OpenGHG 0.18, 0.19, and `devel` tests, lint, and CI gate all
  passed.

The configured mypy environment was also invoked. It remains non-green because
of the repository's existing typing backlog (88 errors across 21 imported
files, including missing third-party stubs); no new adapter-specific type error
was identified beyond the existing missing `pandas` stubs.

- [ ] Implements parts of #509 and #492
- [x] Tests added and passing.
- [x] Documentation and tutorials updated/added.
- [x] Added an entry to `CHANGELOG.md`.
- [x] No new requirements are needed.

### DIFF
--- a/docs/usage/rhime.rst
+++ b/docs/usage/rhime.rst
@@ -209,6 +209,220 @@ they are stacked on ``source``. Inputs with different native frequencies, such
 as hourly and monthly fluxes, require an explicit resampling policy rather than
 implicit xarray alignment.
 
+Source-neutral xarray inputs can be adapted at this boundary without OpenGHG
+data acquisition. The supported container boundary is deliberately narrow:
+
+.. list-table::
+   :header-rows: 1
+   :widths: 25 15 60
+
+   * - Container
+     - Support
+     - Contract
+   * - Ordered ``Mapping[str, Dataset]``
+     - Supported
+     - One active-observation dataset per site; mapping insertion order is
+       retained.
+   * - ``DataTree``
+     - Supported
+     - One direct child dataset per site; child order is retained.
+   * - Direct ``Dataset``
+     - Unsupported
+     - Wrap one or more site-local datasets in an ordered mapping or a
+       direct-child ``DataTree``. Load an already canonical artifact with
+       ``RhimePreparedInputs.load``.
+   * - Dense ``Dataset(site, time)``
+     - Unsupported
+     - No trimming or interpretation of padded rows is attempted. Split it
+       into per-site datasets first.
+   * - Padded dense arrays
+     - Unsupported
+     - Every row in a site-local dataset is active; padding has no MVP
+       semantics.
+   * - Pre-stacked ``nmeasure`` data
+     - Unsupported
+     - Pass per-site data to the adapter. Load an already canonical prepared
+       artifact with ``RhimePreparedInputs.load`` instead.
+
+This matches the cached verification-games workflow: sites may have unequal or
+disjoint time axes, and the existing ragged gather path creates ``nmeasure``
+only after per-site validation and basis projection. For example:
+
+.. code-block:: python
+
+   from collections import OrderedDict
+
+   import numpy as np
+   import xarray as xr
+
+   from openghg_inversions.basis.basis_functions import BasisFunctions
+   from openghg_inversions.inversion_data import (
+       RhimePreparedInputs,
+       prepare_rhime_inputs_from_xarray,
+   )
+   from openghg_inversions.models import RhimeModelSpec, SectorSpec
+   from openghg_inversions.rhime import (
+       RhimeOutputSpec,
+       RhimeRunSpec,
+       RhimeSampler,
+       run_rhime_from_prepared_inputs,
+   )
+
+   # This retained artifact owns the basis operator, source-resolved reference
+   # flux, state labels, and source order used for later reconstruction.
+   basis_functions = BasisFunctions.load("base-country-basis.zarr")
+   sources = ["ocean", "FF", "GPP", "TER"]  # meaningful, non-alphabetical order
+
+   site_times = OrderedDict(
+       TAC=np.array(
+           ["2021-01-01T12:00", "2021-01-02T12:00"],
+           dtype="datetime64[ns]",
+       ),
+       MHD=np.array(
+           ["2021-01-01T15:00", "2021-01-03T15:00", "2021-01-04T15:00"],
+           dtype="datetime64[ns]",
+       ),
+   )
+   site_data = OrderedDict()
+   for site, times in site_times.items():
+       cache = xr.open_zarr(f"{site.lower()}-base-fp-x-flux.zarr")[
+           "fp_x_flux_sectoral"
+       ].sel(source=sources, time=times)
+       # Current verification-games caches store mole fractions with units
+       # "1". Numerically convert them in that consumer before adaptation;
+       # changing metadata alone would be scientifically incorrect.
+       if cache.attrs.get("units") == "1":
+           cache = (cache * 1e6).assign_attrs({**cache.attrs, "units": "ppm"})
+       if cache.attrs.get("units") != "ppm":
+           raise ValueError("Normalize this cache numerically to ppm first.")
+       observation = xr.DataArray(
+           np.zeros(times.size),
+           dims="time",
+           coords={"time": times},
+           attrs={"units": "ppm"},
+       )
+       site_data[site] = xr.Dataset(
+           {
+               "mf": observation,
+               "mf_error": xr.ones_like(observation).assign_attrs(units="ppm"),
+               "mf_repeatability": xr.ones_like(observation).assign_attrs(units="ppm"),
+               "mf_variability": xr.zeros_like(observation).assign_attrs(units="ppm"),
+               "fp_x_flux_sectoral": cache,
+           }
+       )
+
+   prepared = prepare_rhime_inputs_from_xarray(
+       site_data,
+       basis_functions=basis_functions,
+       averaging_period="1h",
+       min_error=0.0,
+       start_date="2021-01-01",
+   )
+
+   # Optional durable checkpoint: no OpenGHG retrieval or adapter projection is
+   # repeated when the prepared artifact is loaded.
+   prepared.save("base-prepared-inputs.zarr")
+   prepared = RhimePreparedInputs.load("base-prepared-inputs.zarr")
+
+   sectors = tuple(
+       SectorSpec(
+           name=source,
+           flux_source=source,
+           x_prior={"pdf": "lognormal", "mean": 1.0, "stdev": 1.0},
+           variable_suffix=source.lower(),
+       )
+       for source in sources
+   )
+   run_spec = RhimeRunSpec(
+       start_date="2021-01-01",
+       end_date="2021-01-05",
+       sites=prepared.sites,
+       averaging_period=prepared.averaging_period,
+       model=RhimeModelSpec(
+           species="co2",
+           domain="EUROPE",
+           sectors=sectors,
+           sigma_freq="monthly",
+       ),
+       output=RhimeOutputSpec(output_format="none"),
+       split_by_sectors=True,
+   )
+   result = run_rhime_from_prepared_inputs(
+       prepared_inputs=prepared,
+       run_spec=run_spec,
+       sampler=RhimeSampler(draws=1000, tune=1000, chains=4),
+   )
+
+Each site dataset must contain ``mf``, ``mf_error``, ``mf_repeatability``,
+``mf_variability``, and either canonical ``H`` or cached ``fp_x_flux`` /
+``fp_x_flux_sectoral``. Cached fields are projected through the retained basis
+before ragged sites are gathered and are not copied into canonical inputs.
+Other time-aligned extension variables are retained. Non-time-dependent data
+variables are removed at this boundary rather than copied into every
+observation. The retained operator and prior/reference flux are carried into
+postprocessing; artifact provenance may be added to the returned basis value.
+
+The adapter treats every supplied row as an active observation. Each site
+therefore needs a nonempty, explicit, unique ``datetime64`` ``time`` coordinate
+with no ``NaT`` values. Valid non-monotonic input ordering is retained.
+Required observations, projected ``H``, and optional baseline contributions
+must be finite. A site emptied by downstream input preparation is reported as
+an error instead of being silently removed. ``min_error_per_site`` defaults to
+``False``, matching the OpenGHG-backed RHIME preparation route; pass ``True``
+to request per-site minimum errors.
+
+``mf.units`` must be a nonempty string. Every present concentration-valued
+field---observation-error components, canonical ``H`` or the selected cache,
+``H_bc``, and ``fixed_baseline``---must declare the same exact unit string at
+every site. The adapter does not perform unit conversion; normalize equivalent
+spellings before calling it. Projected ``H`` inherits the selected cache units.
+
+Canonical ``H`` may use ``(region, time)`` for one source,
+``(region, time, source)`` for multisource data with one shared basis, or a
+gathered state dimension with a ``(source, region_in_source)`` MultiIndex for
+source-specific ragged bases. Multisector inputs must carry source-resolved
+retained prior flux with the same source names and order as ``H``;
+source-specific basis operators must match that order too. The adapter rejects
+a total retained flux for multisector inputs instead of broadcasting it across
+sectors. Labels on explicit ``source`` dimensions must be unique, nonempty
+Python or NumPy strings: bytes, numbers, duplicates, and implicit coercion are
+rejected. Repeated source values within the gathered
+``(source, region_in_source)`` state MultiIndex are valid. Exact source order is
+preserved across the cache, canonical ``H``, retained flux, and operator
+metadata. Ragged state blocks are concatenated with the modern gather helpers;
+the adapter never introduces rectangular zero padding. ``H_bc`` must contain
+exactly ``time`` and ``bc_region`` dimensions in either order.
+
+Release coordinates must be supplied as a pair. Stationary surface or column
+sites may use scalar or singleton ``release_lat`` and ``release_lon`` values;
+the adapter broadcasts them over that site's observations for the PARIS
+template. Mobile platforms must supply both coordinates on exactly the site's
+``time`` dimension, in matching order. Partial, non-finite, or other coordinate
+layouts are rejected rather than flattened or inferred. If any retained site
+supplies release coordinates, every retained site must supply its pair.
+
+Dense padding, pre-stacked layouts, unit conversion, and coercion of source
+labels are intentionally deferred generalizations. They should be added only
+with a concrete consumer and explicit semantics rather than inferred by this
+adapter.
+
+Two independent baseline modes are supported at this prepared-input boundary.
+Include ``H_bc`` for the existing sampled boundary-condition contribution.
+Include ``fixed_baseline(time)`` for a supplied observation-aligned contribution
+in the same units as ``mf`` that is added directly to the model mean. Both
+variables may be present. If any retained site supplies ``fixed_baseline``, all
+retained sites must supply it. The fixed term is not represented as a flux
+sector. The existing INI/OpenGHG configuration wrappers do not yet synthesize
+``fixed_baseline``; callers using that mode should use this xarray adapter and
+``run_rhime_from_prepared_inputs``.
+
+Generic concentration statistics and PARIS concentration products treat the
+effective baseline as fixed plus sampled contributions. The sampled ``mu_bc``
+trace remains unchanged; the fixed contribution is composed only in the model
+mean and output views. The temporary HBMCMC-compatible ``legacy`` output does
+not accept fixed baselines, and neither does the legacy HBMCMC model builder;
+new prepared-input features do not extend those compatibility paths.
+
 Config Files
 ------------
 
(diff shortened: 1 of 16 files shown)


openghg_inversions/inversion_data/xarray_adapter.py
```diff
@@ -0,0 +1,493 @@
...
+from openghg_inversions.inversion_inputs import DatetimeLike, make_inv_inputs
+
+XarraySiteData: TypeAlias = xr.Dataset | xr.DataTree | Mapping[str, xr.Dataset]
+"""Supported xarray containers for source-neutral site data."""
+
+_CACHE_VARIABLES = ("fp_x_flux", "fp_x_flux_sectoral")
+_REQUIRED_OBSERVATION_VARIABLES = ("mf", "mf_error", "mf_repeatability", "mf_variability")
+
+
+def _ordered_unique_strings(values: np.ndarray) -> list[str]:
```
brendan-m-murphy, line 75: Can this use `np.unique`?

openghg_inversions/inversion_data/xarray_adapter.py
```diff
@@ -0,0 +1,493 @@
...
+            if not np.all(np.isfinite(padded_values) & (padded_values == 0)):
+                raise ValueError(
+                    f"Site {site!r} H contains non-zero or non-finite padded state values for "
+                    f"source {source!r} outside retained basis regions {source_regions.tolist()!r}."
+                )
+
+
+def _sites_from_dataset(data: xr.Dataset, sites: Sequence[str] | None) -> dict[str, xr.Dataset]:
+    """Split a site-dimension dataset into an ordered per-site mapping."""
+    if "site" in data.dims:
```
brendan-m-murphy, line 383: **[P1] Require site-specific variables.** For a multi-site `Dataset`, `isel` only slices variables carrying the `site` dimension; shared observations, errors, `H`, or caches are silently reused for every site. Please validate that all site-specific inputs include `site` before splitting.

openghg_inversions/inversion_data/xarray_adapter.py
```diff
@@ -0,0 +1,493 @@
...
+            )
+        if "fixed_baseline" in dataset:
+            fixed_baseline = dataset["fixed_baseline"]
+            if fixed_baseline.dims != ("time",):
+                raise ValueError(
+                    f"Site {site!r} fixed_baseline must have exactly the observation dimension ('time',)."
+                )
+            if not np.isfinite(np.asarray(fixed_baseline.values)).all():
+                raise ValueError(f"Site {site!r} fixed_baseline must contain only finite values.")
+            fixed_units = fixed_baseline.attrs.get("units")
```
brendan-m-murphy: **[P1] Enforce consistent concentration units.** Validation is only intra-site and only runs when both unit attributes exist. Mixed units across sites are gathered as raw values while retaining the first site's metadata, which can silently introduce scale errors. This should enforce one common concentration unit across sites and concentration-valued inputs, including one-sided missing metadata.

openghg_inversions/inversion_data/xarray_adapter.py
```diff
@@ -0,0 +1,493 @@
...
+        ValueError: If site data violate the documented variable, dimension,
+            source-order, or site-alignment contract.
+    """
+    if not isinstance(basis_functions, BasisFunctions):
+        raise TypeError("`basis_functions` must be a self-contained BasisFunctions object.")
+
+    per_site = _sites_from_container(data, sites)
+    release_coordinates = _site_release_coordinates(per_site)
+    prepared_sites = _validate_and_prepare_sites(per_site, basis_functions)
+    site_order = list(prepared_sites)
```
brendan-m-murphy, line 947: **[P1] Reconcile sites after filtering.** `make_inv_inputs` can remove every observation for a site, but the returned site order, release coordinates, and averaging periods remain unchanged. Please reject emptied sites or recompact every site-aligned field after filtering.

openghg_inversions/inversion_data/xarray_adapter.py
```diff
@@ -0,0 +1,493 @@
...
+                f"Site {site!r} observation variable(s) must have exactly dimension ('time',): "
+                f"{invalid_observation_dims!r}."
+            )
+        if "fixed_baseline" in dataset:
+            fixed_baseline = dataset["fixed_baseline"]
+            if fixed_baseline.dims != ("time",):
+                raise ValueError(
+                    f"Site {site!r} fixed_baseline must have exactly the observation dimension ('time',)."
+                )
+            if not np.isfinite(np.asarray(fixed_baseline.values)).all():
```
brendan-m-murphy: **[P2] Permit inactive ragged padding.** This finite-only check rejects `NaN` padding produced by a dense ragged `site × time` dataset before the corresponding inactive observation rows can be removed. Could this trim inactive rows first while continuing to reject infinities and active-row NaNs?

openghg_inversions/inversion_data/xarray_adapter.py
```diff
@@ -0,0 +1,493 @@
...
+        if not np.array_equal(state_coord.values, expected_state.values):
+            raise ValueError(
+                f"Site {site!r} H region coordinate does not match the retained basis operator "
+                f"state coordinate: found {state_coord.values.tolist()!r}, "
+                f"expected {expected_state.values.tolist()!r}."
+            )
+        return
+
+    source_coord = _require_dim_coordinate(sensitivity, "source", context=f"Site {site!r} H")
+    sources = [str(source) for source in source_coord.values]
```
brendan-m-murphy: **[P2] Normalize or reject source labels.** Source labels are compared after string conversion but retained with their original dtype. Numeric labels therefore pass here and later fail when the model selects their string forms. Please reject non-string/duplicate labels or normalize all related coordinates consistently.

openghg_inversions/inversion_data/xarray_adapter.py
```diff
@@ -0,0 +1,493 @@
...
+def prepare_rhime_inputs_from_xarray(
+    data: XarraySiteData,
+    *,
+    basis_functions: BasisFunctions,
+    sites: Sequence[str] | None = None,
+    averaging_period: str | Sequence[str | None] | Mapping[str, str | None] | None = None,
+    bc_freq: str | None = None,
+    sigma_freq: str | None = None,
+    min_error: str | dict[str, float] | int | float = 0.0,
+    min_error_per_site: bool = True,
```
brendan-m-murphy: **[P2] Align the residual-error default.** The adapter defaults calculated residual errors to per-site, whereas the existing RHIME preparation path defaults `by_site` to `False`. Equivalent inputs can therefore produce different likelihoods depending on the preparation path.

openghg_inversions/inversion_data/xarray_adapter.py
```diff
@@ -0,0 +1,493 @@
...
+    if fixed_baseline_sites and len(fixed_baseline_sites) != len(site_data):
+        missing = [site for site in site_data if site not in fixed_baseline_sites]
+        raise ValueError(
+            f"Fixed baseline data must be supplied for every retained site; missing {missing!r}."
+        )
+
+    expected_sources: list[str] | None = None
+    source_layout_set = False
+    for site, dataset in site_data.items():
+        if "time" not in dataset.dims:
```
brendan-m-murphy, line 696: **[P2] Validate the time coordinate.** Only the dimension is required. Duplicate, non-datetime, `NaT`, or empty time coordinates can fail later or make PARIS outputs impossible to unstack. Please validate the coordinate at this public boundary.

openghg_inversions/postprocessing/legacy_outputs.py
```diff
@@ -780,15 +790,22 @@ def make_legacy_hbmcmc_output(
         "xsensitivity": (("nmeasure", "nparam"), Hx.T),
     }
 
-    if use_bc and Hbc is not None and bcouts is not None:
+    if fixed_baseline_values is not None or sampled_baseline:
```
brendan-m-murphy: **[P2] Preserve fixed-only legacy metadata.** Fixed-only outputs create the BC variables here, but `_set_legacy_var_attrs` is later called with `use_bc=False`, so these fields omit the units and semantic long names supplied for sampled baselines. Attribute inclusion should depend on effective-baseline presence rather than sampled-BC presence.

brendan-m-murphy (commented):
Posting the inline findings from the completed multi-agent review, together with Brendan's current review comment, so they are recorded publicly on PR #515.

openghg_inversions/inversion_data/xarray_adapter.py
brendan-m-murphy: This issue might be avoided entirely by using concat-gather from `array_ops.py`.

openghg_inversions/inversion_data/xarray_adapter.py
brendan-m-murphy: Normalising is preferable if it can be done reliably.

openghg_inversions/inversion_data/xarray_adapter.py
```diff
@@ -0,0 +1,493 @@
...
+
+
+def _sites_from_dataset(data: xr.Dataset, sites: Sequence[str] | None) -> dict[str, xr.Dataset]:
+    """Split a site-dimension dataset into an ordered per-site mapping."""
+    if "site" in data.dims:
+        if "site" not in data.coords:
+            raise ValueError("An input Dataset with a 'site' dimension must provide site coordinate values.")
+        available = [str(value) for value in data.coords["site"].values]
+        if len(set(available)) != len(available):
+            raise ValueError("Input Dataset site coordinate values must be unique.")
```
brendan-m-murphy: Is this possible with a stacked (site, time) coordinate?

openghg_inversions/inversion_data/xarray_adapter.py
```diff
@@ -0,0 +1,493 @@
...
+sector-resolved ``fp_x_flux_sectoral`` is used. Optional ``H_bc`` retains the
+existing sampled boundary-condition contribution. Optional
+``fixed_baseline(time)`` is retained as a fixed, observation-aligned model
+contribution in the same units as ``mf``; it may be supplied alone or together
+with ``H_bc``.
+
+Canonical ``H`` uses dimensions ``(region, time)`` or
+``(region, time, source)``. Its region coordinate must exactly match the
+retained operator state. Multisector inputs additionally require source-resolved
+retained prior flux, and any source-specific operator metadata, with exactly the
```
brendan-m-murphy: More details or examples might be needed to make this cogent without reading the code. Don't add details to this summary part, but we need more documentation for this feature: it is explicitly an extension point for user customisation, so it needs to be clear how to use it.

openghg_inversions/inversion_data/xarray_adapter.py
```diff
@@ -0,0 +1,493 @@
...
+    return _ordered_unique_strings(np.asarray(basis_matrix.coords["source"].values))
+
+
+def _validate_state_and_source_compatibility(
+    sensitivity: xr.DataArray,
+    basis_functions: BasisFunctions,
+    *,
+    site: str,
+) -> None:
+    """Validate H state/source coordinates against retained basis and prior flux."""
```
brendan-m-murphy: A docstring explaining the exact contract checked would be useful.

docs/usage/rhime.rst
```diff
@@ -209,6 +209,220 @@ they are stacked on ``source``. Inputs with different native frequencies, such
...
+   * - Pre-stacked ``nmeasure`` data
+     - Unsupported
+     - Pass per-site data to the adapter. Load an already canonical prepared
+       artifact with ``RhimePreparedInputs.load`` instead.
+
+This matches the cached verification-games workflow: sites may have unequal or
+disjoint time axes, and the existing ragged gather path creates ``nmeasure``
+only after per-site validation and basis projection. For example:
+
+.. code-block:: python
```
brendan-m-murphy, line 259: Use multiple code blocks here so you can make the steps clearer. This will also let us emphasise what is "custom code" (e.g. from verification-games in this example) and what is openghg inversions code.

openghg_inversions/inversion_data/xarray_adapter.py
brendan-m-murphy: Is this irrelevant now that we're not accepting a single input Dataset?

openghg_inversions/models/components.py
```diff
@@ -348,9 +356,18 @@ def add_inferpymc_likelihood_component(
...
     )
 
+    fixed_baseline = (
+        add_model_data(data["fixed_baseline"].transpose(output_dim), "fixed_baseline")
+        if "fixed_baseline" in data
+        else None
+    )
+
     if pollution_events_from_obs is True:
+        supplied_baseline = fixed_baseline
```
brendan-m-murphy: This conditional code for different baseline options is messy. I would prefer a `Baseline` component, which can be created in various ways, but is consumed by `Likelihood` components in the same way. If that is outside of the scope of this PR, track this requirement.

Note that verification-games is not currently using this likelihood, and this new baseline code is not portable/modular, so it would need to be duplicated.

The `model_configs` branch was an earlier model component prototype, and it had semantic components, like "Baseline", which could include e.g. standard mu_bc = H_bc @ x_bc style baseline, with or without an offset.

openghg_inversions/postprocessing/make_outputs.py
```diff
@@ -681,13 +681,85 @@ def sort_key(s: str):
     return ds[dv_sorted]  # type: ignore
 
 
+def _fixed_baseline_for_outputs(inv_out: InversionOutput) -> xr.DataArray | None:
```
brendan-m-murphy: Again, a "fixed baseline" should really need separate handling, or if it does, it should be bundled with the "component".

openghg_inversions/inversion_inputs.py
```diff
@@ -294,6 +296,7 @@ def _drop_nan_and_compute(
     to_compute: list[str] = [
         "H",
         "H_bc",
+        "fixed_baseline",
```
brendan-m-murphy: I don't have a great suggestion for avoiding hardcoding variable names here, but eventually we need to stop doing it.

brendan-m-murphy (commented):
The documentation needs a bit more explanation to separate illustrative custom code from the openghg_inversions code that it hooks in to. Also, I think this is stretching what "quickstart" means, so the docs might need a general reorganization, but that can be deferred.

The new fixed baseline option is not implemented in a modular way. In some sense, the framework for how "modular components" should work is still lacking, so perhaps a "modular baseline component" is too much to ask for this PR.

openghg_inversions/inversion_data/xarray_adapter.py
brendan-m-murphy: Yes—the original split/reuse risk is obsolete because direct Dataset inputs are now rejected and no `isel` split remains. The code at the remapped line has the narrower job of rejecting a `site` or `nmeasure` dimension inside a Mapping/DataTree child that must be site-local. I retained that guard and added Mapping/DataTree child coverage.

openghg_inversions/inversion_data/xarray_adapter.py
brendan-m-murphy: Addressed. `mf`, observation-error fields, selected `H`/cache, and optional sampled `H_bc` must all declare the same exact nonempty unit string at every site. Missing or mixed metadata now fails before gathering; the adapter performs no conversion.

openghg_inversions/inversion_data/xarray_adapter.py
brendan-m-murphy: Resolved by narrowing the contract. Dense padded `site × time` inputs are rejected; callers provide site-local Mapping/DataTree children and the existing `concat_gather_datasets` path gathers unequal/disjoint time axes without padding. Child `site`/`nmeasure` rejection is now covered.

openghg_inversions/inversion_data/xarray_adapter.py
brendan-m-murphy: Addressed without coercion: labels on explicit `source` dimensions must be nonempty, unique Python/NumPy strings. Their first-occurrence order is preserved; bytes, numeric labels, and duplicates fail early. Repeated source values in a gathered source-state MultiIndex remain valid.

openghg_inversions/inversion_data/xarray_adapter.py
brendan-m-murphy: Addressed. `min_error_per_site` now defaults to `False`, matching the OpenGHG-backed RHIME preparation route, while retaining the explicit `True` override.

openghg_inversions/postprocessing/legacy_outputs.py
brendan-m-murphy: This is obsolete after the YAGNI cleanup. Fixed-baseline support has been removed from #515 and the legacy output implementation is back to `devel` unchanged. Future deterministic/fixed behavior belongs to the semantic Baseline component tracked in #550, not the legacy adapter.

openghg_inversions/inversion_data/xarray_adapter.py
brendan-m-murphy: A pre-stacked `nmeasure` layout is now explicitly outside this adapter contract and fails with guidance. The supported route is one site-local Dataset per Mapping/DataTree child, followed by the existing concat-gather machinery; already prepared stacked data should be reopened with `RhimePreparedInputs.load`.

openghg_inversions/inversion_data/xarray_adapter.py
brendan-m-murphy: Addressed in the usage guide. The verification-games-shaped example now separates application-specific acquisition/normalization, OpenGHG Inversions adaptation, optional serialization, and prepared execution. It also states that this is an input-assembly boundary, not a model/likelihood plugin interface.

openghg_inversions/inversion_data/xarray_adapter.py
brendan-m-murphy: Addressed. `_validate_state_and_source_compatibility` now documents the supported single-source, shared multisource, and gathered source-specific layouts, exact region/state matching, source order, retained flux/operator compatibility, and its failure contract.

docs/usage/rhime.rst
brendan-m-murphy: Addressed. The example is now four labelled blocks: application-specific verification-games-style assembly, OpenGHG Inversions adaptation, optional save/load, and OpenGHG Inversions execution. The prose identifies exactly where the public adapter begins.

openghg_inversions/models/components.py
brendan-m-murphy: Agreed. Current verification-games uses sampled `H_bc` and a custom likelihood, not this fixed-baseline prototype. I removed fixed-baseline behavior from #515 and created #550 for a reusable semantic Baseline component covering absent/fixed/sampled/combined behavior, likelihood consumption, offsets, serialization, and output reconstruction. Verification-games transition issue #35 now points to sampled `H_bc` for the immediate path.

openghg_inversions/postprocessing/make_outputs.py
brendan-m-murphy: Agreed. The fixed-specific postprocessing overlay has been removed with the deferred fixed-baseline feature. Existing sampled `H_bc` generic/PARIS behavior remains unchanged; #550 owns future component-level composition and reconstruction semantics.

openghg_inversions/inversion_inputs.py
brendan-m-murphy: Agreed. Removing `fixed_baseline` also removes all new hard-coded variable handling introduced here. I have not added a local registry that would merely move the names without defining their policies; #414, #528, and #550 track the component-owned input/reconstruction architecture.

Issue #493, opened 2026-07-06
Define labelled native-covariance and basis-projection contracts


brendan-m-murphy:
## Goal

Define the labelled native-covariance and basis-projection contracts needed to propagate grid-cell prior uncertainty into retained state covariance and observation-space aggregation error.

This issue is the lower-level covariance/projection dependency for the coherent reduction in #566.

## Current status

Some of the original scope has now been addressed:

- #521 added basis-aware prior-width projection and calibration for independent grid-cell perturbations, including gathered multisource bases.
- #564 added backend-neutral aggregation-error representations for diagonal, dense, and low-rank-plus-diagonal covariance and connected them to RHIME likelihood construction.

Those APIs provide useful diagnostics and downstream covariance consumers. OGI still lacks the correlated native-covariance operator and product-block producer needed to construct `C_alpha`, `H B Pi.T`, and `H B H.T`.

### Labelled native covariance operator

Define a backend-neutral, serializable covariance action over a labelled native grid. The central operation should apply `B` to one or more labelled right-hand sides without requiring a dense native `N x N` matrix.

The contract should:

- preserve native grid coordinates and right-hand-side labels;
- validate dimensions, coordinate alignment, units, finite values, and symmetry expectations;
- support multiple right-hand sides efficiently;
- expose configuration/provenance sufficient to reproduce the covariance action;
- permit alternative covariance implementations without changing the projection API.

### Basic separable exponential implementation

Provide a usable initial implementation:

- `K_lat[i,j] = exp(-|lat_i-lat_j| / ell_lat)`;
- `K_lon[i,j] = exp(-|lon_i-lon_j| / ell_lon)`;
- `B = sigma^2 (K_lat ⊗ K_lon)`;
- configurable latitude/longitude length scales and amplitude/grid-cell SD;
- an introductory default of `ell_lat = ell_lon = 1.5°`;
- optional class blocking, `B_class = sum_c M_c B M_c`, which preserves within-class covariance and sets cross-class covariance to zero;
- matrix-free/separable application to dense labelled right-hand sides.

The `1.5°` value should be a convenient documented default, not a hard-coded scientific requirement. Callers remain free to configure length scales, amplitudes, masks, and calibration policy.

### Product-space projection

Given a covariance action `B`, labelled native observation operator `H`, and labelled basis restriction/operator `Pi`, provide public helpers that compute:

- `C_alpha = Pi B Pi.T`;
- `H B Pi.T`;
[...]

PR #516, opened 2026-07-20
Keep RHIME likelihood error components separate

brendan-m-murphy:
## Summary of changes

- Preserve raw `mf_error`, optional diagonal `aggregation_error_sd`,
  `min_error`, inferred `model_error`, and total `epsilon` as distinct values.
- Validate aligned finite nonnegative aggregation SDs and combine independent
  diagonal terms exactly once at the likelihood boundary.
- Propagate averaged aggregation SD as `sqrt(sum(sd**2)) / n` rather than an
  arithmetic mean.
- Keep generic inversion outputs component-aware while explicitly omitting the
  internal aggregation field from fixed PARIS v03/v04 concentration schemas.
- Preserve aggregation error, floors, PEFO mode, no-model-error mode, and
  nondefault power through backward-compatible legacy save/replay, including
  repeated replay generations.
- Avoid misreporting floor-clipped residual variance as inferred model error.

This is the diagonal error-contract slice tracked by #509 and #493. Full
aggregation covariance remains out of scope.

## Validation

- Focused model/input/resampling tests: `20 passed`.
- Legacy/current postprocessing tests: `10 passed`.
- Standard/multisector/save-load/PARIS tests: `5 passed`.
- Final replay, legacy-writer, and floor-masking review selection: `9 passed`.
- Ruff check and format check: passed.
- Scoped Pyright on modern output/runner paths: `0 errors`.
- Independent correctness rereview: no remaining blockers.
- Full tox matrix (current, previous, and development OpenGHG plus lint): passed.

- [ ] Closes #509 or #493 — this PR implements the diagonal error slice only.
- [x] Tests added and passing.
- [x] Documentation and tutorials updated/added.
- [x] Added an entry to `CHANGELOG.md`.
- [x] No new requirements are needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@
 
 ## Code changes
 
+- Added modern RHIME likelihood support for fixed aggregation-error covariance
+  supplied after input preparation: exact dense, low-rank-plus-diagonal, and
+  independent diagonal representations. Structured likelihoods retain the
+  observed ``y`` variable and predictive sampling, preserve the total-marginal
+  minimum-error floor, and leave legacy HBMCMC paths unchanged. Derived basic,
+  PARIS, and legacy products reject these inputs pending representation-neutral
+  postprocessing reconstruction. [PR #516](https://github.com/openghg/openghg_inversions/pull/516)
+
 - Added coordinate-preserving loaders and a weights-first region-constrained
   fixed-outer adapter. Packaged InTEM and raw country/land-sea class maps are
   normalized to the weights grid, outer classes retain one target each, and
--- a/docs/usage/rhime.rst
+++ b/docs/usage/rhime.rst
@@ -165,6 +165,37 @@ multisector layout. The sector count, prepared ``H`` layout, layout flag, and
 output settings are validated before model construction or sampling. Output
 side effects are still controlled by ``RhimeOutputSpec``.
 
+Aggregation-error covariance
+-----------------------------
+
+Modern RHIME can add fixed aggregation error to the observation covariance
+without changing the raw ``mf_error`` values. Attach these arrays only after
+observation filtering, averaging, global site gathering, and basis projection,
+so both covariance axes match the final ``nmeasure`` ordering. Prepared inputs
+may contain one of:
+
+* ``aggregation_error_covariance(nmeasure, nmeasure_cov)`` for the exact dense
+  covariance;
+* ``low_rank_factor(nmeasure, agg_rank)`` together with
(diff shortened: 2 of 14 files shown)


openghg_inversions/inversion_data/getters.py
```diff
@@ -33,6 +33,69 @@
 
 logger = logging.getLogger(__name__)
 
+_AGGREGATION_ERROR_NAMES = ("aggregation_error_sd", "mf_aggregation_error_sd")
+
+
+def _resample_independent_error_sd(
```
brendan-m-murphy: There is already code to do resampling in openghg (it is used by get_obs_surface)

openghg_inversions/hbmcmc/hbmcmc.py
```diff
@@ -24,6 +24,7 @@
 the users OpenGHG config file (default location: ~/.openghg/openghg.conf).
 """
```
brendan-m-murphy, line 26: No new code should go in the hbmcmc submodule unless it is necessary for some change to the modern run rhime pathway

openghg_inversions/inversion_data/getters.py
```diff
@@ -33,6 +33,69 @@
...
+        raise ValueError("Aggregation-error resampling requires non-negative standard deviations.")
+
+    with xr.set_options(keep_attrs=True):
+        count = error_sd.resample(time=averaging_period).count()
+        variance = (error_sd**2).resample(time=averaging_period).sum() / count**2
+        result = variance**0.5
+    return result.rename(error_sd.name)
+
+
+def _replace_mean_resampled_aggregation_error(
```
brendan-m-murphy: Diagonal approximation to aggregation error should be a secondary option. Dense exact aggregation error and truncated SVD approximations (with diagonal terms for truncated singular values) of it are the main options.

openghg_inversions/models/components.py
```diff
@@ -423,6 +440,7 @@ def add_inferpymc_likelihood_component(
     if offset is not None:
         total_mu = total_mu + offset
 
+    pm.Deterministic("model_error", pt.sqrt(model_error_variance), dims=output_dim)
```
brendan-m-murphy: Favor the sigma alignment pattern where deterministics can be recovered after sampling, if needed. Storing deterministics takes a lot of memory

brendan-m-murphy (commented):
This needs to be updated with the current preferred aggregation error options from verification-games, or should be designed to make that an easy extension.

No new features should be added to the hbmcmc submodule unless strictly necessary for modern run rhime additions (so the changes to hbmcmc should preserve its behavior and not add new features).

PR #518, opened 2026-07-21
Fix dtype bug #517

pearson01:
Incorrect dtype int for scalemap_mu and scalemap_mode. Both these scaling maps should have dtype float. Added bug fix.

- [ ] Closes #517
- [ ] Tests added and passing


brendan-m-murphy (approved):
This looks good. Don't waste your copilot budget on a test.

By the way, we are deprecating the hbmcmc submodule. I won't get rid of it entirely, but there is a "modern" way to create outputs that look like the inferpymc_postprocessouts outputs. It is in the post processing submodule; at some point I could help you adapt your code to use it.

Issue #445, opened 2026-06-01
Update CI dependency installation to use uv


brendan-m-murphy:
## Summary

The main test workflow still installs dependencies with `pip`. Most local development has moved toward `uv`, so CI should be reviewed and migrated where it makes sense.

## Context

The current test workflow has an OpenGHG compatibility matrix that installs three OpenGHG targets before installing this package:

- latest minor release
- previous minor release
- `openghg@devel`

That matrix is important. A naive `uv sync` or lockfile-driven migration could accidentally replace the matrix-selected OpenGHG version with the version resolved in `uv.lock` / `pyproject.toml`, so the migration needs to preserve the explicit OpenGHG version under test.

## Things to consider

- Whether to use `uv pip install` as a low-risk first step while preserving the current matrix structure.
- Whether a full `uv sync` workflow is appropriate, and how to override the OpenGHG dependency per matrix leg if so.
- Whether test/development extras should be installed through `openghg_inversions[dev]` or kept as a minimal test-only install.
- Whether `tox` should be used in CI, possibly with `tox-uv`, now that tox is already configured. Historically tox has been too slow locally, but it may still be useful for CI if the environment setup can be made fast enough.
- How this interacts with parallel pytest execution via `pytest-xdist`.

## Outcome wanted

A CI dependency setup that uses `uv` where practical, preserves the OpenGHG compatibility matrix, remains easy to reason about, and does not make the test workflow slower.

Issue #407, opened 2026-04-29
Define high-resolution RHIME inputs with explicit inner and outer domains


brendan-m-murphy:
## Goal

Implement the 6 km / high-resolution case on the modern pathway without creating another divergent
branch.

## Tasks

- Add an `InnerDomainSpec` or equivalent as part of the RHIME run/model spec.
- Represent inner and outer basis objects/operators explicitly.
- Prepare canonical inputs with separate inner and outer sensitivity structures.
- Keep high-resolution data-preparation differences outside generic PyMC components where practical.
- Preserve the M1 pattern: run spec -> prepared run -> model builder -> `InferenceData` -> output.

## Acceptance criteria

- A prepared high-resolution run exposes separate inner and outer basis/sensitivity structures.
- The inputs can be inspected before model construction.
- The same run/output result pattern used by `run_rhime` and `run_rhime_multisector` is retained.

PR #520, opened 2026-07-21
Add generic inner/outer basis class composition

brendan-m-murphy:
* **Summary of changes**

Adds source-neutral composition primitives for already-loaded region classifications: `combine_inner_outer_region_classes` selects the mask-chosen inner or outer class with provenance tags, while `intersect_region_class_layers` forms ordered composite classes from aligned layers. Selected null/unmapped values remain unmapped, and repeated composite labels are interned.

Adds `region_class_mask` for reliable tuple-valued selection, and applies one physical-grid normalization contract at composition and constrained-basis boundaries. Equivalent rectilinear/curvilinear grids are normalized to authoritative coordinate values with nonconflicting metadata retained. Unit equality uses OpenGHG's CF-aware Pint registry (including aliases such as `Degrees_north`, `Degree_N`, and `Degrees_east`), while true unit, coordinate, calendar, unresolved grid-mapping, and CRS conflicts are rejected. Tests exercise all four packaged InTEM/land-sea domain pairs.

This branch is refreshed onto current `devel`, including the connected-basis/component-consolidation work and the split-strategy contract merged in #524. It stays within the compositor/identity/alignment scope of #449: it does not propagate arbitrary split callables through high-level wrappers and does not replace or reproduce the legacy weighted fixed-outer path. The weights-first fixed-outer adapter and coordinate-preserving loaders remain separate in stacked draft PR #525 for #452.

- [ ] Closes #449 (advances the issue; broader constrained-basis work remains)
- [x] Tests added and passing: focused constrained-basis/composition suite (144 passed)


openghg_inversions/basis/algorithms/_constrained.py
```diff
@@ -814,6 +814,84 @@ def region_constrained_basis(
...
+    Args:
+        inner_mask: Two-dimensional Boolean field selecting cells from
+            ``inner_classes``. False cells select from ``outer_classes``.
+        inner_classes: Two-dimensional class field for inner-domain cells.
+        outer_classes: Two-dimensional class field for outer-domain cells.
+        unmapped_values: Selected class values that should leave the output
+            cell unmapped. Selected null values are always unmapped.
+        name: Name for the returned ``DataArray``.
+
+    Returns:
```
brendan-m-murphy, line 1328: This output format is intuitive, but feels like it might be inefficient. Is this required by other code that creates masks using `where`? e.g. `classes.where(classes == ("inner", 1))`

If it doesn't create performance problems, we can keep this method.

docs/usage/rhime.rst
```diff
@@ -140,6 +140,29 @@ class boundaries, but RHIME config and ``run_hbmcmc.py`` do not yet load
...
+   from openghg_inversions.basis.algorithms import (
+       combine_inner_outer_region_classes,
+       region_constrained_basis,
+   )
+
+   region_classes = combine_inner_outer_region_classes(
+       inner_mask,
+       inner_land_sea_classes,
+       outer_region_classes,
+   )
```
brendan-m-murphy, line 537: I think this needs some clarification, since we also have e.g. InTEM outer regions, which are meant to be basis regions, rather than classes. Also it would be useful if e.g. we could use the existing land/sea maps for inner_land_sea_classes; I think this is possible, but from this example it sounds like the land/sea mask should be restricted to the inner region to create `inner_land_sea_classes`, which is extra work (the kind that is often annoying).

Also, requiring outer_region_classes seems like a burden, since typically we use e.g. fixed InTEM outer regions, which are not split by land/sea. So this would require explicitly passing a dummy `outer_region_classes`.

openghg_inversions/basis/algorithms/_constrained.py
brendan-m-murphy: Benchmarking makes this concern stronger. The tuple representation is not needed for normal `xarray.where` selection; in fact, it does not support it:

```python
classes == ("inner", 1)                    # all False
classes.where(classes == ("inner", 1))    # all NaN
classes.isin([("inner", 1)])               # all False
```

The tuples are being used to prevent equal raw inner/outer labels from colliding. The downstream algorithm therefore has a special `_class_value_mask` implementation for tuple comparison rather than relying on ordinary xarray/NumPy equality.

The current loop also constructs a separate tuple (and commonly a NumPy scalar object) for every mapped cell. A direct benchmark gave:

| cells | current implementation | cached unique tuple labels |
| ---: | ---: | ---: |
| 1,000,000 | 0.59 s, +113 MB peak RSS | 0.31 s, +11 MB |
| 5,000,000 | 3.02 s, +554 MB peak RSS | 1.52 s, +46 MB |

Caching one object per unique `(domain, value)` would preserve the current keys while removing most of the immediate allocation cost. If this public DataArray is expected to support ordinary selection, though, the representation itself should change (for example categorical/integer codes plus a mapping, or a dedicated scalar label type), or an explicit tested selection helper should be provided. The same per-cell tuple pattern in `intersect_region_class_layers` should be considered in the same fix.

docs/usage/rhime.rst
brendan-m-murphy: To make the compatibility concern concrete, the standard case today is asymmetric:

1. the InTEM outer labels are already-final basis regions, with one state per outer label;
2. only the inner region is generated into `nbasis` regions;
3. the current weighted/bucket inner algorithm applies the existing land/sea split automatically; and
4. `nbasis` means the requested inner-region count, not a total to redistribute across the fixed outer regions.

A whole-domain land/sea field should be accepted as the inner classification without preprocessing, because values outside `inner_mask` are unselected. Likewise, the InTEM field can supply the outer IDs rather than a dummy field. However, the documented integer `nbasis=150` call treats every tagged inner and outer value as an ordinary splittable class, so it does not preserve the standard fixed-outer semantics.

In the short term there should be an obvious compatibility path through the code introduced here. I think the acceptance criterion should be that `fixed_outer_regions_basis(..., basis_algorithm="weighted")` can be refactored to delegate to this shared composition/allocation path (or a common lower-level primitive) while preserving:

- one basis region per InTEM outer label;
- automatic use of the existing land/sea field inside the inner mask;
- `nbasis` as the generated inner count;
- weighted/bucket partition behavior and existing label/output semantics; and
- the existing saved-basis regression test.

If the constrained default or its compatibility strategy cannot reproduce that behavior, the difference needs to be explicit and tested. Otherwise we retain two subtly different ways to request what users will understand as the same fixed-outer plus land/sea-inner basis. The usage documentation should lead with this standard case, then explain the fully generic class-composition case separately.

tests/test_constrained_basis_algorithm.py
```diff
@@ -132,6 +133,147 @@ def test_region_constrained_basis_splits_zero_weight_classes_by_area():
...
+    weights = xr.DataArray(np.ones((2, 4)), dims=("lat", "lon"))
+    inner_mask = xr.DataArray(
+        np.array([[True, True, False, False], [True, True, False, False]]),
+        dims=weights.dims,
+    )
+    inner_classes = xr.DataArray(np.full(weights.shape, "shared", dtype=object), dims=weights.dims)
+    outer_classes = xr.DataArray(np.full(weights.shape, "shared", dtype=object), dims=weights.dims)
+    classes = combine_inner_outer_region_classes(inner_mask, inner_classes, outer_classes)
+
+    labels = region_constrained_basis(
```
brendan-m-murphy, line 373: This integration test has no spatial coordinates, so it misses an end-to-end alignment failure. `combine_inner_outer_region_classes` validates the mask and class arrays against one another, but `region_constrained_basis` subsequently uses `xr.align(join="exact")`, which does not compare ordinary 2-D auxiliary latitude/longitude coordinates. I reproduced combined classes on 50–51° being accepted positionally against weights on 60–61°, after which the output silently inherited the weight coordinates. Please apply the auxiliary-grid validator at the weights/classes boundary and add mismatched and transposed curvilinear-weight cases here.

openghg_inversions/basis/algorithms/_constrained.py
```diff
@@ -994,6 +1072,41 @@ def _align_2d_inputs(
...
+
+    for coordinate_name in reference_coordinates:
+        reference_coordinate = reference.coords[coordinate_name]
+        candidate_coordinate = candidate.coords[coordinate_name]
+        if set(reference_coordinate.dims) != set(candidate_coordinate.dims):
+            raise xr.AlignmentError(
+                f"Coordinate {coordinate_name!r} on {candidate_name} does not align with inner_mask."
+            )
+        candidate_coordinate = candidate_coordinate.transpose(*reference_coordinate.dims)
+        if not reference_coordinate.variable.equals(candidate_coordinate.variable):
```
brendan-m-murphy: `Variable.equals` ignores coordinate attributes and accepts equivalent numeric values with different dtypes. For example, identical values marked `degrees_north` on the mask and `radians` on a class field pass this "exact" check, and the class metadata is silently discarded. Scalar CRS/grid-mapping coordinates are also excluded above. If "exact grid agreement" is a physical-grid contract, use `identical` or explicitly compare the relevant units/calendar/CRS metadata and add regression coverage; otherwise narrow the documented guarantee.

brendan-m-murphy (commented):
Additional review after exercising the new public path:

- The standard fixed-InTEM-outer plus land/sea-inner compatibility requirement is clarified in the existing documentation thread.
- The tuple-label thread now includes measured runtime/memory results and confirms that ordinary xarray equality, `.where`, and `.isin` do not select these labels.
- Two end-to-end grid correctness gaps are noted inline below.

The PR description should also be corrected: it says the helper rejects overlapping/non-covering assignments, but the implementation intentionally selects only the mask-chosen side, ignores the unselected side, and permits selected nulls as unmapped cells. The current implementation therefore does not perform that stated validation.

I would address the standard-path compatibility and downstream grid-alignment issue before merging. The tuple-label selection/performance contract should also be resolved or explicitly narrowed and tested.

docs/usage/rhime.rst
brendan-m-murphy: Addressed in 0193380. The docs now lead with the historical fixed-InTEM-outer compatibility route, then document generic class composition separately. Whole-domain inner classes and InTEM outer IDs work without preprocessing or dummy fields. Following the design clarification, class layout/allocation is independent from `SplitStrategy` end-to-end; the legacy weighted fixed-outer behavior remains unchanged and is explicitly distinguished from the class-local bucket strategy.

openghg_inversions/basis/algorithms/_constrained.py
brendan-m-murphy: Addressed in 0193380. Both combinators now intern repeated composite labels, and `region_class_mask` provides a tested tuple-safe selector for `.where`. Existing tuple keys remain supported by allocation. Regression tests cover selection and object reuse.

tests/test_constrained_basis_algorithm.py
brendan-m-murphy: Addressed in 0193380. The shared alignment boundary now validates attached auxiliary grid coordinates before positional NumPy work. Tests reject mismatched curvilinear weights and accept physically identical transposed grids while preserving the weight layout.

openghg_inversions/basis/algorithms/_constrained.py
brendan-m-murphy: Addressed in 0193380. Coordinate validation now includes scalar coordinates, explicitly compares dtype, and uses `Variable.identical` for values and attributes. Parametrized regressions cover units attributes, float32/float64, scalar CRS values, and CRS attributes.

docs/usage/rhime.rst
```diff
@@ -140,6 +140,81 @@ class boundaries, but RHIME config and ``run_hbmcmc.py`` do not yet load
...
+   import numpy as np
+
+   from openghg_inversions.basis.algorithms import (
+       GreedyAxisParallelSplitStrategy,
+       combine_inner_outer_region_classes,
+       region_class_mask,
+       region_constrained_basis,
+   )
+
+   region_classes = combine_inner_outer_region_classes(
```
brendan-m-murphy, line 524: **[P1] Make the documented bundled-file workflow executable.** I reproduced this example with the repository's real InTEM outer-region and land/sea files, and it fails for every packaged domain tested (EUROPE, EASTASIA, SAUSSIE, and WESTUSA). The coordinate metadata differ between independently stored files (`long_name`/units versus `longname` or missing attributes), and EUROPE also has small numeric coordinate differences. At the weight-first boundary, float32 weight coordinates versus float64 land/sea coordinates can additionally make the exact label-based `.sel` raise `KeyError: not all values found in index 'lat'`. This directly contradicts the claim that whole-domain land/sea classes and InTEM IDs work without preprocessing. Please add a common-grid normalisation/loading path and an end-to-end regression using the actual packaged files before documenting this as a usable workflow. The normalisation should adopt one authoritative spatial grid after tolerance-aware value comparison while still rejecting incompatible units/CRS. Also, `inner_land` below is unused, so this currently reads as a low-level sketch rather than an obvious standard workflow.

openghg_inversions/basis/algorithms/_constrained.py
```diff
@@ -963,35 +1112,108 @@ def allocate_nbasis_by_class(
...
+                f"{reference_coordinate.dims!r} on {reference_name}."
+            )
+        candidate_coordinate = candidate_coordinate.transpose(*reference_coordinate.dims)
+        if reference_coordinate.dtype != candidate_coordinate.dtype:
+            raise xr.AlignmentError(
+                f"Coordinate {coordinate_name!r} on {candidate_name} has dtype "
+                f"{candidate_coordinate.dtype!r}, not {reference_coordinate.dtype!r} as on "
+                f"{reference_name}."
+            )
+        if not reference_coordinate.variable.identical(candidate_coordinate.variable):
```
brendan-m-murphy: **[P1] Validate physical grid compatibility rather than complete coordinate identity.** This overcorrects the earlier metadata concern by making every coordinate attribute, exact dtype, and attached scalar coordinate part of grid identity. Independently loaded arrays with the same physical grid are rejected for harmless differences such as `long_name` versus `longname`, float32 versus float64 storage, or an incidental scalar `source`/`species` coordinate. That is why the repository's own InTEM and land/sea files fail the new public workflow. Please compare the spatial coordinate values with a precision-appropriate tolerance, validate only grid-defining metadata such as units/calendar/CRS, and then normalise to the authoritative weight grid. Scalar CRS/grid-mapping coordinates should still be checked, but arbitrary non-grid scalar/provenance coordinates should not make two spatial grids incompatible. Add both actual bundled-file coverage and regressions showing that genuinely conflicting units/CRS are still rejected.

openghg_inversions/basis/_functions.py
```diff
@@ -793,6 +842,7 @@ def fixed_outer_regions_basis(
                 "region_classes": region_classes,
                 "allocation": region_allocation,
                 "min_regions_per_class": min_regions_per_class,
+                "split_strategy": split_strategy,
```
brendan-m-murphy: **[P1] This does not yet resolve the fixed-outer composition concern.** The only change here is forwarding a strategy when `basis_algorithm == "region_constrained"`. The function still loads the InTEM map, assumes its maximum value is the inner region, calls a registry algorithm on the cropped inner mask, and splices labels back with `.loc`; `basis_algorithm="weighted"` remains the separate legacy route. The new inner/outer compositor is not used by any production path. In the eventual design, constrained composition should own masks and target allocation, dispatch a chosen generator per class, and allow each InTEM outer class to use a fixed-map/one-region generator while inner land/sea classes use an aggregation generator. For #520, I would remove this wrapper-wide strategy plumbing and keep the compositor focused; carry the strategy design to #455 and the weights-first/fixed-outer compatibility adapter to #452. Preserve the current weighted fixed-outer path explicitly for reproducibility until the replacement has a deliberate compatibility policy and frozen regression coverage.

tests/test_basis_functions.py
```diff
@@ -821,9 +914,11 @@ def test_fixed_outer_regions_can_use_region_constrained_algorithm(tmp_path):
         nbasis=4,
         country_directory=str(tmp_path),
         region_classes=region_classes,
+        split_strategy=_one_region_per_class_strategy,
     )
 
     labels = basis_func.squeeze("time", drop=True)
+    assert len(np.unique(labels.values)) == 2
```
brendan-m-murphy: **[P1] This test contains no fixed outer region.** `outer_regions` is all ones, and the production function treats the maximum label as the inner region, so the mask is true everywhere. The assertion therefore tests only that the custom strategy was forwarded to the inner algorithm; it cannot detect loss of fixed outer IDs, label collisions/offset mistakes, incorrect `nbasis` semantics, or land/sea crossings. Please use a bounded maximum-valued inner area surrounded by at least two distinct outer labels, supply inner land/sea classes, and assert that each outer ID remains exactly one disjoint output state while inner labels occupy a separate range and do not cross land/sea. If behaviour is intentionally changed from legacy weighted output, add an explicit change-policy test; otherwise retain the existing frozen weighted compatibility test.

openghg_inversions/basis/_wrapper.py
```diff
@@ -47,6 +48,7 @@ def make_basis_functions(
     contrast_tau: float | None = None,
     contrast_sigma_design: float | None = None,
     contrast_s_diag: xr.DataArray | None = None,
+    split_strategy: SplitStrategy | None = None,
```
brendan-m-murphy: **[P2] Move this public-wrapper expansion to the weights/strategy follow-ups.** The class-allocation/strategy seam already existed in the constrained core; this commit now carries an arbitrary Python callable through the retained and legacy wrappers even though it is silently irrelevant for weighted/quadtree and is not representable in normal RHIME configuration. That is material API scope better handled by #455/#452 with an explicit serialisable option model. The expansion is also incomplete: these wrappers still type `nbasis` as `int`, while constrained allocation accepts an `NbasisAllocation` mapping—the mechanism needed to say “one region for every outer class, distribute the requested targets among inner classes.” If this plumbing stays anywhere, use the correct allocation type, document the algorithm-dependent meaning, and add mapping coverage.

openghg_inversions/basis/algorithms/_constrained.py
```diff
@@ -776,18 +779,32 @@ def region_constrained_basis(
...
+        weights,
+        region_classes,
+        reference_name="weights",
+        candidate_name="region_classes",
+    )
     weight_values = _validate_weights(weights)
     class_values = region_classes.to_numpy()
     mapped_classes = _mapped_classes(class_values, unmapped_values)
-    strategy = split_strategy or GreedyAxisParallelSplitStrategy()
+    strategy = split_strategy if split_strategy is not None else GreedyAxisParallelSplitStrategy()
```
brendan-m-murphy: **[P2] Validate the newly public strategy boundary.** Once arbitrary `SplitStrategy` callables are exposed through the high-level wrappers, a malformed strategy can return the wrong shape, non-integral/NaN labels, labels outside `class_mask`, or zero labels over part of a mapped class. The loop below silently leaves those cells as global label 0, contradicting the public guarantee that mapped class cells receive basis labels. Please validate output shape, integer/positive labels on all selected cells, zero outside the class, and complete class coverage, with clear errors. If the public propagation moves to #455, carry this contract and its regression tests with it.

openghg_inversions/basis/algorithms/_constrained.py
```diff
@@ -891,6 +1000,36 @@ class tuple.
...
+        contains ``class_value``.
+
+    Notes:
+        Raw expressions such as ``region_classes == ("land", "inner")`` are
+        unreliable for object arrays because NumPy may treat the tuple as an
+        array-like operand and broadcast its elements instead of comparing the
+        tuple as one value. This helper performs elementwise tuple comparison
+        through the same path used by constrained allocation and splitting.
+    """
+    mask = _class_value_mask(region_classes.to_numpy(), class_value)
```
brendan-m-murphy, line 1501: **[P2] Interning fixes the immediate allocation spike, but not the scaling model.** The selector is needed because ordinary xarray/NumPy tuple equality broadcasts the tuple as a sequence, so this does address the original `.where` correctness concern. Reusing one tuple object per unique class also removes most of the per-cell allocation cost. However, allocation and basis generation still call the Python `frompyfunc` comparison over the entire object array once or more per class, making a many-class fixed-outer case roughly O(classes × cells) in Python. Please track factorising the composite classes once into integer codes and reusing those codes for capacities, scores, masks, and splitting. I would not grow that representation refactor in a narrowed #520, but it should travel with #455/#452 rather than treating tuple-object arrays as the durable performant interface.

docs/usage/rhime.rst
```diff
@@ -140,6 +140,81 @@ class boundaries, but RHIME config and ``run_hbmcmc.py`` do not yet load
...
+equal raw inner and outer labels distinct. Use ``region_class_mask`` to select a
+tagged tuple label because ordinary NumPy/xarray comparison treats a tuple as a
+sequence instead of one object-valued class.
+
+An integer ``nbasis`` in the generic call is a total distributed across every
+selected inner and outer class. Use an explicit mapping when outer classes must
+remain fixed at one region each. This generic constrained layout does not claim
+to reproduce the legacy weighted bucket split shown above. Class composition
+and allocation are independent of the local split algorithm: replace
+``GreedyAxisParallelSplitStrategy`` with another ``SplitStrategy`` without
```
brendan-m-murphy: **[P2] Clarify the abstraction levels before advertising interchangeability.** The ideal “constrained” layer should manage classes, masks, target distribution, and global label composition; greedy, weighted, quadtree, and a fixed basis map are possible per-class basis generators. Separately, the aggregation generators can be built from interchangeable split/partition proposals, optimisation/orchestration, split acceptance, distance measures, and similar components. The current `SplitStrategy` is a full class-local dense-label generator, not merely a split step, and `AxisAlignedWeightedSplitStrategy` explicitly does not reproduce legacy weighted+land/sea behaviour. Please keep this documentation narrow in #520 and develop the broader component vocabulary/API in #455. The legacy weighted route should remain an explicit reproducibility option during migration, while quadtree's scale-sensitive optimiser should not be implied to be an equally reliable default.

CHANGELOG.md
```diff
@@ -4,6 +4,19 @@
...
+  of an inner-domain mask, tags their provenance to prevent label collisions,
+  interns repeated composite labels, and validates indexed, curvilinear, and
+  scalar grid coordinates including dtype and metadata. A companion selector
+  handles tuple-valued classes reliably. Class composition and target
+  allocation remain independent of the caller-selected local split strategy;
+  the legacy fixed-InTEM-outer weighted route remains unchanged. This advances
+  [#449](https://github.com/openghg/openghg_inversions/issues/449) and supports
+  [#407](https://github.com/openghg/openghg_inversions/issues/407),
+  [#456](https://github.com/openghg/openghg_inversions/issues/456), and
+  [#509](https://github.com/openghg/openghg_inversions/issues/509).
```
brendan-m-murphy: **[P2] Narrow the issue claims and separate the API change.** This PR creates one flat transient class field; it does not create grouped inner/outer weights, sensitivities, state layout, or `basis_group` metadata, so saying it supports #407/#456/#509 overstates what is delivered. #449 is the direct class-composition owner. If the strategy propagation remains, call it out as a separate public API change and link #455; wrapper/weights-first changes belong to #452. My preference is to move those changes out and keep this changelog entry specific to #520's compositor/selection/alignment scope.

brendan-m-murphy (commented):
I agree with the direction of separating class constraints from basis generation, but I think this follow-up has made #520 too broad and still has two blocking integration gaps. I would split this before merging.

The design I think we should work toward is:

- `region_constrained` is the composition/orchestration layer: it owns class construction, masking, allocating target region counts between classes, and globally relabelling the results.
- Each class is then handled by a selected basis generator. Greedy splitting is one option; legacy weighted/bucket and quadtree are other aggregation options; an already-fixed basis map (such as the InTEM outer labels) should also be usable as an option.
- The aggregation generators themselves can be decomposed further into interchangeable partition/split proposals, optimisation or orchestration, split-acceptance policies, distance measures, and related levers. That work belongs in the strategy-design issue, rather than being grown incidentally through every public wrapper in this PR.

One issue-number correction: #445 is the closed CI/uv migration issue. The relevant algorithm-composition issue is #455, “Define partition-step strategies for basis splitting algorithms.” The weights-first and compatibility-wrapper work belongs in #452.

My preferred split is therefore:

1. Keep #520 close to its original scope: the source-neutral inner/outer class compositor, provenance-safe class identity, the tuple allocation/selection fix, correct physical-grid alignment, and focused core tests/documentation.
2. Move the broad `SplitStrategy` propagation and aggregation-component design to a PR addressing #455.
3. Move weights-first adapters, legacy/high-level wrapper routing, configuration/serialisation concerns, and any common fixed-outer compatibility adapter to a PR addressing #452.
4. Preserve the current weighted+land/sea algorithm explicitly as a reproducibility path for existing runs while the compositional replacement is developed.

The last point matters because the current weighted land/sea behaviour is not naturally one of the clean class-first algorithms: it chooses rectangular geometry without land/sea constraints and fragments those rectangles by land/sea afterwards, while its threshold optimisation counts the post-fragmentation labels. My intuition is also that this can create many ocean/sea fragments with very little weight, but we need to retain it for reproducibility for at least a transition period. Its optimiser is not especially strong, although it is now fast enough that improving it is not urgent. Quadtree is also not a particularly safe default: its optimiser is scale-sensitive, and fixes for that have historically been lost during merges. Those facts favour treating these as explicit algorithm implementations/compatibility choices rather than making the current greedy implementation define the abstraction.

The inline comments cover the concrete blockers and follow-ups. Most importantly, the documented generic InTEM+land/sea path fails with the repository's actual packaged files, and the new fixed-outer test has no outer cells. The existing `fixed_outer_regions_basis(..., basis_algorithm="weighted")` path is still separate and unchanged; the new compositor has no production call site.

Validation is otherwise healthy: the focused suite passes all 162 tests, and Ruff and Pyright pass on the changed Python files. The missing coverage is realistic cross-file/grid integration and actual fixed-outer semantics, not the synthetic unit cases.

docs/usage/rhime.rst
brendan-m-murphy: Addressed by splitting the workflow out of #520. #520 now documents only the already-loaded core composition API and its physical-grid contract. Draft PR #525 adds coordinate-preserving InTEM/land-sea loaders plus an executable weights-first fixed-outer adapter, removes the exact-label `.sel` boundary, and exercises EUROPE, EASTASIA, SAUSSIE, and WESTUSA with the packaged files. Unit aliases are compared through OpenGHG's CF-aware Pint registry.

openghg_inversions/basis/algorithms/_constrained.py
brendan-m-murphy: Addressed in #520. The shared alignment boundary now accepts physically equivalent storage precision and benign metadata/provenance differences, normalizes to authoritative coordinates, merges nonconflicting metadata, and rejects real value, calendar, unresolved grid-mapping, CRS, and unit conflicts. Unit equality now uses `openghg.util.cf_ureg.parse_units`, including OpenGHG's known CF spelling aliases. Actual packaged-file and conflict regressions are included.

openghg_inversions/basis/_functions.py
brendan-m-murphy: Addressed by narrowing #520 and moving the integration to draft PR #525 for #452. All wrapper-wide strategy plumbing was removed from #520. #525 adds the compositional weights-first fixed-outer adapter with inner-only allocation and one state per outer ID. The historical weighted fixed-outer implementation and frozen-output regression remain unchanged for reproducibility.

tests/test_basis_functions.py
brendan-m-murphy: Addressed by removing this strategy-forwarding test from #520 and replacing the integration coverage in draft PR #525. The new synthetic case has a bounded maximum-valued inner area, two distinct outer IDs, whole-domain land/sea classes, inner-only `nbasis` allocation, disjoint outer/inner labels, and explicit no-land/sea-crossing assertions. The frozen weighted regression remains unchanged.

openghg_inversions/basis/_wrapper.py
brendan-m-murphy: Addressed by removing arbitrary `SplitStrategy` propagation from all high-level wrappers in #520. Draft PR #524 keeps validation at the existing core strategy boundary only; draft PR #525 exposes a specific weights-first fixed-outer adapter with correct inner-only integer `nbasis` semantics. Serializable RHIME/configuration routing remains deliberately deferred.

openghg_inversions/basis/algorithms/_constrained.py
brendan-m-murphy: Tracked without expanding #520. The descriptions of draft PRs #524 and #525 now explicitly defer factorizing composite class values once into reusable integer codes as part of the broader #455/#452 generator work. #520 retains only interning and the tuple-safe selector for its narrow interface.

docs/usage/rhime.rst
brendan-m-murphy: Addressed by narrowing #520's documentation to class composition, selection, and physical-grid alignment. Draft PR #524 describes `SplitStrategy` accurately as a full class-local dense-label generator and distinguishes `AxisAlignedWeightedSplitStrategy` from the historical weighted land/sea pipeline. High-level interchangeability and serializable routing remain follow-up design work.

openghg_inversions/basis/algorithms/_constrained.py
brendan-m-murphy: Moved and addressed in draft PR #524 for #455. It validates every strategy result for exact shape, integer non-Boolean dtype, positive complete coverage inside the class, and zeros outside, with focused malformed-output regressions. No arbitrary callable propagation remains in #520's public wrappers.

CHANGELOG.md
brendan-m-murphy: Addressed. #520's changelog and PR description now claim only the #449 compositor/selection/alignment scope. Strategy contract work has its own changelog and draft PR #524 for #455; weights-first fixed-outer/loaders have their own changelog and stacked draft PR #525 for #452.

openghg_inversions/basis/algorithms/_constrained.py
```diff
@@ -963,35 +1109,314 @@ def allocate_nbasis_by_class(
...
+    return coordinate
+
+
+def _spatial_coordinate_names(array: xr.DataArray) -> set[Hashable]:
+    """Return non-scalar coordinates attached only to the two grid dimensions."""
+    grid_dimensions = set(array.dims)
+    return {
+        name
+        for name, coordinate in array.coords.items()
+        if coordinate.dims and set(coordinate.dims).issubset(grid_dimensions)
```
brendan-m-murphy, line 1754: **[P1] Do not make every grid-shaped auxiliary coordinate part of physical grid identity.** This includes any coordinate whose dimensions are a subset of the two grid dimensions, so valid weights with an auxiliary measure such as `cell_area(lat, lon)` fail against otherwise identical classes that only carry `lat`/`lon`. I reproduced `region_constrained_basis` raising `AlignmentError: ... missing ['cell_area']`. `cell_area` is data attached to the grid, not a coordinate needed to identify the grid. Please restrict required identity to dimension indexes, recognized CF spatial coordinates, and referenced CRS/grid-mapping coordinates, and add a regression with an unrelated 2-D auxiliary coordinate.

brendan-m-murphy (commented):
The narrowed compositor/identity scope now makes sense. I have one remaining compatibility regression inline.

On the tuple-label question: the tuples are semantically justified because `("inner", value)` and `("outer", value)` must not collide; they are not required by `.where`—ordinary xarray/NumPy tuple equality is precisely what is unreliable here. Interning removes the earlier per-cell object-allocation spike, and `region_class_mask` gives the representation an explicit selection contract.

The remaining cost is the Python comparison path. On the packaged EUROPE fixed-outer case (114,563 cells, 8 composed classes), one full eight-mask sweep took about 0.13 s; factorising once and comparing integer codes took about 0.0045 s. Profiling the complete fixed-outer helper at `nbasis=100` put `_class_value_mask` at about 0.67 s of a 1.39 s run. That absolute cost is acceptable for the standard one-off InTEM workflow, so I would not enlarge this narrowed PR solely for factorisation. It is not a durable scaling model, however: an equal-sized synthetic field took about 2.1 s per sweep at 128 classes and 8.5 s at 512. #452/#455 should therefore contain an explicit integer-code factorisation acceptance item rather than only a prose deferral; #525 is where the repeated scans become part of a public fixed-outer workflow.

PR #523, opened 2026-07-21
Preserve missing draws in multisector totals

brendan-m-murphy:
## Summary

Fixes multisector spatial and country reconstruction when burn slicing and predictive sampling produce equal-length ArviZ groups with different draw labels. With `draws=2000, burn=1000`, retained posterior groups previously kept labels `1000..1999` while the 1,000 prior-predictive samples used `0..999`; the downstream outer merge therefore created a padded 2,000-label union.

Both modern RHIME and fixed-basis sampling now reset retained draw coordinates to `0..n_retained-1` immediately after burn slicing and before predictive groups are attached. The explicit outer merge remains as a defensive fallback for genuinely unequal external groups, and multisector totals still require a value from every sector so structural zeroes cannot hide padded missing draws.

Latest-PARIS country processing promotes projected country samples to float64 before aggregation and uncertainty statistics for both single- and multisector inversions. Total and per-sector stdev/covariance therefore use the same promoted country-level inputs, while the much larger spatial draw arrays retain their original dtype and outputs are cast only at the template boundary.

The discarded burn count is recorded on the InferenceData root and draw-bearing groups and now survives standalone trace, InversionOutput/DataTree, and fixed-basis compatibility serialization paths.

## Validation

- Full `tests/test_inferpymc.py`: 16 passed
- Full `tests/postprocessing/test_make_paris_outputs.py`: 7 passed
- Focused RHIME sampling, serialization, and latest-PARIS checks: 6 passed
- Ruff check and format checks passed
- Targeted Pyright passed; legacy files retain unrelated baseline type errors
- Full tox matrix not run locally

Real-data evidence remains in #519. Test-inversions job `18115001` passed the strict latest-PARIS multisector validator end to end in 4:01; the test definition and validator are in openghg/test_inversions#12. Its six sampling divergences make that functional/output-contract evidence only and are unrelated to the draw-coordinate mismatch.

## Checklist

- [x] Closes #519
- [x] Regression and round-trip tests added and passing
- [x] Behavior documented in `CHANGELOG.md`
- [x] No new requirements

### DIFF
--- a/openghg_inversions/postprocessing/inversion_output.py
+++ b/openghg_inversions/postprocessing/inversion_output.py
@@ -115,16 +115,38 @@ def _open_datatree_loaded(file_path: str | Path) -> xr.DataTree:
 
 
 def _inferencedata_to_datatree(idata: az.InferenceData) -> xr.DataTree:
-    """Convert an ArviZ InferenceData object to a DataTree."""
+    """Convert inference data to a serialization-safe data tree.
+
+    Args:
+        idata: Inference data whose root attributes and groups should be
+            retained. Group MultiIndexes are expanded before serialization.
+
+    Returns:
+        A data tree whose root contains ``idata.attrs`` and whose children
+        contain the serialization-safe inference-data groups.
+    """
     return xr.DataTree.from_dict(
-        {group: _reset_serialisation_multiindexes(idata[group]) for group in idata.groups()}
+        {
+            "/": xr.Dataset(attrs=dict(idata.attrs)),
+            **{group: _reset_serialisation_multiindexes(idata[group]) for group in idata.groups()},
+        }
     )
 
 
 def _inferencedata_from_datatree(dt: xr.DataTree) -> az.InferenceData:
-    """Convert a DataTree of InferenceData groups back to ArviZ InferenceData."""
+    """Restore inference data from a serialized data tree.
+
+    Args:
+        dt: Data tree whose root attributes and child groups represent an
+            ``InferenceData`` object.
+
+    Returns:
+        Reconstructed inference data with root attributes preserved and group
+        MultiIndexes restored.
+    """
     return cast(Any, az.InferenceData)(
-        **{group: _restore_serialisation_multiindexes(child.to_dataset()) for group, child in dt.items()}
+        attrs=dict(dt.attrs),
+        **{group: _restore_serialisation_multiindexes(child.to_dataset()) for group, child in dt.items()},
     )
 
 
--- a/openghg_inversions/postprocessing/inversion_output.py
+++ b/openghg_inversions/postprocessing/inversion_output.py
@@ -279,7 +279,7 @@ def convert_idata_to_dataset(
             if "chain" in trace.dims:
                 trace = trace.isel(chain=0, drop=True)
             traces.append(trace)
-    return xr.merge(traces)
+    return xr.merge(traces, join="outer")
 
 
 def _add_attributes_to_trace_dataset(trace_ds: xr.Dataset, obs_units: str, obs_longname: str) -> None:
--- a/openghg_inversions/postprocessing/make_paris_outputs.py
+++ b/openghg_inversions/postprocessing/make_paris_outputs.py
@@ -912,18 +912,32 @@ def _multisector_country_trace_kg(
         ValueError: If required multisector metadata is missing or retained
             basis, flux, and country grids cannot be aligned.
     """
-    country_trace = make_multisector_country_trace_outputs(inv_out, countries) * 1e-3
-    rename = {
-        "country_total_prior": "country_prior",
-        "country_total_posterior": "country_posterior",
-    }
+    projected_trace = make_multisector_country_trace_outputs(inv_out, countries)
+    rename = {}
     for variable_suffix, sector_name in sector_name_by_suffix.items():
         for when in ("prior", "posterior"):
             source_name = f"country_{variable_suffix}_{when}"
-            if source_name in country_trace:
+            if source_name in projected_trace:
                 rename[source_name] = f"country_{sector_name}_{when}"
 
-    result = country_trace[list(rename)].rename(rename)
+    # Promote only the projected country samples. PARIS country uncertainty
+    # statistics and totals then share one aligned float64 sector intermediate,
+    # while the much larger spatial draw arrays keep their original dtype.
+    sector_trace = projected_trace[list(rename)].rename(rename).astype(np.float64) * 1e-3
+    result = sector_trace.copy(deep=False)
+    for when in ("prior", "posterior"):
+        sector_variables = [
+            f"country_{sector_name}_{when}"
+            for sector_name in sector_name_by_suffix.values()
+            if f"country_{sector_name}_{when}" in sector_trace
+        ]
+        result[f"country_{when}"] = xr.concat(
+            [sector_trace[name] for name in sector_variables],
+            dim="sector",
+        ).sum("sector", min_count=len(sector_name_by_suffix))
+
+    result = result[["country_prior", "country_posterior", *sector_trace.data_vars]]
+    result.attrs = dict(projected_trace.attrs)
     for name in result.data_vars:
         result[name].attrs["units"] = "kg/yr"
     return result
--- a/openghg_inversions/postprocessing/make_paris_outputs.py
+++ b/openghg_inversions/postprocessing/make_paris_outputs.py
@@ -943,6 +943,31 @@ def _multisector_country_trace_kg(
     return result
 
 
+def _single_sector_country_trace_kg(
+    inv_out: InversionOutput,
+    countries: Countries,
+) -> xr.Dataset:
+    """Return a float64 single-sector country trace in kilograms per year.
+
+    Args:
+        inv_out: Single-sector inversion output.
+        countries: Country masks and cell areas used for country projection.
+
+    Returns:
+        Projected prior and posterior country traces promoted to float64 before
+        conversion from grams to kilograms per year. Dimensions, coordinates,
+        and dataset attributes are preserved, and every output variable has
+        ``units="kg/yr"``; variables typically have dimensions
+        ``(flux_time, country, draw)``.
+    """
+    projected_trace = countries.get_country_trace(inv_out=inv_out)
+    result = projected_trace.astype(np.float64) * 1e-3
+    result.attrs = dict(projected_trace.attrs)
+    for name in result.data_vars:
+        result[name].attrs["units"] = "kg/yr"
+    return result
+
+
 def _latest_country_outputs(
     inv_out: InversionOutput,
     countries: Countries,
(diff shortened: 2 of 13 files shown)


openghg_inversions/postprocessing/inversion_output.py
```diff
@@ -279,7 +279,7 @@ def convert_idata_to_dataset(
             if "chain" in trace.dims:
                 trace = trace.isel(chain=0, drop=True)
             traces.append(trace)
-    return xr.merge(traces)
+    return xr.merge(traces, join="outer")
```
brendan-m-murphy, line 304: The real run does not appear to have unequal sample counts: burn slicing leaves 1,000 posterior samples labelled `1000..1999`, while the default 1,000 prior-predictive samples are labelled `0..999`. Making the outer join explicit preserves that upstream mismatch and creates the padded 2,000-label union. Please reset the retained posterior `draw` coordinate to `0..n_retained-1` before extending predictive groups, record `burn` as metadata, and add a regression using the actual `draws=2000, burn=1000` path. This outer join can remain as a defensive fallback for explicitly unequal external groups.

openghg_inversions/postprocessing/make_paris_outputs.py
```diff
@@ -1019,9 +1019,12 @@ def _country_posterior_covariance_kg(
     )
 
     posterior = posterior.isel(flux_time=valid_indices).dropna("draw", how="all")
-    values = posterior.transpose("flux_time", "country", "draw").values
+    values = np.asarray(
```
brendan-m-murphy, line 1063: Promoting here prevents further loss during centering, but it is too late if `country_total_posterior` was already formed by summing float32 sectors. The high-magnitude test likewise constructs `total = ff + ocean` in float32, so it validates the variance of an already-rounded total without checking it against the summed sector covariance. Please cast the projected country-sector samples before aggregation and compute total/sector stdev and covariance from the same aligned float64 array, casting only at the template boundary. Keeping this limited to country-level arrays should avoid worsening the much larger spatial postprocessing memory footprint.

brendan-m-murphy (commented):
Thanks for tracing the downstream symptom and adding the `min_count=len(sectors)` protection. I think two changes are needed before merging:

1. The production failure is primarily a draw-coordinate problem, not unequal sample counts. With `draws=2000, burn=1000`, burn slicing retains posterior labels `1000..1999`, while the default 1,000 prior-predictive samples are labelled `0..999`. Both groups have the expected 1,000 samples; the outer merge turns their disjoint labels into a padded 2,000-element union. Please normalize the retained `draw` coordinate after burn, record `burn` as metadata, and add a regression for this equal-length/disjoint-label path. Keeping outer alignment as a defensive fallback for explicitly unequal external groups is reasonable.

2. Float64 calculation is worthwhile even when stored inputs are float32, but the promotion must happen before sector totals and summary statistics are calculated. Please construct the total and sector stdev/covariance from the same aligned float64 projected-country sector array, then cast only at the PARIS template boundary. To limit memory impact, this promotion should remain at the bounded country-level intermediate rather than the spatial draw arrays.

Divergences do not remove draws, so they are unrelated to the coordinate mismatch. The sector `min_count` change remains a useful defensive safeguard once the source alignment is corrected.

openghg_inversions/postprocessing/make_paris_outputs.py
brendan-m-murphy: Addressed in f73e9a6. Latest-PARIS country processing now discards the pre-rounded projected total, promotes the aligned per-sector country samples to float64, and recomputes total prior/posterior samples before stdev and covariance are calculated. The promotion stays at the bounded country-level intermediate; spatial draw arrays are unchanged and final variables are cast at the template boundary. The high-magnitude regression now forms its reference by promoting sector samples separately before summation, and checks total stdev², total covariance, and the full summed sector cross-covariance against that same reference.

openghg_inversions/postprocessing/inversion_output.py
brendan-m-murphy: Addressed in f73e9a6. `RhimeSampler.sample` now resets every retained draw-bearing group to `0..n_retained-1` immediately after burn slicing, and records `burn` on the returned trace and affected group attrs before predictive groups are extended. The regression uses the exact mocked `draws=2000, burn=1000` path and verifies posterior, sample-stat, log-likelihood, prior, and predictive coordinates are all `0..999`; `convert_idata_to_dataset` remains at 1,000 draws with no padding. The explicit outer merge remains as the fallback for genuinely unequal external groups.

openghg_inversions/rhime/sampling.py
```diff
@@ -74,6 +74,29 @@ def _log_sample_stats(trace: az.InferenceData, *, label: str) -> None:
         log_timing(label, 0.0, **fields)
 
 
+def _reset_retained_draws(trace: az.InferenceData, *, burn: int) -> az.InferenceData:
```
brendan-m-murphy, line 78: This fixes the modern RHIME path, but the same sampling sequence remains in `openghg_inversions/hbmcmc/inversion_pymc.py:250-256`: burn slicing preserves labels such as `1000..1999`, then `inferpymc()` requests equal-length prior/posterior predictive groups labelled `0..999`. That trace is used by fixed-basis modern `InversionOutput` postprocessing, so `burn > 0` still produces the original padded outer union there. Please move/reuse this reset at a shared sampling boundary and add a regression that combines nonzero burn with predictive groups for `hbmcmc.sample()`.

openghg_inversions/postprocessing/make_paris_outputs.py
```diff
@@ -1019,9 +1033,12 @@ def _country_posterior_covariance_kg(
...
```
brendan-m-murphy, line 1063: This helper applies to single-sector outputs too, but their statistics still use `countries.get_country_trace(...) * 1e-3` in float32 at lines 983-986. Covariance is now accumulated in float64 here, so high-magnitude float32 inputs can violate `diag(covariance) == stdev**2` (adjacent representable values near `1e12` can differ by order one). Please promote the single-sector projected country trace before unit conversion/statistics and have covariance consume the same float64 trace; add a high-magnitude single-sector regression.

openghg_inversions/rhime/sampling.py
```diff
@@ -74,6 +74,29 @@ def _log_sample_stats(trace: az.InferenceData, *, label: str) -> None:
...
+    Args:
+        trace: Inference data whose draw-bearing groups are relabelled in place.
+        burn: Number of discarded burn-in draws to record in metadata.
+
+    Returns:
+        The mutated inference data, with each draw coordinate reset to
+        consecutive zero-based integers and ``burn`` stored on the trace and
+        draw-bearing groups.
+    """
+    trace.attrs["burn"] = burn
```
brendan-m-murphy: Non-blocking metadata follow-up: this top-level attribute is discarded by both `_save_inferencedata()` and `_inferencedata_to_datatree()`, which reconstruct `InferenceData` from groups without passing `attrs=idata.attrs`. The posterior/sample-stat group attributes survive, but a saved-and-reloaded trace loses `reloaded.attrs['burn']`. Please either preserve top-level attrs and add a round-trip test, or document/test the group/output metadata as the durable source.

brendan-m-murphy (commented):
Final pass on `f73e9a6`: the two original review requests are correctly addressed for modern RHIME/latest multisector PARIS. The exact `draws=2000, burn=1000` regression now proves all standard groups align on 1,000 finite draws, and multisector country totals/stdev/covariance share a float64 country-level sector intermediate without promoting spatial traces. The GitHub Actions matrix is green.

I found two remaining blockers before merge:

1. The fixed-basis `hbmcmc.inversion_pymc.sample()` path still burn-slices and then attaches equal-length predictive groups without resetting draw labels. `inferpymc()` requests those groups by default and the trace feeds modern `InversionOutput` postprocessing, so `burn > 0` still reproduces the original padded outer union there. The reset should be shared by both sampling paths and covered by a combined burn-plus-predictive regression.

2. The covariance helper is shared by single- and multisector PARIS, but only the multisector statistics path is promoted before reduction. A high-magnitude float32 single-sector trace can therefore report a float32 stdev whose square disagrees materially with the newly float64 covariance diagonal. Promote the single-sector projected country trace before unit conversion/statistics and reuse that float64 input for covariance.

Non-blocking: top-level `InferenceData.attrs['burn']` is currently lost when trace/InversionOutput serialization reconstructs `InferenceData` from groups. Existing sampled-group attrs survive, but the intended durable metadata source should be explicit and round-trip tested. The PR description should also be refreshed: the production cause was equal sample counts with disjoint post-burn labels, and the test/check count has increased.

openghg_inversions/rhime/sampling.py
brendan-m-murphy: Addressed in fb037fe. Retained-draw normalization now lives in the neutral `openghg_inversions._sampling` helper and is called by both `RhimeSampler.sample` and `hbmcmc.inversion_pymc.sample` immediately after burn slicing and before predictive extension. The new fixed-basis regression exercises `draws=6, burn=3` with prior/posterior predictive groups and verifies every standard group is `0..2`, the merged trace has three finite draws, and burn metadata is present.

openghg_inversions/postprocessing/make_paris_outputs.py
brendan-m-murphy: Addressed in fb037fe. `_single_sector_country_trace_kg` promotes the projected country trace to float64 before kg conversion. `paris_flux_output_latest` constructs this trace once and passes the same object to country statistics and covariance. The high-magnitude float32 regression verifies the promoted values and `covariance == stdev² ==` the same float64 population variance; existing latest-PARIS coverage confirms final template dtypes remain float32.

openghg_inversions/rhime/sampling.py
brendan-m-murphy: Addressed in fb037fe. Root `InferenceData` attrs are now preserved by `_save_inferencedata`, both DataTree conversion directions, and the fixed-basis compatibility reconstruction/save paths; group attrs continue to survive. Standalone NetCDF and `InversionOutput` round-trip regressions assert both root and group `burn` metadata after reload.

openghg_inversions/postprocessing/make_paris_outputs.py
```diff
@@ -912,18 +912,57 @@ def _multisector_country_trace_kg(
...
+    rename = {}
     for variable_suffix, sector_name in sector_name_by_suffix.items():
         for when in ("prior", "posterior"):
             source_name = f"country_{variable_suffix}_{when}"
-            if source_name in country_trace:
+            if source_name in projected_trace:
                 rename[source_name] = f"country_{sector_name}_{when}"
 
-    result = country_trace[list(rename)].rename(rename)
+    # Promote only the projected country samples. PARIS country uncertainty
```
brendan-m-murphy: This comment should begin with motivation (or it should be placed elsewhere). So e.g. it should say "Using Float32 for these calculations can incur significant round-off error, so data is cast to float64..."

PR #524, opened 2026-07-22
Validate constrained split strategy outputs

brendan-m-murphy:
* **Summary of changes**

Hardens the existing class-local `SplitStrategy` boundary used by constrained basis generation. Explicit falsy callables are preserved, and every returned dense label array is validated for shape, integer/non-Boolean dtype, complete positive coverage inside the selected class, and zeros outside it. Strategies may still stop below their target and may use non-contiguous local labels.

Clarifies that `AxisAlignedWeightedSplitStrategy` is a class-local bucket-derived generator, not a compatibility implementation of the historical weighted land/sea pipeline.

This branch is refreshed onto current `devel`, including the connected-basis and component-consolidation work from #534, #538, #546, and #547. Validation remains at the class-local strategy boundary, before global relabelling and optional component consolidation.

This is focused groundwork for #455; it does not close the broader partition-step/orchestration design issue and does not expose Python callables through RHIME configuration or high-level wrappers. The weights-first fixed-outer integration remains separate in stacked draft PR #525 for #452.

- [ ] Closes #455 (advances the issue; further design work remains)
- [x] Tests added and passing: focused constrained-basis suite (111 passed)
- [x] Documentation and tutorials updated/added (public API docstrings)


openghg_inversions/basis/algorithms/_constrained.py
```diff
@@ -806,7 +822,10 @@ def region_constrained_basis(
     for class_value in mapped_classes:
         target_regions = targets[class_value]
         class_mask = _class_value_mask(class_values, class_value)
-        local_labels = strategy(weight_values, class_mask, target_regions)
+        local_labels = _validate_split_strategy_labels(
+            strategy(weight_values, class_mask, target_regions),
+            class_mask,
```
brendan-m-murphy, line 1281: **[P1] Enforce the allocation target as an upper bound.** The validator checks shape, dtype, coverage, and outside-mask zeros, but a strategy allocated `target_regions=1` can return four unique positive labels and pass. The global result then exceeds both the class allocation and total `nbasis`. The protocol says a strategy may stop below its target, which implies it must not overshoot. Please pass `target_regions` into `_validate_split_strategy_labels`, reject `len(_positive_labels(labels)) > target_regions`, and add an overshoot regression. If overshoot is intentionally allowed, the target and allocation APIs need to be redefined as advisory in both directions; I do not recommend that.

brendan-m-murphy (commented):
This is useful #455 groundwork, but the allocation upper-bound hole noted inline should be fixed before the strategy contract is relied on by fixed-map/one-region classes.

Two documentation/integration points should also be corrected:

- The historical weighted path does not simply optimise a layout and only then apply land/sea. Rectangle geometry ignores the land/sea constraint and is fragmented afterwards, but each threshold objective counts the post-fragmentation labels. Please use that precise distinction here and update the stale merged #486 design note that calls `AxisAlignedWeightedSplitStrategy` a compatibility strategy.
- #525 is not stacked on this PR; it is stacked on #520. The least ambiguous merge sequence is #520, then this PR rebased onto updated `devel`, then #525 rebased onto both and retargeted to `devel`. The conflict resolution needs to preserve #520's physical-grid normalization as well as this PR's strategy validation and `is not None` fix.

The `region_constrained_basis` Raises section should also retain the existing invalid-input, alignment, weight, and impossible-allocation errors instead of listing only new strategy-output failures.

PR #525, opened 2026-07-22
Add weights-first fixed-outer composition

brendan-m-murphy:
* **Summary of changes**

Adds coordinate-preserving loaders for packaged InTEM outer regions and raw country/land-sea class maps, plus a public arbitrary-reference spatial-grid normalizer backed by #520's alignment contract. Caller-supplied multi-class integer maps are preserved as distinct classes. Equal-sized reversed coordinates are physically reindexed, unrelated grid-shaped auxiliary coordinates no longer define grid identity, and genuine coordinate/unit/CRS conflicts remain errors.

Adds a weights-first `region_constrained_fixed_outer_basis_from_weights` adapter. The largest InTEM label selects the bounded inner region; `nbasis` is allocated only among selected inner classes. Each non-null outer ID remains one fixed map/state even when disconnected, while callers may choose a separate class-local `SplitStrategy` for the inner classes. Whole-domain weights are required because the output includes outer states; null outer cells remain label 0.

Factorizes class values once inside the constrained core and reuses integer codes for capacities, allocation scores, masks, and splitting while preserving public tuple identities and first-seen label order. The general weights-first constrained adapter also accepts an explicit strategy and now correctly reindexes cropped or reversed whole-domain class fields before strict physical normalization.

OpenGHG's `align_lat_lon` was evaluated but is not interchangeable: it canonicalizes one field to a named OpenGHG domain and does not provide arbitrary-reference, curvilinear, unit, CRS, or descending-coordinate handling.

This PR is refreshed and retargeted onto current `devel`, including merged #520 and #524. It is #452 groundwork, not the completed legacy fixed-outer refactor. The historical `fixed_outer_regions_basis(..., basis_algorithm="weighted")` path and frozen output remain unchanged for reproducibility. Shared high-level orchestration, serializable RHIME/config routing, explicit state metadata, and compatibility-adapter convergence remain follow-up work in #452.

- [x] Progress toward #452 (advances the issue; high-level orchestration and configuration remain)
- [x] Tests added and passing: focused basis/constrained suites (248 passed)
- [x] Independent correctness and scientific-array reviews found no remaining actionable issues
- [x] Documentation and public API contracts updated
- [x] Added any new requirements to `requirements.txt` (none required) hi


openghg_inversions/basis/_functions.py
```diff
@@ -447,6 +525,189 @@ def region_constrained_basis_from_weights(
...
+            weights,
+            region_classes,
+            reference_name="weights",
+            candidate_name="region_classes",
+        )
+
+    region_classes = region_classes.transpose(*weights.dims)
+    indexers: dict[Hashable, np.ndarray] = {}
+    for dimension in weights.dims:
+        if weights.sizes[dimension] == region_classes.sizes[dimension]:
```
brendan-m-murphy: **[P1] Reindex equal-sized coordinates too.** Skipping a dimension solely because its size matches loses the old label-based selection behavior. I reproduced weights with latitude `[50, 51]` and the same class grid stored as `[51, 50]`; this branch builds no indexer, and physical normalization then raises `AlignmentError` instead of reordering the classes. Descending latitude is a common representation of the same physical grid. Please resolve/reindex every 1-D indexed dimension, including equal-sized ones, before normalization, and add an ascending-versus-descending regression that verifies the labels follow physical cells.

openghg_inversions/basis/_functions.py
```diff
@@ -245,6 +250,79 @@ def basis_weights_from_fp_all(
...
+        landsea_path = default_directory / f"country-land-sea_{domain}.nc"
+        if not landsea_path.exists():
+            logger.warning(
+                f"No land-sea file found for domain {domain}. Defaulting to EUROPE "
+                "(country-EUROPE-UKMO-landsea-2023.nc)"
+            )
+            landsea_path = default_directory / "country-EUROPE-UKMO-landsea-2023.nc"
+
+    with xr.open_dataset(landsea_path, decode_coords="all") as dataset:
+        return dataset["country"].load()
```
brendan-m-murphy, line 429: **[P2] Enforce the advertised binary land/sea contract.** This returns the raw `country` field without validating or converting its values. A caller-supplied multi-country file therefore silently creates one inner class per country despite the function name and docstring promising binary land/sea classes. Please either validate mapped values are the supported binary set and raise clearly, convert using an explicit documented rule, or rename/document this as a generic class loader. Packaged-file tests only show that the bundled files happen to be binary.

brendan-m-murphy (commented):
This is a useful lower-level proof of the standard InTEM-outer/land-sea-inner layout, but it should be described as #452 groundwork rather than a completed `fixed_outer_regions_basis` refactor. I found one coordinate-order correctness regression inline; the binary-loader contract also needs resolving. The tuple comparison issue is measurable but not an absolute blocker for the packaged workflow, provided its scope/follow-up is made concrete.

#486 already completed the main general #452 slice: weight-first quadtree, legacy weighted/bucket and constrained helpers; `basis_weights_from_fp_all`; common finalization; and legacy `fp_all` delegation. This PR adds the missing coordinate-preserving fixed-outer sources and a weight-first compositional adapter. After it merges, #452 should still remain open for:

- convergence of the existing `fixed_outer_regions_basis` and its special wrapper branch onto shared orchestration;
- explicit per-class generator dispatch, including a fixed-map generator for outer IDs and selectable greedy/class-weighted/quadtree generators for inner classes;
- serializable config/registry routing that separates layout from generator while preserving existing `.ini` behavior;
- a public source-neutral footprint/flux weight builder (the current implementation remains private), plus any deliberately supported separate inner/outer or observation-aware weights;
- integer factorisation of composed classes, user-facing contracts/documentation, and a decision on the issue's still-unimplemented registry dataclass item.

For me, a complete fixed-outer refactor means `fixed_outer_regions_basis` becomes a thin compatibility adapter over one shared path for loading/alignment, inner-region selection, target allocation, per-class generator dispatch, global relabelling, output finalization, and state metadata. The old `basis_algorithm="weighted"` route must remain numerically reproducible—probably as an explicitly named legacy generator/path—because it chooses unconstrained rectangle geometry and then fragments it by land/sea, whereas a new class-local weighted generator optimizes land and sea separately. The public modern route should make `layout="intem_fixed_outer"`, inner generator, inner classes, and inner-only `nbasis` independently expressible. Existing configs must retain their old meaning unless users opt in.

This PR currently stops earlier: it constructs the correct classes and targets, but uses the same default greedy strategy for all classes, infers outer classes from tuple representation, emits no inner/outer state metadata, adds no config route, and leaves the legacy load/crop/run/splice implementation untouched. That is acceptable as a staged PR if those boundaries are explicit and tracked.

Please also put the whole-domain weight requirement in the public `weights` docstring and RHIME example, contrasting it with the cropped-weight support of the general constrained adapter. Loader `FileNotFoundError`/`KeyError` and null outer cells remaining label `0` belong in the public exception/result contract.

Finally, after #520 and rebased #524 merge, rebase this PR onto their combined `devel`, preserve both alignment and strategy-validation changes, retarget it to `devel`, and rerun the focused/full checks.

openghg_inversions/basis/_functions.py
```diff
@@ -447,6 +525,189 @@ def region_constrained_basis_from_weights(
...
+
+    inner_mask = loaded_outer_regions == inner_region_value
+    composed_classes = combine_inner_outer_region_classes(
+        inner_mask,
+        loaded_region_classes,
+        loaded_outer_regions,
+    )
+    inner_classes = composed_classes.where(inner_mask)
+    raw_weights = _sanitize_generated_basis_weights(weights, algorithm="fixed-outer region-constrained")
+    inner_targets = allocate_nbasis_by_class(
```
brendan-m-murphy: **[P2] Factorise composed classes once, or narrow and explicitly track the limitation.** This first allocates over tagged inner object-tuples, then `region_constrained_basis_from_weights` allocates/scans all composed tuple classes again. On packaged EUROPE (114,563 cells, 8 classes, `nbasis=100`), `_class_value_mask` took about 0.67 s of a 1.39 s run; one eight-class sweep was 0.13 s versus 0.0045 s for `pd.factorize` plus integer-code masks. That absolute one-off cost is acceptable for the standard workflow, but the public arbitrary-class path scales poorly: at the same grid size, one sweep took about 2.1 s for 128 classes and 8.5 s for 512. `.where` itself was negligible; the cost is repeated Python `frompyfunc` equality. Preserve the useful public tuple identities, but factorise once internally and reuse integer codes for capacities, scores, and masks. If that is intentionally outside this PR, constrain the advertised inputs to the measured InTEM/binary case and link a dedicated issue with this acceptance criterion rather than an open-ended #452/#455 deferral.

openghg_inversions/basis/_functions.py
```diff
@@ -447,6 +525,189 @@ def region_constrained_basis_from_weights(
...
+    outer_classes: list[Hashable] = []
+    for value in combined_values:
+        if isinstance(value, tuple) and len(value) == 2 and value[0] == "outer":
+            class_value = cast(Hashable, value)
+            if class_value not in outer_classes:
+                outer_classes.append(class_value)
+
+    targets: dict[Hashable, int] = {class_value: 1 for class_value in outer_classes}
+    targets.update(inner_targets)
+    return region_constrained_basis_from_weights(
```
brendan-m-murphy, line 907: **[P2] Keep layout composition independent from the generator.** This final call forwards neither a class-local `SplitStrategy` nor the existing split-acceptance options, so the new fixed-outer API is coupled to the default greedy generator. That is useful as a proof of layout, but it is not the intended constrained abstraction. To keep this PR small, one option is to extract a pure helper returning the composed classes plus explicit inner/outer targets; callers can then choose a generator through the constrained core. Alternatively accept/forward a strategy or serializable generator option here. Please add one non-default-generator regression when that seam is introduced, and make the current greedy-only limitation explicit if it remains deferred.

brendan-m-murphy (commented):
One further scope boundary from the independent correctness pass: the function below should either expose the generator seam or be named/documented explicitly as a greedy-only convenience. This does not require pulling RHIME configuration into #525; a pure composition/target helper would keep the PR narrow.

openghg_inversions/basis/_functions.py
brendan-m-murphy: Multiple classes in a single integer "map", like country or flat basis regions, is a useful way of specifying classes.

openghg_inversions/basis/algorithms/_constrained.py
```diff
@@ -1158,6 +1158,42 @@ def _align_2d_inputs(
     return reference, candidate
 
 
+def normalize_spatial_grid(
```
brendan-m-murphy, line 1624: There is a similar "align lat lon" function in openghg. Even if that is not suitable for our purposes, this seems like a general utility function that might be needed elsewhere (eg anywhere we use additional spatial data)

openghg_inversions/basis/_functions.py
brendan-m-murphy: Fixed. _region_classes_on_weights_grid now reindexes every eligible one-dimensional indexed dimension, including equal-sized coordinates, before physical-grid normalization. I added a regression covering reversed equal-sized latitude/longitude coordinates.

openghg_inversions/basis/_functions.py
brendan-m-murphy: Agreed with the later direction in this thread. I renamed the API to load_country_region_classes; it now preserves source values and dtype, and every distinct non-null value is treated as a class instead of enforcing or coercing a binary land/sea map. The tests include a multi-class NetCDF map and exercise it through basis generation.

openghg_inversions/basis/_functions.py
brendan-m-murphy: Implemented. The core path now factorizes mapped class identities once, in first-seen order, and reuses integer codes for capacities, bincount allocation scores, masks, and splitting. Public tuple identities are preserved; null/unmapped cells use code -1. The fixed-outer adapter allocates against raw inner identities before composing provenance, so the tuple field reaches the core factorization only once. A regression also ensures the hot path does not fall back to _class_value_mask.

openghg_inversions/basis/_functions.py
brendan-m-murphy: Addressed by separating layout from generation. The weights-first APIs now accept a SplitStrategy, and the fixed-outer adapter applies it only to inner classes. Outer IDs go through a fixed-outer wrapper and always remain one prescribed map/state even when disconnected. A connected-component regression verifies both properties. The general adapter also rejects combining an explicit strategy with a non-default contrast split_acceptance to keep the choice unambiguous.

openghg_inversions/basis/algorithms/_constrained.py
brendan-m-murphy: I evaluated openghg.util.align_lat_lon. It canonicalizes one object against a named OpenGHG domain, so it does not cover this helper's arbitrary reference grid, curvilinear coordinates, coordinate reordering, or units/CRS validation. I documented that distinction here. Moving the general grid-normalization helper into a shared utility or upstreaming it still makes sense as a separate follow-up, but doing so in #525 would expand this PR beyond the fixed-outer composition groundwork.

openghg_inversions/basis/_functions.py
```diff
@@ -347,6 +352,81 @@ def basis_weights_from_fp_all(
...
+        coordinates and metadata.
+
+    Raises:
+        FileNotFoundError: If the selected region file does not exist.
+        KeyError: If the selected dataset does not contain ``region``.
+    """
+    if country_directory is None:
+        logger.info(f"Loading default InTEM outer region file for domain {domain}.")
+        outer_regions_path = Path(__file__).parent / f"outer_region_definition_{domain}.nc"
+    else:
```
brendan-m-murphy, line 380: The country directory isn't really a natural place to store an 'outer regions map'

openghg_inversions/basis/algorithms/_constrained.py
```diff
@@ -1902,28 +1945,92 @@ def _validate_weights(weights: xr.DataArray) -> np.ndarray:
...
+def _allocate_nbasis_by_code(
+    weight_values: np.ndarray,
+    mapped_classes: list[Hashable],
+    class_codes: np.ndarray,
+    nbasis: NbasisAllocation,
+    *,
+    allocation: AllocationMode,
+    min_regions_per_class: int,
+) -> dict[Hashable, int]:
+    """Allocate class targets using already-factorized integer class codes."""
```
brendan-m-murphy: All private helpers with many arguments should have full google style docstrings

openghg_inversions/basis/_functions.py
brendan-m-murphy: Agreed. The new weights-first path now accepts outer_regions_path as a direct NetCDF file, independently from country_directory, which is reserved for country or land/sea class maps. Already-loaded arrays take precedence over their path inputs, and that precedence is documented. The legacy fixed_outer_regions_basis shared-directory behavior is unchanged. A regression uses an arbitrarily named outer-map file and a distinct country directory to verify the routing.

openghg_inversions/basis/algorithms/_constrained.py
brendan-m-murphy: Addressed. The new private multi-argument helpers now have complete Google-style Args, Returns, and Raises sections where applicable, including _allocate_nbasis_by_code, _is_spatial_coordinate, _region_classes_on_weights_grid, and the fixed-outer strategy methods. I also ran a separate docstring audit and tightened the public loader/adapter contracts and RHIME usage wording.

PR #526, opened 2026-07-23
Move sigma data preparation to the component boundary

### LINKED ISSUE #404: Extract the first model-component data-preparation seam using sigma as the test case
## Goal

Address the design issue highlighted by #397 without doing a full component-class rewrite in the
first milestone.

## Context

Some arrays needed by PyMC components are not raw inversion inputs. For example, sigma alignment uses
model data such as `site_indicator` and `sigma_freq_index`. Those arrays are also needed later to
reconstruct derived variables such as `sigma_aligned` without storing large `pm.Deterministic`
outputs in the trace.

## Tasks

- Add a small `SigmaComponentData` or equivalent prepared-data object.
- Add a pure data-preparation helper for sigma alignment inputs, independent of PyMC.
- Make the PyMC sigma component consume this prepared-data object.
- Add a reconstruction helper that can compute aligned sigma from `InferenceData` plus model data.
- Keep this narrow; do not rewrite every component as a class in M1.

## Acceptance criteria

- Sigma model-data creation can be tested without building a PyMC model.
- The PyMC sigma component still works for the existing RHIME path.
- A test reconstructs `sigma_aligned` from trace/model data without requiring it to be saved as a
  PyMC deterministic.

brendan-m-murphy:
* **Summary of changes**

This moves sigma-specific alignment out of general inversion-input preparation and into a backend-neutral sigma boundary.

- Adds a pure `SigmaAlignment` value object with constructors for frequency-derived, explicit, and registered model-data indexes.
- Makes canonical `make_inv_inputs` output sigma-free; modern RHIME derives sigma alignment from `RhimeModelSpec` during model construction.
- Simplifies `add_sigma_component` to consume only prepared alignment plus its prior, without unreleased compatibility overloads.
- Keeps `fixedbasisMCMC` working through an explicit legacy adapter: fixedbasis preparation still adds `sigma_freq_index`, anchored to `start_date`, and `hbmcmc` translates it into `SigmaAlignment` before calling the modern builder.
- Adds a thin ArviZ reconstruction helper for observation-aligned posterior sigma values.
- Uses canonical `sigma_site_index` and `sigma_period_index` model data while preserving legacy postprocessing fallback.

`SigmaAlignment` lives at the package top level rather than under `models`: importing the existing `models` package eagerly initializes PyTensor/PyMC, which would defeat the backend-independent preparation requirement.

* **Validation**

- `PYTEST_XDIST_WORKERS=1 tox -p`
  - `py310-openghgCur`: passed
  - `py310-openghgPrev`: passed
  - `py310-openghgDev`: passed
  - `lint`: passed
- Post-rebase focused regression suite: 188 passed
- Targeted Pyright checks for the new backend-neutral and reconstruction modules: passed
- Independent architecture, implementation, and scientific/xarray reviews: no remaining actionable findings

- [x] Closes #404
- [x] Tests added and passing


openghg_inversions/inversion_data/preparation.py
```diff
@@ -232,6 +233,7 @@ def _make_inv_inputs(
         min_error=min_error,
         min_error_per_site=min_error_options.get("by_site", False),
         start_date=start_date,
+        include_sigma_freq_index=include_sigma_freq_index,
```
brendan-m-murphy: Include note in docstring to explain that this is for backwards compat. We probably want a full google-style docstring here.

openghg_inversions/component_data.py
```diff
@@ -0,0 +1,363 @@
+"""Prepare and reconstruct observation-aligned sigma component data.
```
brendan-m-murphy: I could imagine there are import order issues at play, but it seems like this file should be in the `model` submodule. (Although perhaps the issue is that the model component functions are PyMC specific, so the "PyMC agnostic" requirement puts this file outside of `model`.)

openghg_inversions/models/components.py
```diff
@@ -203,7 +204,7 @@ def add_linear_component(
 
 
 def add_sigma_component(
-    site_indicator: xr.DataArray,
+    site_indicator: xr.DataArray | SigmaComponentData,
```
brendan-m-murphy: `SigmaComponentData` is not semantically the same as a "site indicator".

brendan-m-murphy: This "component" is crying out for a class with multiple constructors. The components are functions because an older class base prototype was considered too complex. That criticism was true, so classes aren't inherently forbidden, but it is a bigger change.

brendan-m-murphy (commented):
Two main criticisms:
1. considering that this feature was implemented, but not "active", there is a lot of new code. Much more than I would expect from extracting the relevant code from `add_sigma_component`
2. model components are new and unreleased, so keeping backwards compatibility in the components is not necessary. The current signature for `add_sigma_component` is a bit of a mess.

openghg_inversions/sigma.py
```diff
@@ -0,0 +1,199 @@
...
+            obs_index.get_level_values("time").to_numpy(),
+            dims=(_OBS_DIM,),
+            coords={_OBS_DIM: index.coords[_OBS_DIM]},
+            name="time",
+        )
+    return None
+
+
+@dataclass(frozen=True, eq=False, slots=True)
+class SigmaAlignment:
```
brendan-m-murphy, line 59: This is an excellent name.

PR #541, opened 2026-07-29
Timestamp precision causing lots of  nans resolved

SutarPrasad:
* **Summary of changes**
The difference between timestamps is causing lots of nans and resolving to almost 1 timepoint for a day of inversions. The fix contains resolving fp timestamps within 1s of obs timestamps

- [x] Closes #551 (Replace xxxx with the Github issue number)
- [x] Tests added and passing


brendan-m-murphy (commented):
Thanks for tackling the timestamp mismatch and PARIS BC output semantics. I found four blockers before this is ready: the BC scaling is inconsistent with the additive retrieval correction, real coarse datetime dtypes are not actually snapped, filtering can silently disable satellite scaling, and the current head fails lint due to merge residue. The PARIS changes themselves look consistent with the templates.

openghg_inversions/inversion_data/preparation.py
```diff
@@ -939,6 +939,66 @@ def _warn_for_nan_inputs(inv_inputs: xr.Dataset, *, use_bc: bool) -> None:
...
+    if not bool(satellite_mask.any()):
+        return inv_inputs
+
+    with xr.set_options(keep_attrs="default"):
+        raw_column = (
+            inv_inputs["mf"]
+            + inv_inputs["mf_prior_factor"]
+            + inv_inputs["mf_prior_upper_level_factor"]
+        )
+        scale = xr.where(raw_column > 0, inv_inputs["mf"] / raw_column, 1.0)
```
brendan-m-murphy, line 993: **[P1] BC correction uses the observation in the forward operator.** The prior factors are an additive retrieval correction, but this scales `H_bc` by `mf / (mf + prior factors)`. That makes the forward operator depend on the observed concentration and incorrectly varies the fixed prior contribution with inferred BC coefficients, potentially biasing BC/emissions partitioning. Please derive `H_bc` in the appropriate averaging-kernel/pressure-weighted column space or model the factors as a fixed additive term, and validate against the retrieval equation rather than only asserting this ratio.

openghg_inversions/inversion_data/scenario.py
```diff
@@ -10,6 +14,102 @@
...
+
+    targets, counts = np.unique(indexer[matched_positions], return_counts=True)
+    ambiguous_targets = set(targets[counts > 1])
+    positions_to_snap = [
+        position for position in matched_positions if indexer[position] not in ambiguous_targets
+    ]
+    if not positions_to_snap:
+        return
+
+    snapped_times = fp_time.values.copy()
```
brendan-m-murphy: **[P1] Coarse datetime dtype defeats timestamp snapping.** `fp_time.values.copy()` retains `datetime64[s]`/`[us]` precision, so assigning a nanosecond observation timestamp truncates it back to the footprint precision and the coordinates remain unequal. I reproduced this with a genuine `datetime64[us]` coordinate. The tests use `.floor()`, which retains `datetime64[ns]` and misses this case. Please promote to a sufficiently fine common dtype and add an explicitly coarse-dtype test.

openghg_inversions/inversion_data/preparation.py
```diff
@@ -1467,6 +1527,11 @@ def prepare_fixedbasis_inversion_data(
         calculate_min_error=calculate_min_error,
         min_error_per_site=min_error_options["by_site"],
     )
+    inv_inputs = _scale_satellite_bc_sensitivity_to_column_signal(
+        inv_inputs,
+        sites=prepared_sites,
+        platform=platform,
```
brendan-m-murphy: **[P1] Site filtering silently disables satellite scaling.** This pairs retained `prepared_sites` with the original `platform` sequence. If filtering removes a site, `_platform_by_site` sees a length mismatch and maps every retained site to `None`, silently skipping the scaling. Please pass `prepared_site_options.platform` here and `filtered_merged.site_options.platform` in the RHIME path, with mixed-platform coverage where one site is removed.

openghg_inversions/inversion_data/scenario.py
```diff
@@ -1,3 +1,7 @@
+import logging
```
brendan-m-murphy: **[P1] Merge residue blocks the lint gate.** The module docstring now sits between import groups, producing four Ruff `E402` failures, and the logger/helper definitions are duplicated later in this file. Please restore the docstring at line 1 and retain one import block, logger, and helper definition. This currently causes the lint and CI-gate jobs to fail.

openghg_inversions/inversion_data/preparation.py
brendan-m-murphy: I think this is a deliberate hack in this case, since the info required to do this properly is not available. @SutarPrasad is that right?

openghg_inversions/inversion_data/scenario.py
SutarPrasad: applied the lint change

openghg_inversions/inversion_data/preparation.py
SutarPrasad: yes

openghg_inversions/inversion_data/scenario.py
SutarPrasad: promotted to datetime64[ns] and not datetime64[us] now

openghg_inversions/inversion_data/preparation.py
SutarPrasad: Filtered site metadata now drives satellite BC

brendan-m-murphy (approved):
Looks good!

PR #559, opened 2026-08-08
Prototype PyMC 6 / ArviZ 1 DataTree migration

brendan-m-murphy:
## Summary

Prototype for #443, kept separate from the immediate dependency-pin PR #558.

- raise the minimum Python version to 3.12
- target PyMC 6 and ArviZ 1
- use `xarray.DataTree` as the runtime trace container
- add `openghg_inversions.serialization.load_trace` for standalone current and legacy group-based trace files
- preserve root/group metadata and restore serialized MultiIndexes
- keep complete `InversionOutput` artifacts on their existing owning loader

## Design explored

This prototype does **not** add an internal wrapper around DataTree. The application boundary speaks `xr.DataTree` directly, with a small load/serialization adaptor at the persistence boundary.

That has a smaller maintenance surface and follows PyMC 6 / ArviZ 1's native representation. The tradeoff is that DataTree API changes are visible in application code and tests.

The alternative of vendoring ArviZ's old NetCDF loader would isolate upstream changes more strongly, but would copy backend/group-selection behavior that xarray already provides and would need ongoing maintenance. The current adaptor instead relies on the stable NetCDF group layout and keeps only OpenGHG-specific compatibility logic locally.

A third option is to retain a project-owned wrapper. That would provide the strongest API insulation, but introduces a second trace abstraction and conversion/debugging burden; this prototype is intended to show that it is not necessary for opening old data.

## Compatibility observations

- PyMC 6.2.0 resolves with ArviZ 1.2.0.
- PyMC 6 requires Python 3.12+, so this is an environment rebuild for existing users.
- Existing trace files remain readable because old `InferenceData.to_netcdf` artifacts use the same root/group structure that `xr.open_datatree` reads.
- ArviZ remains in use for statistics such as HDI calculation; the migration only removes `InferenceData` as the container API.

## Validation

- Python 3.12: `uv run --python ... --locked pytest -q tests/test_trace_serialization.py` — 2 passed
- focused Pyright on the migrated persistence/sampling boundary — clean
- `ruff check openghg_inversions tests/test_trace_serialization.py` — passed
- Ruff format check on changed implementation files — passed
- `uv lock --check` and `pixi lock --check` — passed
- `git diff --check` — passed

## Known prototype gaps

- The broad test suite still contains many direct constructions and annotations using `az.InferenceData`; those tests need a mechanical DataTree migration before this is merge-ready.
- The full OpenGHG compatibility matrix has not been run for this draft.
- CI will help expose any PyMC 6 statistical/API changes beyond the trace-container boundary.

Refs #443

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -31,10 +31,10 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Set up Python 3.11
+      - name: Set up Python 3.12
         uses: actions/setup-python@v5
         with:
-          python-version: '3.11'
+          python-version: '3.12'
 
       - name: Set up uv
         uses: astral-sh/setup-uv@v5
--- a/.github/workflows/publish-pypi.yaml
+++ b/.github/workflows/publish-pypi.yaml
@@ -28,7 +28,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.11'
+          python-version: '3.12'
       
       - name: Build package
         run: uv build
--- a/.github/workflows/test-publish.yaml
+++ b/.github/workflows/test-publish.yaml
@@ -23,7 +23,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.11'
+          python-version: '3.12'
       
       - name: Build package
         run: uv build
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -47,7 +47,7 @@ jobs:
       max-parallel: 3
       matrix:
         os: ["ubuntu-latest"]
-        python-version: ["3.10"]
+        python-version: ["3.12"]
         openghg-version: ["${{ needs.openghg-version.outputs.latest_minor }}", "${{ needs.openghg-version.outputs.previous_minor }}", "${{ needs.openghg-version.outputs.devel }}"]
     steps:
       - name: check versions
@@ -88,10 +88,10 @@ jobs:
         with:
           version: "latest"
           enable-cache: true
-      - name: Set up Python 3.10
+      - name: Set up Python 3.12
         uses: actions/setup-python@v5
         with:
-          python-version: "3.10"
+          python-version: "3.12"
       - name: Install tox
         run: uv tool install tox --with tox-uv
       - name: Run lint
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@
 
 ## Code changes
 
+- Prototyped the PyMC 6 and ArviZ 1 migration using ``xarray.DataTree`` as the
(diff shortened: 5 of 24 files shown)


brendan-m-murphy:
This prototype now has broad architectural drift from current `devel`; if it proceeds, treat the rebase as a migration refresh rather than resolving markers mechanically.

- Preserve PR #681’s released-package-first/Pixi-contributor installation structure and update only the adopted Python/runtime requirements.
- Audit active guide pages—especially `co2_model_family.rst`, `co2_models.rst`, `rhime_standard_tutorial.rst`, `rhime.rst`, and README—for the actual `xarray.DataTree` return/load contract.
- Put compatibility loading for pre-ArviZ-1 artifacts in `legacy_and_migration.rst`, not only the no-longer-primary `getting_started.rst`.
- Reconcile current output/sampling/serialization contracts first, then regenerate both lockfiles.

Please rerun the full Python-3.12 compatibility matrix, lock checks, trace serialization, model-family docs, Ruff/type checks, `git diff --check`, and full Sphinx; search the active built guide for stale `InferenceData`, ArviZ-loader, and Python-3.10 claims.

brendan-m-murphy:
## Refresh assessment (2026-09-22)

I attempted to rebase this branch onto current `devel` (`9bdf7d83`). The single prototype commit now produces 12 content/modify-delete conflicts. I did not force-push a conflict resolution because the resulting change would not be internally coherent.

The Python 3.12 part has already landed through #690 and should be removed from this PR. The direct `xr.DataTree` architecture is still the right destination, but the implementation needs a fresh port onto current `devel`:

- current RHIME sampling/output/serialization code has changed substantially, and `rhime/runner.py` has been replaced;
- newer standard, multisector, staged, CO2, CO2/O2, and cached-sigma paths still expose or mutate `InferenceData` and use removed ArviZ 1 APIs such as callable `.groups()`, `.extend()`, and `.add_groups()`;
- the current serialization boundary now includes `save_inferencedata`/`load_inferencedata` and stronger MultiIndex handling, which must be retained and adapted rather than replaced by the prototype snapshot;
- CO2 trace annotation would silently lose scientific-role/unit metadata unless child DataTree nodes are converted or handled natively;
- the ArviZ 1 statistics boundary (`hdi`, `rhat`, `summary`, `r2_score`) still needs an independent compatibility audit;
- the added "legacy" loader test writes a new DataTree rather than opening a genuine pre-ArviZ-1 fixture, and there is no Zarr coverage;
- release documentation should use a Towncrier fragment rather than editing `CHANGELOG.md` directly.

The original CI already reported 103 failures and 10 errors; current `devel` has added more trace consumers since then. Recommendation: keep this as a prototype/reference, and replace its commit with a fresh, focused migration series on `devel` (serialization/legacy fixtures; runtime consumers; ArviZ statistics; dependency locks/docs). The dependency bump and runtime cutover should still merge atomically.

brendan-m-murphy:
Superseded by #716

Issue #565, opened 2026-08-09
Add a labelled correlated LogNormal state block with coherent marginalization


brendan-m-murphy:
## Goal

Add a backend-neutral and PyMC-backed joint state component for gathered RHIME states whose positive, mean-one coefficients have a labelled dense arithmetic covariance.

This is the highest-priority model gap exposed by the recent 14-site coherent RHIME runs. It should be usable by the standard OGI model builder rather than requiring a downstream complete-model replacement.

## Required behaviour

- Accept a gathered state coordinate, preserving a MultiIndex such as `(source, region_in_source)`.
- Accept an arithmetic mean vector and labelled arithmetic covariance `C_alpha`.
- For positive mean-one states, construct the latent Gaussian covariance elementwise as `Sigma_z = log1p(C_alpha)` and latent mean `mu_z = -0.5 * diag(Sigma_z)`.
- Fail clearly when dimensions, labels, symmetry, finiteness, or positive-definiteness are invalid.
- Use a whitened latent parameterization and expose the physical correlated state as the public effective state.
- Permit covariance across regions and across source/sector blocks.
- Record serializable variable roles, state labels, prior interpretation, and moment-conversion provenance.

## Activity semantics

Do not overload current fixed-value `StateActivity` semantics.

The component must distinguish:

1. structural padding;
2. configured deterministic fixing;
3. states removed by coherent marginalization.

For coherent marginalization, select the correct labelled marginal covariance at the reduction boundary. Omitted uncertainty may already be represented in an unresolved covariance term; it must not be reconstructed as a constant or added twice.

## Acceptance criteria

- Independent diagonal covariance reproduces the existing mean-one array LogNormal moments.
- A non-diagonal gathered two-source fixture reproduces requested prior means and covariances by analytic calculation and Monte Carlo tolerance.
- Reordered state coordinates are either aligned by labels or rejected; no positional reinterpretation.
- Cross-source covariance is retained after model compilation and trace save/load.
- Fixed and coherently marginalized states have distinct explicit representations and tests.
- Existing standard RHIME priors and public builders remain compatible.
- Documentation includes the arithmetic-to-latent moment contract and the difference between fixing and marginalizing.

## Relationships

- Follow-up under #528.
- Coherent `B/H/Pi` preparation is a separate issue; this component consumes its labelled `C_alpha`.
- Current dense/low-rank aggregation covariance support remains unchanged.

brendan-m-murphy:
Delivered across:
[...]

PR #571, opened 2026-08-09
Add correlated LogNormal state foundation

brendan-m-murphy:
## What changed

This starts #565 with the independently mergeable correlated-state foundation:

- adds a backend-neutral `CorrelatedLognormalPrior` arithmetic-moment contract with one public validating constructor;
- validates gathered state labels, covariance row/column identity, symmetry, PSD, and transformed latent positive-definiteness;
- owns mutable xarray/NumPy inputs and returns independent array/dataset copies;
- converts general positive arithmetic means/covariance to a whitened correlated LogNormal state;
- explicitly accepts an already-reduced dense covariance and emits an operational warning above 1,000 sampled states;
- adds `add_correlated_lognormal_state` as a public PyMC component;
- serializes and revalidates both arithmetic and latent moment spaces through OGI's MultiIndex boundary;
- documents that coherent covariance/forward-model/aggregation-error identities are exact only for a jointly Gaussian state, while a LogNormal retained state plus Gaussian residual is a moment-matched closure;
- keeps known-exact `StateActivity` fixing distinct from uncertainty marginalization.

The earlier standalone prior-marginal selection and retained/omitted ledger have been removed. Prior transformation, state removal, forward-model transformation, and aggregation error must be implemented together.

## Why

The recent coherent 14-site RHIME runs use one gathered state with cross-region and cross-source covariance. Existing per-sector array LogNormal priors are independent.

This foundation lets a custom OGI model builder consume an already-reduced dense state covariance immediately. It does not accept or realize a native-grid covariance: native covariances with tens or hundreds of thousands of cells must remain structured (for example, as Kronecker products) and be projected without constructing the full dense native matrix.

## Validation

- focused correlated-state regression suite after the final API cleanup: 33 passed;
- broader correlated/state/coordinate regression suite: 73 passed;
- Ruff: passed;
- Pyright on the new correlated-state module: passed;
- source-tree Sphinx build: passed (unrelated existing import warnings remain);
- `git diff --check`: passed;
- full required `tox -p --parallel-no-spinner`: passed before the final alias-only cleanup; the focused suite was rerun afterward.

## Deliberately remaining in #565 / #566

This draft does not perform native-to-reduced uncertainty transformation, state-coordinate removal, or aggregation-error construction. #566 will own that coherent unit and its explicit Gaussian/moment-closure semantics.

It also does not yet make the joint state selectable through built-in `RhimeModelSpec`. A later slice will:

- bind a prepared correlated-prior artifact/reference into the model spec;
- create one gathered joint `x(state)` at the multisector flux-construction boundary;
- expose per-sector compatibility selectors/deterministics;
- persist variable roles and coherent-reduction provenance through the standard run output;
- add standard-runner save/load coverage.

The private compiler is intentionally not changed in this foundation; unequal source blocks and structured native covariance products need explicit labelled contracts first.

Part of #565.

### DIFF
--- a/docs/usage/rhime.rst
+++ b/docs/usage/rhime.rst
@@ -720,7 +720,7 @@ public state ``<name>``.
        dims="state",
        coords={"state": ["region-a", "region-b"]},
    )
-   prior = CorrelatedLognormalPrior.from_moments(
+   prior = CorrelatedLognormalPrior(
        mean,
        np.array([[0.16, 0.03], [0.03, 0.09]]),
    )
--- a/docs/usage/rhime.rst
+++ b/docs/usage/rhime.rst
@@ -697,6 +697,7 @@ public state ``<name>``.
 
 .. code-block:: python
 
+   import numpy as np
    import pymc as pm
    import xarray as xr
 
@@ -712,11 +713,11 @@ public state ``<name>``.
    )
    prior = CorrelatedLognormalPrior.from_moments(
        mean,
-       [[0.16, 0.03], [0.03, 0.09]],
+       np.array([[0.16, 0.03], [0.03, 0.09]]),
    )
 
    with pm.Model():
-       state = add_correlated_lognormal_state(prior, var_name="x")
+       state_result = add_correlated_lognormal_state(prior, var_name="x")
 
 This component is intentionally separate from ``StateActivity``. An inactive
 ``StateActivity`` entry is conditioned on a fixed value and restored into the
--- a/docs/usage/rhime.rst
+++ b/docs/usage/rhime.rst
@@ -676,6 +676,72 @@ baseline component.
 The legacy single-sector ``inferpymc`` / ``fixedbasisMCMC`` compatibility path
 continues to sample its full flux state and does not gain multisector behavior.
 
+Correlated Positive States And Marginalization
+----------------------------------------------
+
+``CorrelatedLognormalPrior`` is the low-level contract for one labelled joint
+positive state. It accepts an arithmetic mean vector ``m`` and covariance
+``C`` and constructs the latent Gaussian moments
+
+.. math::
+
+   \Sigma_{ij} = \log\left(1 + \frac{C_{ij}}{m_i m_j}\right),
+   \qquad
+   \mu_i = \log(m_i) - \frac{1}{2}\Sigma_{ii}.
+
+For a mean-one scale state this reduces to ``Sigma = log1p(C)``. The contract
+validates labelled state order, arithmetic covariance, and positive
+definiteness of the derived latent covariance. ``add_correlated_lognormal_state``
+then creates a whitened standard-normal ``<name>_latent`` and the positive
+public state ``<name>``.
+
+.. code-block:: python
+
+   import pymc as pm
+   import xarray as xr
+
+   from openghg_inversions.models import (
+       CorrelatedLognormalPrior,
+       add_correlated_lognormal_state,
+   )
+
+   mean = xr.DataArray(
+       [1.0, 1.0],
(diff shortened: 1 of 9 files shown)


docs/usage/rhime.rst
```diff
@@ -676,6 +676,73 @@ baseline component.
...
+positive state. It accepts an arithmetic mean vector ``m`` and covariance
+``C`` and constructs the latent Gaussian moments
+
+.. math::
+
+   \Sigma_{ij} = \log\left(1 + \frac{C_{ij}}{m_i m_j}\right),
+   \qquad
+   \mu_i = \log(m_i) - \frac{1}{2}\Sigma_{ii}.
+
+For a mean-one scale state this reduces to ``Sigma = log1p(C)``. The contract
```
brendan-m-murphy: What does `log1p(C)` mean? Should it say `Sigma = log(1 + C)`?

openghg_inversions/correlated_state.py
```diff
@@ -0,0 +1,586 @@
...
+    )
+    return _copy_state_auxiliary_coords(result, mean, state_dim=state_dim)
+
+
+def _matrix_tolerance(values: np.ndarray) -> float:
+    """Return a scale-aware numerical validation tolerance."""
+    return 1e-10 * max(float(np.max(np.abs(values))), 1.0)
+
+
+@dataclass(frozen=True, init=False)
```
brendan-m-murphy: Given the complicated constructor, what is the benefit of using a dataclass here? (Or would a normal class be about as complex, but without the benefit of listing the key attributes and their types at the top of the class definition?

Is there some other pattern/datatype that would work here? (Probably a future issue.) E.g. does pydantic make this type of problem easier to address?

openghg_inversions/correlated_state.py
```diff
@@ -0,0 +1,586 @@
...
+        mean: xr.DataArray,
+        arithmetic_covariance: xr.DataArray,
+        latent_mean: xr.DataArray,
+        latent_covariance: xr.DataArray,
+        latent_cholesky: xr.DataArray,
+        state_dim: str,
+        covariance_dim: str,
+    ) -> CorrelatedLognormalPrior:
+        """Construct an instance after the public validation boundary."""
+        instance = object.__new__(cls)
```
brendan-m-murphy: Why can't you use e.g. `return cls(mean, arithmetic_covariance,...)`? Because you need to make copies?

openghg_inversions/correlated_state.py
```diff
@@ -0,0 +1,586 @@
...
+
+    This object intentionally has no fixed-value or full-state reconstruction
+    field.  Omitted states are integrated out, not conditioned on constants.
+    """
+
+    full_prior: CorrelatedLognormalPrior
+    _retained: xr.DataArray
+    prior: CorrelatedLognormalPrior
+
+    def __init__(self, *_args: object, **_kwargs: object) -> None:
```
brendan-m-murphy: Alternate take to criticisms above: why not just use this `__init__` to do the validated construction? Is this so that `retained` can return copy (and we don't want `_retained` in the signature)?

openghg_inversions/correlated_state.py
```diff
@@ -0,0 +1,586 @@
...
+        """Require construction through validated marginal selection."""
+        raise TypeError("Use CorrelatedLognormalPrior.select_marginal(...) to construct a marginal prior.")
+
+    @classmethod
+    def _from_validated(
```
brendan-m-murphy: This should probably have a full google-style docstring (as should any reasonably complex function, private or not). "from validated" is make some specific assumptions about the inputs, since this method doesn't do any validation.

PR #577, opened 2026-08-10
Separate greedy partitioning from constrained basis orchestration

brendan-m-murphy:
* **Summary of changes**

- Keep `region_constrained_basis` focused on basis-group masks, target allocation, per-group algorithm dispatch, output validation, and global relabelling.
- Move partition geometry, concrete partition steps, split-acceptance policies, node helpers, and `GreedySplitStrategy` into `_partition.py`.
- Move `AxisAlignedWeightedSplitStrategy` and its bucket-label helpers into `_weighted.py`, demonstrating that constrained orchestration can call a non-greedy group algorithm through the same `SplitStrategy` contract.
- Extract the generic priority-queue refinement engine as the public `greedy_partitioning` API and validate malformed multi-child proposals.
- Replace the unreleased `GreedyAxisParallelSplitStrategy` name with neutral `GreedySplitStrategy`, requiring explicit composition with a `PartitionStep`.
- Preserve package-level imports while correcting internal ownership imports in contrast and consolidation code.
- Defer a quadtree group adapter: its approximate target count and irregular-mask semantics need an explicit contract rather than being folded into this mechanical separation.

The API is unreleased, so no compatibility façade is retained. The dormant verification-games consumer migration is tracked separately in https://github.com/openghg/verification-games/issues/37.

**Validation**

- Initial branch validation: `tox -p --parallel-no-spinner`
  - `py310-openghgCur`: passed
  - `lint`: passed
- Review revision: 265 focused tests passed across partitioning, constrained algorithms, and higher-level basis wrappers
- Ruff check and format check on all changed Python files: passed
- Scoped Pyright on `_partition.py`, `_constrained.py`, and `_weighted.py`: 0 errors
- `git diff --check`: passed

- [x] Closes #455
- [x] Tests added and passing


openghg_inversions/basis/algorithms/_constrained.py
```diff
@@ -112,24 +121,6 @@ def __call__(
...
-
-        Returns:
-            Child grid-node partitions. Returning one child leaves the
-            partition effectively unsplit; returning multiple children lets the
-            greedy orchestrator decide how many can fit in the target count.
-        """
-        ...
-
-
 class SplitGeometry(Protocol):
```
brendan-m-murphy: I would have expected this to be moved to the partition submodule.

openghg_inversions/basis/algorithms/_constrained.py
```diff
@@ -295,84 +286,6 @@ def coordinates(
...
-            ``children``. False freezes ``parent`` as a completed partition.
-        """
-        ...
-
-
-SplitAcceptance: TypeAlias = SplitAcceptancePolicy | TargetSplitAcceptancePolicy
-
-
 @dataclass(frozen=True)
 class MinChildWeightShare:
```
brendan-m-murphy: Again, this is related to splits, not basis masks and weights, so I would expect it to be moved to _partition.py

openghg_inversions/basis/algorithms/_constrained.py
```diff
@@ -984,17 +897,22 @@ def __call__(
 
 
 @dataclass(frozen=True)
-class GreedyAxisParallelSplitStrategy:
-    """Class-local greedy repeated-bisection strategy.
+class GreedySplitStrategy:
```
brendan-m-murphy: Again, this does not seem like it is intrinsically part of the "constrained basis" concept, which should just be splitting weights and number of basis regions over different masked regions.

PR #579, opened 2026-08-11
Satellite multisector

SutarPrasad:
* **Summary of changes**

- Added a time_resolved option for selecting integrated or high-frequency footprints per site.
- Wired it through RHIME/fixed-basis configuration, preparation, and OpenGHG footprint retrieval.
- Added satellite template support for platform, max_level, and time_resolved.
- Kept CO₂ time-resolved modelling on OpenGHG’s existing high-resolution convolution; no duplicate method was added.
- Added tests for option validation, forwarding, site alignment, and satellite multisector preparation.
- Added OCO-2 three-sector integration coverage plus a reproducible short-run configuration and documentation.

**The parts that match existing multisector behavior are unchanged:**

- flux_sources supplies the individual source fields.
- sector_sources maps those sources to output sectors.
- split_by_sectors=True asks ModelScenario for fp_x_flux_sectoral.
- Existing basis projection retains the source dimension and produces H(region, nmeasure, source).
- Existing PARIS/diagnostics logic creates total and per-sector products.

- [x] Closes #580  (Replace xxxx with the Github issue number)
- [x] Tests added and passing


brendan-m-murphy:
@SutarPrasad Sorry I've been changing things in response to some feedback that the new code was hard to understand (which was true)... so the code will look more like the old hbmcmc layout. I'll try to fix the conflicts for you once that is done (should be today). Sorry for all the churn...

brendan-m-murphy (changes requested):
Please rebase this PR before the next review; GitHub reports it conflicting and its current delta includes stale historical changes.

On current devel, the public RHIME acquisition path has moved from `rhime/runner.py` to `rhime/preparation.py`, while the authoritative preparation-option schema is also different. The branch only forwards `time_resolved` through the removed runner, so a superficial conflict resolution could accept the option but silently drop it before footprint retrieval. Please rebase it onto current devel, thread the option through the current schema and acquisition path, and add an end-to-end runner-to-`_prepare_merged_data` test against those current modules.

Please also record and validate `time_resolved` in merged-data caches (or reject selector-bearing cache reloads): integrated and high-frequency footprints have different sensitivity/time semantics, so reusing a cache across selector values silently changes the scientific result. Add a mismatch-cache regression test.

Use a Towncrier fragment rather than editing `CHANGELOG.md` directly when rebasing.

brendan-m-murphy (approved):
Basic re-review of the rebased current-path implementation found no remaining correctness issues. Explicit time_resolved selectors are validated before cache reuse, including missing/mismatched provenance, and 957ca465 now also rejects incompatible legacy sector-cache layouts before relabeling them. Focused routing/cache tests and Ruff passed. This approval covers the known selector-routing and cache-provenance behavior only; the unreachable user's unspecified bug has no reproduction and is not claimed fixed.

PR #578, opened 2026-08-11
[Superseded by #581–#583] Add coherent native covariance products

brendan-m-murphy:
> [!IMPORTANT]
> **Superseded by the reviewable stack [#581](https://github.com/openghg/openghg_inversions/pull/581) → [#582](https://github.com/openghg/openghg_inversions/pull/582) → [#583](https://github.com/openghg/openghg_inversions/pull/583). Do not merge this aggregate draft.**
>
> This PR is retained as the original OPE-17 discussion and production-validation record. The stacked PRs include additional independent-review hardening.

## Summary

- add labelled, matrix-free native covariance actions with separable exponential kernels, optional class blocking, anisotropic correlation lengths, and ordered independent-source blocks
- add the retained projection strategy boundary and the covariance-compatible restriction
  `Pi_U = (U_bucket.T B^-1 U_bucket)^-1 U_bucket.T B^-1`, preserving `U_* = U_bucket`
- compute and serialize coherent `C_alpha`, `H U_*`, `H B Pi.T`, and dense or diagonal `H B H.T` products with stable labels and separate source/view identities
- document the distinction between `Pi`, `U_*`, `U_bucket`, basis geometry, covariance actions, affine centring, and eager observation batching
- add dense-oracle, multisource, class-blocking, strict-coordinate, real datetime MultiIndex, NetCDF persistence, identity-binding, and protocol-extension tests

This is the low-level native covariance preparation foundation tracked by Linear OPE-17. It does not yet construct `B_perp`, unresolved covariance, arbitrary reporting-functional products, or the coherent reduced RHIME likelihood; those remain downstream OPE-18 work.

## Verification-games production comparison

Verification-games commit [`b976de2`](https://github.com/openghg/verification-games/commit/b976de23e11a77d1708fe89ae26acefd8d919705) tested exact PR revision `946db07f57c66e4b47a5fe787298d4ebf4ae2f02` through the public `RetainedProjectionStrategy` seam.

The January 2021 coherent 14-site production restriction is the source-blocked absolute-UOB-BASE-prior-flux-weighted regional mean

`Pi_(s,r),(s',i) = 1[s=s'] 1[i in r] |F_UOB_BASE_si A_i| / sum_(j in r) |F_UOB_BASE_sj A_j|`.

[...]

Issue #528, opened 2026-07-24
Define an extensible semantic model representation for RHIME


brendan-m-murphy:
## Goal

Define a product- and backend-neutral semantic model representation that can express current RHIME models and foreseeable extensions without treating input source names, sectors, latent states, PyMC variables, or output product names as interchangeable identities.

This is an umbrella design issue. It should guide incremental work without blocking the narrow milestone 9 closure of #402 and #403.

## Motivation

The current standard and multisector paths are useful concrete cases, but future requirements break their convenient one-to-one relationships:

- several flux sources may combine into one physical component or one latent state;
- several components may have independent states on a shared basis;
- inner/outer 6 km models introduce basis groups and state partitions that are orthogonal to sectors;
- one flux state may feed several observation channels, such as linked CO2 and O2;
- conversion coefficients may be fixed or prior-backed and may require labelled coordinate alignment;
- `total` may be a virtual flux or observation aggregate without being a source or latent state;
- generic and PARIS outputs need stable scientific identities independent of PyMC suffixes.

## Design questions

- Define distinct identities and cardinalities for input sources, source groups, flux components, basis groups, state blocks, forward terms, observation channels, virtual aggregates, and output views.
- Specify the boundary between canonical `RhimePreparedInputs`, semantic normalization, backend-neutral prepared component data, compilation plans, and backend adapters.
- Represent fixed and prior-backed term coefficients, including units, sign/direction, coordinate scope, and alignment. Record that a sampled coefficient multiplying a sampled state is bilinear in the joint parameters.
- Define source-group combination policies and validation for time, grid, units, and basis compatibility.
- Represent shared-basis, ragged/non-shared, and inner/outer state layouts without positional slice or suffix inference.
- Represent one state feeding multiple species/tracer channels with channel-local observations, baselines, error components, and likelihoods.
- Define virtual total semantics for modelled observations, physical fluxes, and uncertainty, including covariance cross terms.
- Define an explicit output manifest from semantic identities to trace variables, product-neutral coordinates, and product-specific names.

## Expected outcomes

- An ADR or equivalent design note with a small relational schema and worked examples.
- Pressure tests covering standard RHIME, grouped sources, shared and ragged sectors, 6 km inner/outer, linked CO2/O2, and combinations of those features.
- Clear ownership boundaries with existing issues and a set of independently implementable follow-up issues.
- Migration guidance that keeps current trace names and outputs compatible while removing semantic dependence on them.

## Non-goals

[...]

PR #581, opened 2026-08-11
Add matrix-free separable native covariance

brendan-m-murphy:
## Stack

1. **This PR — Slice A:** no-class separable native covariance action
2. [#582 — Slice B: class/source covariance and serialization](https://github.com/openghg/openghg_inversions/pull/582)
3. [#583 — Slice C: retained projection and covariance-product artifacts](https://github.com/openghg/openghg_inversions/pull/583)

Base: `devel`. Review this PR as the foundation of the stack.

## Summary of changes

- add labelled `NativeCovarianceAction` and `InvertibleNativeCovarianceAction` protocols
- add a matrix-free separable exponential covariance action with anisotropic length scales
- support labelled single and multiple right-hand sides for apply and direct Cholesky solve
- use an ordinary slotted, identity-based computational object rather than an array-bearing dataclass
- copy the small coordinate vectors once at the explicit eager constructor boundary; return latitude/longitude through PR #586's static `BorrowedDataArray` marker with unchanged runtime identity, alongside read-only scalar configuration
- require explicit reconstruction when parameters change, with no `dataclasses.replace` or `with_parameters` contract
- add dense-oracle, ownership, eager-boundary, and API documentation coverage

This is Slice A of [OPE-32](https://linear.app/openghg-inversions/issue/OPE-32/split-ope-17-pr-578-into-a-stacked-pr-series), delivering the first part of [OPE-17](https://linear.app/openghg-inversions/issue/OPE-17/foundation-labelled-native-covariance-actions-and-product-blocks). It supersedes the corresponding foundation portion of draft #578 and preserves the design discussion from #528.

## Review notes

Rebased onto merged #586 and updated for the ownership and execution-boundary guidance now on `devel`: the covariance action owns eager coordinate copies at construction, exposes latitude/longitude through the static borrowed-reference marker without copy-on-access or runtime wrapping, keeps scalar configuration read-only, and remains a narrow behavioural action. Independent review also drove fixes for conflicting labels, zero-sized RHS axes, eager materialization, and cached-factor consistency.

## Validation

- focused Slice A suite — 21 passed
- Ruff — passed
- full `tox -p --parallel-no-spinner` on the assembled stack — passed, including lint

## Checklist

- [x] Tests added and passing
- [x] Documentation updated
- [ ] CHANGELOG entry — included in Slice C
- [x] No new requirements

### DIFF
--- a/openghg_inversions/native_covariance.py
+++ b/openghg_inversions/native_covariance.py
@@ -1,0 +1,460 @@
+"""Labelled matrix-free covariance actions on a native latitude/longitude grid.
+
+The native scaling perturbation is denoted by ``x`` and its covariance by
+``B = Cov(x)``.  A covariance action applies ``B`` to labelled right-hand
+sides without constructing the dense native ``N x N`` matrix.  The initial
+implementation uses
+
+``B = sigma**2 * (K_lat kron K_lon)``,
+
+where ``K_axis[i, j] = exp(-abs(q_i - q_j) / ell_axis)``.  Grid
+arrays use row-major ``(latitude, longitude)`` vectorisation, so the action on
+a field ``X`` is ``K_lat @ X @ K_lon.T``.  The default correlation length is
+1.5 degrees; it is an introductory configurable value, not a universal
+scientific constant.
+
+Both :meth:`SeparableExponentialCovariance.apply` and
+:meth:`SeparableExponentialCovariance.solve` preserve all input dimensions,
+coordinates, name, and attributes. The action materialises the two
+one-dimensional covariance factors and their Cholesky factors, rather than the
+dense native covariance. Apply and solve eagerly convert each labelled
+right-hand side to a NumPy array; they do not preserve lazy Dask execution.
+Distances are coordinate-wise degrees, not geodesic or longitude-wrapped
+distances. A missing coordinate ``units`` attr is interpreted as degrees for
+compatibility with existing OGI grids.
+
+``SeparableExponentialCovariance`` is an ordinary slotted, identity-based
+object. Construction owns eager coordinate copies; coordinate properties
+return borrowed arrays whose in-place mutation is unsupported. Scalar
+configuration properties are read-only, so changed parameters require explicit
+reconstruction.
+"""
+
+from __future__ import annotations
+
+from typing import Protocol
+
+import numpy as np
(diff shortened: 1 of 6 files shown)


openghg_inversions/native_covariance.py
```diff
@@ -0,0 +1,448 @@
...
+        ...
+
+
+class InvertibleNativeCovarianceAction(NativeCovarianceAction, Protocol):
+    """Native covariance action that can also solve systems in ``B``."""
+
+    def solve(self, rhs: xr.DataArray) -> xr.DataArray:
+        """Return labelled ``B^-1 rhs`` while preserving the input layout.
+
+        Args:
```
brendan-m-murphy, line 82: Could we simplify the object model before this becomes a public API? The dataclass is working against the actual contract here: generated initialization and equality are disabled, most fields are hidden from `repr`, construction uses `object.__setattr__`, and private provenance arguments exist largely so `dataclasses.replace()` can reconstruct normalized state and derived caches. The field model consequently mixes public configuration, owned inputs, replacement bookkeeping, and compiled numerical state.

`frozen=True` provides only shallow non-rebinding; it does not make the contained xarray/NumPy objects immutable. Copying the two small coordinate vectors once at construction may be justified because cached factors depend on them. Copying again on every property access is a separate, backend-dependent policy, however, and forces internal code to bypass the public API. If callers do not need the coordinates, a narrower action interface (`native_dims`, `apply`, `solve`) would avoid exposing cache-sensitive state; if they do need independent values, an explicitly named snapshot method would make the cost clear.

My preference is therefore a normal slotted class for the compiled covariance action. Add an explicit `with_parameters()` only if real callers need reconstruction, with unambiguous inheritance rules. A small frozen configuration dataclass would make sense only if configuration has an independent lifecycle—for example serialization, comparison, or reuse apart from the compiled factors.

This follows the project guidance in [Numerical data ownership and execution boundaries — “Dataclass fit and common antipatterns”](https://github.com/openghg/openghg_inversions/blob/devel/docs/plans/numerical_data_ownership_and_execution_boundaries.md#dataclass-fit-and-common-antipatterns) and [“Ownership and public API guidance”](https://github.com/openghg/openghg_inversions/blob/devel/docs/plans/numerical_data_ownership_and_execution_boundaries.md#ownership-and-public-api-guidance).

brendan-m-murphy (commented):
One object-model concern before this becomes a public API.

PR #582, opened 2026-08-11
Add class-blocked and source covariance actions

brendan-m-murphy:
## Stack

1. [#581 — Slice A: no-class separable native covariance](https://github.com/openghg/openghg_inversions/pull/581) — merged
2. **This PR — Slice B:** class blocking, ordered independent-source blocks, and serialization
3. [#583 — Slice C: retained projection and covariance-product artifacts](https://github.com/openghg/openghg_inversions/pull/583)

Base: `devel`, including merged #581. Review the Slice B changes in this PR.

## Summary of changes

- extend the separable covariance action with class-blocked application and iterative solve
- add ordered `IndependentSourceCovariance` dispatch across source-native blocks
- keep both computational actions as ordinary slotted, identity-based objects with read-only configuration
- materialize class labels once into owned eager state while validating indexed coordinates before execution; expose optional labels through PR #586's static `BorrowedDataArray` marker with unchanged runtime identity
- add versioned, typed xarray serialization for native, class, and source covariance configurations
- return independently mutable export coordinates and reject schema-name collisions or contradictory serialized lengths
- preserve source/class label identity and auxiliary coordinates through NetCDF-safe round trips
- add class/source dense-oracle, ownership, Dask-boundary, and serialization regression tests plus API documentation

This is Slice B of [OPE-32](https://linear.app/openghg-inversions/issue/OPE-32/split-ope-17-pr-578-into-a-stacked-pr-series), continuing [OPE-17](https://linear.app/openghg-inversions/issue/OPE-17/foundation-labelled-native-covariance-actions-and-product-blocks). It supersedes the class/source/serialization portion of draft #578.

## Review notes

Rebased onto the #581 marker integration. Independent review confirmed that optional class labels narrow correctly to `BorrowedDataArray`, property access remains an identity-only cast, and serialization consumes the borrowed reference observationally. Earlier review drove fixes for typed tuple/scalar labels, source dimension validation, SciPy CG compatibility, serialized state flags, auxiliary-coordinate preservation, Dask-backed class labels, export aliasing, schema collisions, and explicit-versus-fallback length consistency. The final re-review found no blocking issues.

The class-blocked solve remains matrix-free: SciPy's flattened `N x N` operator shape is virtual, and each CG iteration applies the separable kernel through masked native-grid work arrays. The module documentation now records the repeated per-class computational cost, especially for highly imbalanced masks. Improving that case without sacrificing the memory bound is tracked as [OPE-39](https://linear.app/openghg-inversions/issue/OPE-39/improve-matrix-free-class-blocked-covariance-solves-for-imbalanced).

## Validation

- focused cumulative Slice A+B plus borrowed runtime suite — 58 passed
- Ruff — passed
- full `tox -p --parallel-no-spinner` on the assembled stack — passed, including lint

## Checklist

- [x] Tests added and passing
- [x] Documentation updated
- [ ] CHANGELOG entry — included in Slice C
- [x] No new requirements

### DIFF
--- a/openghg_inversions/native_covariance.py
+++ b/openghg_inversions/native_covariance.py
@@ -24,7 +24,15 @@ boolean mask per class, but does not construct the diagonal ``M_c`` matrices or
 the dense native covariance. Apply and solve eagerly convert each labelled
 right-hand side to a NumPy array; they do not preserve lazy Dask execution.
 Unblocked solves use the separable Cholesky factors; multi-class solves use
-matrix-free conjugate gradients. Distances are coordinate-wise degrees, not
+matrix-free conjugate gradients. The linear operator reports the flattened
+``N x N`` shape required by SciPy but does not materialise that matrix: each
+iteration applies the separable kernel once per class through masked native-grid
+work arrays. Grouping coordinates by class would expose a block-diagonal
+matrix, but arbitrary class blocks are dense principal submatrices and are not
+generally separable. The iterative solve therefore preserves the
+``O(N + n_lat**2 + n_lon**2)`` storage bound, at the cost of repeated
+per-class kernel applications; highly imbalanced classes can consequently have
+noticeable computational overhead. Distances are coordinate-wise degrees, not
 geodesic or longitude-wrapped distances. A missing coordinate ``units`` attr
 is interpreted as degrees for compatibility with existing OGI grids.
 
--- a/openghg_inversions/native_covariance.py
+++ b/openghg_inversions/native_covariance.py
@@ -15,30 +15,41 @@ scientific constant.
 
 Both :meth:`SeparableExponentialCovariance.apply` and
 :meth:`SeparableExponentialCovariance.solve` preserve all input dimensions,
-coordinates, name, and attributes. The action materialises the two
+coordinates, name, and attributes.  An unblocked action materialises the two
 one-dimensional covariance factors and their Cholesky factors, rather than the
-dense native covariance. Apply and solve eagerly convert each labelled
+dense native covariance. Optional native-grid class labels replace ``B`` by
+the blocked action ``B_class = sum_c M_c B M_c``, where ``M_c`` is the diagonal
+indicator for class ``c``. A blocked action additionally stores one native-grid
+boolean mask per class, but does not construct the diagonal ``M_c`` matrices or
+the dense native covariance. Apply and solve eagerly convert each labelled
 right-hand side to a NumPy array; they do not preserve lazy Dask execution.
-Distances are coordinate-wise degrees, not geodesic or longitude-wrapped
-distances. A missing coordinate ``units`` attr is interpreted as degrees for
-compatibility with existing OGI grids.
+Unblocked solves use the separable Cholesky factors; multi-class solves use
+matrix-free conjugate gradients. Distances are coordinate-wise degrees, not
+geodesic or longitude-wrapped distances. A missing coordinate ``units`` attr
+is interpreted as degrees for compatibility with existing OGI grids.
 
 ``SeparableExponentialCovariance`` is an ordinary slotted, identity-based
-object. Construction owns eager coordinate copies; coordinate properties
-return borrowed arrays whose in-place mutation is unsupported. Scalar
-configuration properties are read-only, so changed parameters require explicit
-reconstruction.
+object. Construction owns eager coordinate and class-label copies; array
+properties return borrowed values whose in-place mutation is unsupported.
+Scalar configuration properties are read-only, so changed parameters require
+explicit reconstruction.
 """
 
 from __future__ import annotations
 
-from typing import Protocol
+import inspect
+import json
+from typing import Callable, Protocol, cast
 
 import numpy as np
 from scipy.linalg import cho_factor, cho_solve  # type: ignore[import-untyped]
+from scipy.sparse.linalg import LinearOperator, cg  # type: ignore[import-untyped]
 import xarray as xr
 
 from openghg_inversions.borrowed import BorrowedDataArray, borrow
+
+_CG_RELATIVE_TOLERANCE = "rtol" if "rtol" in inspect.signature(cg).parameters else "tol"
+_NATIVE_SERIALIZED_VARIABLE_NAMES = {"class_label_encoded"}
 
 __all__ = [
     "InvertibleNativeCovarianceAction",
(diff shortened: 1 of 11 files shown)


openghg_inversions/native_covariance.py
```diff
@@ -426,6 +656,123 @@ def _validate_coordinate(coordinate: xr.DataArray | None, axis_name: str) -> xr.
     return coordinate
 
 
+def _validate_class_labels(
```
brendan-m-murphy, line 678: Does this share logic with the 'constrained basis' algorithm?

openghg_inversions/native_covariance.py
```diff
@@ -15,31 +15,42 @@
...
+dense native covariance. Optional native-grid class labels replace ``B`` by
+the blocked action ``B_class = sum_c M_c B M_c``, where ``M_c`` is the diagonal
+indicator for class ``c``. A blocked action additionally stores one native-grid
+boolean mask per class, but does not construct the diagonal ``M_c`` matrices or
+the dense native covariance. Apply and solve eagerly convert each labelled
 right-hand side to a NumPy array; they do not preserve lazy Dask execution.
-Distances are coordinate-wise degrees, not geodesic or longitude-wrapped
-distances. A missing coordinate ``units`` attr is interpreted as degrees for
-compatibility with existing OGI grids.
+Unblocked solves use the separable Cholesky factors; multi-class solves use
```
brendan-m-murphy, line 26: Why different methods? In some order of the coordinates the Covariance is block diagonal; does using a sparse matrix avoid this reordering?

brendan-m-murphy (commented):
It seems like some of the class related code could potentially be shared with the constrained basis logic. Not that all separate basis groups need to have their own within block covariance and no covariance outside of the group, but in terms of code, groups defined by masks are used in both cases.

openghg_inversions/source_covariance.py
```diff
@@ -0,0 +1,566 @@
...
+            TypeError: If ``rhs`` is not an xarray data array.
+            ValueError: If source or spatial labels are missing or do not
+                exactly match the configured labels, or values are invalid.
+            numpy.linalg.LinAlgError: If a requested class-blocked solve does
+                not converge.
+        """
+        if not isinstance(rhs, xr.DataArray):
+            raise TypeError("rhs must be an xarray.DataArray")
+        if self.source_dim not in rhs.dims or self.source_dim not in rhs.coords:
+            raise ValueError(f"rhs must contain labelled source dimension {self.source_dim!r}")
```
brendan-m-murphy:
```suggestion
            raise ValueError(f"rhs must contain labelled source dimension coordinate {self.source_dim!r}")
```

openghg_inversions/source_covariance.py
```diff
@@ -0,0 +1,566 @@
...
+        actual_labels = tuple(rhs.coords[self.source_dim].values.tolist())
+        if any(not isinstance(label, str) for label in actual_labels) or actual_labels != self.source_labels:
+            raise ValueError(
+                "rhs source labels/order do not match covariance configuration: "
```
brendan-m-murphy: Why not let xarray handle this? You can chunk on source and use apply_ufunc

openghg_inversions/source_covariance.py
```diff
@@ -0,0 +1,566 @@
...
+            attrs=rhs.coords[self.source_dim].attrs,
+        )
+        combined = xr.concat(results, dim=source_coordinate)
+        combined = combined.assign_coords({name: coordinate for name, coordinate in rhs.coords.items()})
+        combined.name = rhs.name
+        combined.attrs = rhs.attrs
+        return combined.transpose(*original_dims)
+
+
+def _encode_class_label(value: object) -> str:
```
brendan-m-murphy: Except for the error message, this seems like a generic JSON encoding helper. Same with the next two functions. I'm fairly certain we have other code that does this, and it should be done once, if possible.

openghg_inversions/source_covariance.py
```diff
@@ -0,0 +1,566 @@
...
+
+def _positive_finite(value: SupportsFloat, name: str) -> float:
+    """Decode one serialized positive finite covariance parameter."""
+    resolved = float(value)
+    if not np.isfinite(resolved) or resolved <= 0.0:
+        raise ValueError(f"Serialized {name} must be finite and strictly positive")
+    return resolved
+
+
+def _serialized_boolean(value: object, name: str) -> bool:
```
brendan-m-murphy: Another JSON/numpy helper?

openghg_inversions/native_covariance.py
brendan-m-murphy: I compared the two paths in detail. They share the concept of grouping cells by a label, but the policies differ: covariance requires exact native-grid alignment, complete non-null coverage, an eager owned copy, and cached masks; constrained basis permits null/explicitly unmapped cells, requires hashable allocation keys, and uses tolerant one-off comparisons during allocation/splitting. Reusing the basis helper here would also invert the core-to-basis dependency. I have kept those policy wrappers separate. If a third consumer or measured mask-construction cost justifies it, the reusable seam would be a small value-level first-seen factorizer in a neutral leaf module, not either current validator.

openghg_inversions/source_covariance.py
brendan-m-murphy: Applied this wording change.

openghg_inversions/source_covariance.py
brendan-m-murphy: Agreed that xarray should own the labelled-coordinate mechanics. `_operate` now builds the expected source index, validates with `xr.align(..., join="exact")`, selects each source with `.sel()`, and concatenates through xarray; the manual value/order comparison and blanket `assign_coords` reconstruction are gone. The original RHS is used as the result template, so auxiliary-coordinate dimensions/order and lazy coordinate graphs are preserved. I retained only the explicit per-source loop because each source has a distinct covariance action and consumes the full spatial core; source-only chunking or a single `apply_ufunc` would not provide equivalent Dask execution without first defining an ndarray kernel and rechunking contract.

openghg_inversions/source_covariance.py
brendan-m-murphy: Yes. A repository-wide search found other JSON policies, but they are intentionally lossy (`default=str`) or canonicalized for hashing rather than reversible typed values. I moved the duplicated `tagged_json_v1` encoder/decoder into a lightweight private `_serialization_codecs.py` module shared by native and source covariance. The valid wire bytes are unchanged, and focused tests now freeze every supported Python/NumPy type and nested tuples while rejecting malformed tags and payloads.

openghg_inversions/source_covariance.py
brendan-m-murphy: Consolidated with the same lightweight codec module. The NumPy-scalar JSON fallback and strict Boolean/0-or-1 decoder now have one shared implementation rather than copies in both covariance modules.

PR #583, opened 2026-08-11
Add retained native covariance product artifacts

brendan-m-murphy:
## Stack

1. [#581 — Slice A: no-class separable native covariance](https://github.com/openghg/openghg_inversions/pull/581)
2. [#582 — Slice B: class/source covariance and serialization](https://github.com/openghg/openghg_inversions/pull/582)
3. **This PR — Slice C:** retained projection and covariance-product artifacts

Base: `codex/ope-32-b-class-source-covariance`. Review only the changes after #582.

## Summary of changes

- add the retained-projection strategy protocol and bucket-preserving covariance-compatible restriction
- compute labelled `C_alpha`, `H U_*`, `H B Pi.T`, and dense/diagonal `H B H.T` without constructing dense native `B`
- add source/view/configuration identities and DataTree/NetCDF persistence
- harden product loading against malformed shapes, labels, views, namespace metadata, complex/non-finite values, and missing MultiIndexes
- make observation/state coordinate namespaces collision-safe, including shared MultiIndex level names
- document `Pi`, `U_*`, `U_bucket`, centring, batching, persistence, and the downstream OPE-18 boundary
- add the integrated usage guide and CHANGELOG entry

This is Slice C of [OPE-32](https://linear.app/openghg-inversions/issue/OPE-32/split-ope-17-pr-578-into-a-stacked-pr-series), completing the stacked delivery of [OPE-17](https://linear.app/openghg-inversions/issue/OPE-17/foundation-labelled-native-covariance-actions-and-product-blocks). Together #581 → #582 → this PR supersede draft #578.

## Production evidence and review context

The mathematical and production discussion remains recorded on #578 and GitHub #528. Verification-games commit [b976de2](https://github.com/openghg/verification-games/commit/b976de23e11a77d1708fe89ae26acefd8d919705) exercised the public strategy seam at original revision `946db07` across all 361 retained states and 14 sites, with worst OGI/VG product relative Frobenius difference `5.44e-16`. It also established that supplied physical `Pi` is production-capable while the bucket-first default is not production-equivalent; the supplied-`Pi` concrete strategy remains follow-up OPE-31.

Restacked onto the #581/#582 borrowed-marker adoption with no Slice C ownership changes; the OPE-38 and OPE-17 changelog entries are both preserved. Independent code, correctness, architecture, and scientific-array reviews were run slice-by-slice. C follow-up review drove collision-safe MultiIndex construction/serialization, finite-real enforcement, strong row/column binding, namespace-owner validation, configuration digest strengthening, and complex-input rejection.

## Validation

- Ruff — passed
- `git diff --check` — passed
- `tox -e docs` — passed with existing repository warnings
- final `tox -p --parallel-no-spinner` — current OpenGHG integration, borrowed types, and lint passed

## Checklist

- [x] Tests added and passing
- [x] No new requirements

### DIFF
--- a/docs/usage/native_covariance.rst
+++ b/docs/usage/native_covariance.rst
@@ -185,8 +185,10 @@ The structured covariance action stores axis factors rather than dense native
 :math:`B`, but the numerical product kernel is otherwise eager. The upstream
 pipeline must materialize the related sensitivity and canonical basis
 prolongation together before calling it; lazy inputs are rejected rather than
-computed implicitly. Custom restrictions may remain sparse or Dask-backed and
-are densified only in explicit retained-state right-hand-side blocks. For
+computed implicitly. A custom restriction may remain sparse or Dask-backed
+until the explicit projection boundary, where it is materialized and densified
+once. The eager restriction is then reused across retained-state
+right-hand-side blocks, avoiding repeated execution of its lazy graph. For
 native size :math:`N`, retained size
 :math:`d`, and observation count :math:`M`, important dense storage includes
 :math:`M N` for :math:`H`, :math:`N d` for each native-by-retained array,
--- a/docs/usage/native_covariance.rst
+++ b/docs/usage/native_covariance.rst
@@ -28,8 +28,10 @@ The classes have the following distinct roles.
        native source dimension. In both cases, the transpose does not define
        the retained restriction :math:`\Pi`.
    * - ``InvertibleNativeCovarianceAction``
-     - Structural interface for applying :math:`B` and solving systems in
-       :math:`B` without constructing a dense native covariance matrix.
+     - Structural interface for a labelled, self-adjoint positive-definite
+       :math:`B`, including its compatible inverse solve, without constructing
+       a dense native covariance matrix. The kernel trusts this semantic
+       contract rather than globally certifying a matrix-free action.
    * - ``SeparableExponentialCovariance``
      - Concrete latitude/longitude covariance action. Optional class labels
        set cross-class covariance to zero; this alone does not assert
@@ -41,16 +43,18 @@ The classes have the following distinct roles.
        and solves use the common action interface; only multisource basis
        expansion needs a separate source-aware path.
    * - ``RetainedProjectionStrategy``
-     - Policy interface that chooses a compatible restriction :math:`\Pi` and
-       covariance-natural prolongation :math:`U_*`.
+     - Pi-first policy interface that chooses the authoritative labelled
+       restriction :math:`\Pi`. The kernel derives its covariance-natural
+       prolongation :math:`U_*`.
    * - ``PreserveBucketProlongation``
-     - Current structural implementation of the strategy interface. It fixes
-       :math:`U_* = U_{\mathrm{bucket}}` and derives the compatible
-       :math:`\Pi_U`.
+     - Current strategy implementation. It uses
+       :math:`U_{\mathrm{bucket}}` to derive the compatible authoritative
+       restriction :math:`\Pi_U`; the kernel still derives :math:`U_*` from
+       that returned restriction.
    * - ``RetainedProjection``
      - Frozen value dataclass returned by a strategy. It carries
-       :math:`(\Pi, U_*)` and the strategy identifier; it is not a protocol,
-       and its contained xarray objects remain mutable.
+       :math:`\Pi` and the strategy identifier; it is not a protocol, and its
+       contained xarray object remains mutable.
    * - ``project_native_covariance``
      - Validates and combines :math:`H`, :math:`B`, basis geometry, and a
        projection strategy.
@@ -67,7 +71,7 @@ The data flow is:
                               ├─> RetainedProjectionStrategy
    covariance action ──> B ───┘              │
                                              v
-                              RetainedProjection(Pi, U_*)
+                              RetainedProjection(Pi)
                                              │
    native sensitivity H ─────────────────────┤
    covariance action B ──────────────────────┤
@@ -137,7 +141,9 @@ the established bucket-scaling interpretation by choosing
    U_{\mathrm{bucket}}^\mathsf{T} B^{-1}.
 
 This gives :math:`\Pi_U U_{\mathrm{bucket}} = I` and
-:math:`U_* = U_{\mathrm{bucket}}`. OPE-17 returns :math:`C_\alpha`,
+:math:`U_* = U_{\mathrm{bucket}}`; the latter identity follows from the
+kernel's generic derivation :math:`U_* = B\Pi^\mathsf{T}C_\alpha^{-1}`, not
+from a strategy-supplied lift. OPE-17 returns :math:`C_\alpha`,
 :math:`H U_*`, :math:`H B \Pi^\mathsf{T}`, and either dense
 :math:`H B H^\mathsf{T}` or its diagonal.
 
@@ -243,9 +249,10 @@ setting. Both policies require a supplied-:math:`\Pi`-first OGI strategy.
        covariance-natural lift. Current 14-site production uses the
        absolute-prior-flux-weighted regional mean above.
      - Deliberate difference
-     - The first OPE-17 strategy is compatibility-oriented: it fixes
-       :math:`U_{\mathrm{bucket}}` first and derives :math:`\Pi_U`. The strategy
-       interface leaves room for a future :math:`\Pi`-first policy.
+     - The first OPE-17 strategy is compatibility-oriented: it uses
+       :math:`U_{\mathrm{bucket}}` internally to derive :math:`\Pi_U`. The
+       strategy interface itself is Pi-first, so later policies may supply a
+       physical restriction directly.
    * - Cross-source covariance, arbitrary reporting functionals :math:`Q`,
        :math:`B_\perp` products, coherent likelihood reduction, reconstruction,
        and covariance approximation
--- a/docs/usage/native_covariance.rst
+++ b/docs/usage/native_covariance.rst
@@ -1,0 +1,257 @@
+Native Covariance Projection
+============================
+
+The native covariance projection API prepares labelled covariance and
+observation product blocks for reducing a gridded scaling state. It keeps basis
+geometry, native covariance, and retained-state semantics separate, and never
+constructs a dense native-grid covariance matrix. This is the low-level OPE-17
+boundary: it is not connected to RHIME input preparation or likelihoods, and
+its product blocks are not the complete coherent-reduction artifact planned by
+OPE-18.
+
+Component overview
+------------------
+
+The classes have the following distinct roles.
+
+.. list-table::
+   :header-rows: 1
+   :widths: 28 72
+
+   * - Component
+     - Responsibility
+   * - ``BasisOperator``
+     - Owns basis geometry and retained-state labels. For a single source,
+       ``basis_matrix`` is the bucket prolongation :math:`U_{\mathrm{bucket}}`.
+       A gathered multisource matrix is a spatial template that the projection
+       code expands onto an explicit native source dimension. In both cases,
+       the transpose does not define the retained restriction :math:`\Pi`.
+   * - ``InvertibleNativeCovarianceAction``
+     - Structural interface for applying :math:`B` and solving systems in
+       :math:`B` without constructing a dense native covariance matrix.
+   * - ``SeparableExponentialCovariance``
+     - Concrete latitude/longitude covariance action. Optional class labels
+       set cross-class covariance to zero; this alone does not assert
+       probabilistic independence outside a joint Gaussian model.
+   * - ``IndependentSourceCovariance``
+     - Composes ordered, source-specific spatial covariances into a
+       block-diagonal native covariance. Its blocks currently must be
+       ``SeparableExponentialCovariance`` instances. Covariance application
+       and solves use the common action interface; only multisource basis
+       expansion needs a separate source-aware path.
+   * - ``RetainedProjectionStrategy``
+     - Policy interface that chooses a compatible restriction :math:`\Pi` and
+       covariance-natural prolongation :math:`U_*`.
+   * - ``PreserveBucketProlongation``
+     - Current structural implementation of the strategy interface. It fixes
+       :math:`U_* = U_{\mathrm{bucket}}` and derives the compatible
+       :math:`\Pi_U`.
+   * - ``RetainedProjection``
+     - Frozen value dataclass returned by a strategy. It carries
+       :math:`(\Pi, U_*)` and the strategy identifier; it is not a protocol,
+       and its contained xarray objects remain mutable.
+   * - ``project_native_covariance``
+     - Validates and combines :math:`H`, :math:`B`, basis geometry, and a
+       projection strategy.
+   * - ``NativeCovarianceProducts``
+     - Frozen result and serialization dataclass containing the labelled
+       product blocks. Its contained xarray objects and provenance mapping
+       remain mutable.
+
+The data flow is:
+
+.. code-block:: text
+
+   BasisOperator ──> U_bucket ─┐
+                              ├─> RetainedProjectionStrategy
+   covariance action ──> B ───┘              │
+                                             v
+                              RetainedProjection(Pi, U_*)
+                                             │
+   native sensitivity H ─────────────────────┤
+   covariance action B ──────────────────────┤
+                                             v
+                              project_native_covariance
+                                             │
+                                             v
+                              NativeCovarianceProducts
+
+``FluxWeightedBasis`` is a data-preparation wrapper that pairs an operator
+with a flux field for sensitivity projection and flux reconstruction. It does
+not define :math:`\Pi`, apply native covariance, or own covariance transforms.
+By the time this API is called, :math:`H` already contains footprint times
+prior flux, so both the native and retained states are multiplicative
+scalings.
+
+Projection and centring
+-----------------------
+
+Let the native scaling state have mean :math:`m` and covariance :math:`B`:
+
+.. math::
+
+   \operatorname{E}[x] = m,
+   \qquad
+   \operatorname{Cov}(x) = B,
+   \qquad
+   \delta x = x - m.
+
+For a retained restriction :math:`\Pi`, define
+
+.. math::
+
+   \alpha = \Pi x,
+   \qquad
+   \delta\alpha = \alpha - \Pi m = \Pi\,\delta x,
+   \qquad
+   C_\alpha = \Pi B \Pi^\mathsf{T}.
+
+The covariance-natural prolongation associated with that restriction is
+
+.. math::
+
+   U_* = B \Pi^\mathsf{T} C_\alpha^{-1}.
+
+The full affine lift is therefore
+
+.. math::
+
+   \widehat{x}(\alpha)
+   = m + U_*\delta\alpha
+   = m + U_*(\alpha - \Pi m),
+
+not the uncentred expression :math:`U_*\alpha`. For a joint Gaussian model
+this lift is :math:`\operatorname{E}[x\mid\alpha]`. Without Gaussianity it is
+the linear-Bayes lift determined by the first two moments.
+
+In general, :math:`U_*`, :math:`U_{\mathrm{bucket}}`, and
+:math:`\Pi^\mathsf{T}` are different operators. The initial strategy preserves
+the established bucket-scaling interpretation by choosing
+
+.. math::
+
+   \Pi_U =
+   \left(U_{\mathrm{bucket}}^\mathsf{T} B^{-1}
+   U_{\mathrm{bucket}}\right)^{-1}
+   U_{\mathrm{bucket}}^\mathsf{T} B^{-1}.
+
+This gives :math:`\Pi_U U_{\mathrm{bucket}} = I` and
+:math:`U_* = U_{\mathrm{bucket}}`. OPE-17 returns :math:`C_\alpha`,
+:math:`H U_*`, :math:`H B \Pi^\mathsf{T}`, and either dense
+:math:`H B H^\mathsf{T}` or its diagonal.
+
+What OPE-17 does not construct
+------------------------------
+
+The exact Gaussian reduction also needs
+
+.. math::
+
+   B_\perp
+   = B - U_* C_\alpha U_*^\mathsf{T},
+
+and the centred conditional observation model
+
+.. math::
+
+   y \mid \alpha \sim \mathcal N\!\left(
+   Hm + H U_*(\alpha - \Pi m),
+   R + H B_\perp H^\mathsf{T}
+   \right).
+
+OPE-18 owns that solve-based transformation, unresolved covariance, and the
+single coherent artifact that binds prior, forward, residual, reconstruction,
+and state-treatment information. The OPE-17 product blocks are inputs to that
+work; they must not be described or persisted as though they were already the
+complete artifact. Arbitrary reporting-function products :math:`Q`,
+low-rank-plus-diagonal numerical views, and likelihood integration are also
+outside this API.
+
+Identity and serialization
+--------------------------
+
+``source_content_identity`` binds the shared :math:`B/H/\Pi/U_*` source
+content independently of the requested dense or diagonal observation product.
+``observation_covariance_view`` records that selector, and ``view_identity``
+is derived from it and the source identity. These values bind provenance and
+shared metadata; they are not load-time integrity checksums. Deserialization
+does not rehash product values to detect numeric tampering. The later
+coherent-reduction artifact can reference the OPE-17 source identity, but
+needs its own identity for the additional prior, affine-forward, residual,
+and reconstruction content.
+
+Memory and execution boundary
+-----------------------------
+
+The structured covariance action stores axis factors rather than dense native
+:math:`B`, but this API is otherwise eager. It materializes the native
+sensitivity, bucket prolongation, restriction/prolongation, reduced products,
+and requested observation product. For native size :math:`N`, retained size
+:math:`d`, and observation count :math:`M`, important dense storage includes
+:math:`M N` for :math:`H`, :math:`N d` for each native-by-retained array,
+:math:`d^2` for retained products, and either :math:`M^2` for dense
+:math:`H B H^\mathsf{T}` or :math:`M` for its diagonal.
+
+``observation_batch_size`` is an eager execution setting, not an xarray or
+Dask chunk size. Each batch applies :math:`B` to a group of columns from
+:math:`H^\mathsf{T}`, limiting the temporary :math:`N`-by-batch working set.
+Dense batches still fill a preallocated complete quadratic observation
+covariance. Changing the batch size changes execution only, not the scientific
+inputs or requested numerical form.
+
+Correspondence with ``verification-games``
+-------------------------------------------
+
+OPE-17 deliberately ports only the reusable lower-level prototype boundary.
+The recent 14-site production runs define retained coefficients with a
+source-blocked, absolute-prior-flux-weighted mean. For source :math:`s`, region
+:math:`r`, and native cell :math:`i`, let
+
+.. math::
+
+   q_{si} = \bar F^{\mathrm{UOB\,BASE}}_{si} A_i,
+   \qquad
+   \Pi_{(s,r),(s',i)} =
+   \mathbf 1[s=s']\,\mathbf 1[i\in r]\,
+   \frac{|q_{si}|}{\sum_{j\in r}|q_{sj}|}.
+
+Thus each row is nonnegative, sums to one, and averages native scaling factors
+within exactly one source and one active region. This physical restriction is
+chosen independently of :math:`B`; the covariance then determines
+:math:`C_\alpha` and :math:`U_*`. The prototype also supports a distinct
+signed-flux-total policy, but that is not the current 14-site production
+setting. Both policies require a supplied-:math:`\Pi`-first OGI strategy.
+
+.. list-table::
+   :header-rows: 1
+   :widths: 31 21 48
+
+   * - Prototype capability
+     - OPE-17 status
+     - Correspondence
+   * - Separable exponential covariance action, optional class blocking, and
+       batched :math:`\Pi B\Pi^\mathsf{T}`,
+       :math:`H B\Pi^\mathsf{T}`, and :math:`H B H^\mathsf{T}` products
+     - Faithful port
+     - Re-expressed as labelled xarray actions and product objects; OPE-17 also
+       supports ordered independent-source blocks.
+   * - Choose a physical restriction :math:`\Pi` first, then derive its
+       covariance-natural lift. Current 14-site production uses the
+       absolute-prior-flux-weighted regional mean above.
+     - Deliberate difference
+     - The first OPE-17 strategy is compatibility-oriented: it fixes
+       :math:`U_{\mathrm{bucket}}` first and derives :math:`\Pi_U`. The strategy
+       interface leaves room for a future :math:`\Pi`-first policy.
+   * - Cross-source covariance, arbitrary reporting functionals :math:`Q`,
+       :math:`B_\perp` products, coherent likelihood reduction, reconstruction,
+       and covariance approximation
+     - Not ported
+     - These remain prototype or downstream OPE-18 work and must not be inferred
+       from the presence of OPE-17 product blocks.
+
+API reference
+-------------
+
+See :mod:`openghg_inversions.native_covariance`,
+:mod:`openghg_inversions.source_covariance`, and
+:mod:`openghg_inversions.basis.covariance_products` for the public interfaces.
--- a/docs/usage/native_covariance.rst
+++ b/docs/usage/native_covariance.rst
@@ -23,9 +23,10 @@ The classes have the following distinct roles.
    * - ``BasisOperator``
      - Owns basis geometry and retained-state labels. For a single source,
        ``basis_matrix`` is the bucket prolongation :math:`U_{\mathrm{bucket}}`.
-       A gathered multisource matrix is a spatial template that the projection
-       code expands onto an explicit native source dimension. In both cases,
-       the transpose does not define the retained restriction :math:`\Pi`.
+       A gathered multisource matrix is a spatial template; the basis-side
+       ``native_prolongation`` adapter expands it onto a canonical explicit
+       native source dimension. In both cases, the transpose does not define
+       the retained restriction :math:`\Pi`.
    * - ``InvertibleNativeCovarianceAction``
      - Structural interface for applying :math:`B` and solving systems in
        :math:`B` without constructing a dense native covariance matrix.
@@ -54,9 +55,9 @@ The classes have the following distinct roles.
      - Validates and combines :math:`H`, :math:`B`, basis geometry, and a
        projection strategy.
    * - ``NativeCovarianceProducts``
-     - Frozen result and serialization dataclass containing the labelled
-       product blocks. Its contained xarray objects and provenance mapping
-       remain mutable.
+     - Frozen in-memory result dataclass containing the labelled product
+       blocks. Its contained xarray objects remain mutable. Durable identity,
+       schema design, and persistence are deferred to OPE-40.
 
 The data flow is:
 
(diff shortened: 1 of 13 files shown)


openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,1890 @@
...
+    """
+    expected_dims = set((*native_dims, observation_dim))
+    if set(sensitivity.dims) != expected_dims or len(sensitivity.dims) != len(expected_dims):
+        raise ValueError(
+            "native_sensitivity must have exactly one observation dimension and the native grid "
+            f"dimensions; got {sensitivity.dims!r}"
+        )
+    if observation_dim not in sensitivity.coords:
+        raise ValueError(f"native_sensitivity is missing observation coordinate {observation_dim!r}")
+    result = _densify(sensitivity).transpose(observation_dim, *native_dims)
```
brendan-m-murphy: The input sensitivity is normally Dask-backed in OpenGHG workflows, but `_validated_sensitivity()` calls `_densify()`, computing and converting the complete array to NumPy before `observation_batch_size` is applied. The subsequent loop therefore recreates blocked execution manually after discarding the existing Dask graph and chunks.

The batch size bounds only the native-by-block `B @ H_block.T` workspace; it does not bound full `H` materialization or the preallocated dense quadratic observation covariance. The method also calls `covariance.apply()` on one disposable row for validation, then applies the covariance again during the real calculation.

Please choose one explicit execution model. My preference, given the current eager covariance-action contract, is to make the pipeline the named materialization boundary, require/reject inputs accordingly here, and describe this as an eager numerical kernel with an `rhs_block_size` memory contract. Alternatively preserve Dask orchestration and compute actual observation chunks. In either case, remove the throwaway covariance application and add a Dask-backed regression test. I consider this merge-blocking architecture, not merely terminology or documentation.

brendan-m-murphy (commented):
One substantive execution-boundary concern.

openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,1890 @@
...
+            absent from the native sensitivity.
+    """
+    basis_grid_dims = tuple(basis_operator.meta.grid_dims)
+    basis_matrix = _densify(basis_operator.basis_matrix)
+    if native_dims == basis_grid_dims:
+        return basis_matrix
+
+    native_source_dim = native_dims[0]
+    basis_source_dim = getattr(basis_operator, "source_dim", None)
+    if not isinstance(basis_source_dim, str) or basis_source_dim not in basis_matrix.coords:
```
brendan-m-murphy: Yes. A `BasisOperator` method or basis-side adapter should own expansion of the gathered multisource template into canonical, full-native `U`, because that operation interprets the operator's ragged state/source representation.

The covariance-products layer should consume the resulting labelled `U`; it should not discover an undeclared `source_dim` convention with `getattr`. The basis boundary should own geometry and source expansion, but not densification, covariance solves, or execution policy.

openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,1890 @@
...
+        raise ValueError(
+            "A covariance with a leading native source dimension requires a gathered "
+            "multisource basis carrying source labels on its state coordinate"
+        )
+    state_sources = np.asarray(basis_matrix.coords[basis_source_dim].values)
```
brendan-m-murphy: This should be expressed as labelled xarray broadcasting/alignment at the basis boundary. The current NumPy source mask and manual `DataArray` reconstruction do more than add complexity: they drop native auxiliary coordinates such as 2-D grid metadata, source-axis metadata, and scalar context.

Please construct the full-native prolongation on the basis side, preserve coordinates whose dimensions are subsets of the native axes, and cover single-/multisource auxiliary-coordinate propagation in tests.

openghg_inversions/basis/covariance_products.py
brendan-m-murphy: Preferred resolution: materialize `H` explicitly at the upstream pipeline boundary, ideally together with related arrays, and make this kernel reject lazy input with a focused message. If the intended contract is instead lazy, preserve the Dask graph through real RHS blocks. The current middle ground—hidden full materialization followed by manual batching—should not remain.

openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,1890 @@
...
+                for value in source_values
+            ],
+            dtype=str,
+        )
+    else:
+        values = np.asarray(source_values)
+    return xr.DataArray(values, dims=column_dim, attrs=coordinate.attrs)
+
+
+def _json_label_default(value: object) -> str:
```
brendan-m-murphy: Yes, this overlaps the shared serialization-codec family, but the current variant is also lossy. When any row label is a tuple, it calls `list(value)` for every label: a heterogeneous tuple/scalar index raises `TypeError`, and strings in a mixed index become character lists. MultiIndex tuples are converted to JSON strings, so row and column axes lose equivalent typed selection/alignment.

Prefer a renamed typed column Index/MultiIndex and the existing MultiIndex serializer. If a codec is still required at the later artifact-I/O stage, extend the shared tagged codec rather than adding another permissive `str()` fallback.

openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,1890 @@
...
+            suffix += 1
+        result[column_name] = xr.DataArray(
+            np.asarray(coordinate.values),
+            dims=column_dim,
+            attrs=coordinate.attrs,
+        )
+    return result
+
+
+def _validate_symmetric(values: np.ndarray, name: str, *, tolerance: float = 1e-10) -> None:
```
brendan-m-murphy: These are generic square-matrix concerns, but extracting the helpers alone will not fix the design. The current path computes `C_alpha` symmetry/eigenvalue diagnostics in `_validate_positive_definite()` and immediately repeats them in `_with_matrix_diagnostics()`.

Please establish one repository-wide convention for labelled square matrices: distinct typed row/column dimensions with deterministic renamed level coordinates, one validation/diagnostic pass, and reusable construction/alignment helpers. Durable product packing can then be designed in the later artifact-I/O stage rather than growing a local namespace protocol here.

openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,1890 @@
...
+   imported rather than redefined here.
+3. :class:`RetainedProjectionStrategy` chooses a coherent ``(Pi, U_*)`` pair
+   and returns it as the :class:`RetainedProjection` value object.
+   :class:`PreserveBucketProlongation` is the initial concrete structural
+   implementation of that strategy protocol.
+4. :func:`project_native_covariance` combines those inputs into the frozen,
+   serializable :class:`NativeCovarianceProducts` dataclass. Its contained
+   xarray objects remain mutable.
+
+Projected artifacts carry a source identity derived from covariance
```
brendan-m-murphy: Revised conclusion: basis grouping, native covariance classes, and covariance source blocks are related inputs but mathematically distinct concepts.

- `basis_group` / `basis_partition` describe retained-state semantics.
- covariance `class_labels` define native zero-cross-covariance blocks.
- source covariance defines probabilistic blocks on the native source axis.

They may originate from the same scientific configuration, but this product kernel should not infer that they coincide. The pipeline/configuration layer should establish and validate any intended relationship once, while the numerical kernel consumes one canonical labelled native layout and requires exact alignment. The cohesion problem here is that this submodule currently reconciles all three label models itself.

openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,1890 @@
...
+            source/view identities, and basis-provenance metadata. Native
+            covariance configuration is intentionally excluded because it
+            occupies a separate child node in :meth:`to_datatree`.
+
+            The identities bind declared source and view metadata across the
+            artifact. They are not integrity checksums of the numerical output
+            variables, whose values are deliberately not re-hashed here.
+        """
+        configuration_digest = self._validated_configuration_digest()
+        arrays, coordinate_namespaces = _namespace_product_coordinates(
```
brendan-m-murphy: [blocking] Accepted dimension and auxiliary-coordinate names can collide with these fixed product variable names. I reproduced `state_dim="state_covariance"`: xarray omits/promotes that data variable and `from_dataset()` rejects the API's own output; an auxiliary coordinate with the same name can instead raise `MergeError`. If this serialization remains here, reserve every schema name or store products in independent DataTree nodes. Deferring durable product persistence to the later artifact-I/O stage would also avoid freezing this failure-prone layout now.

openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,1890 @@
...
+
+    Returns:
+        Frozen product dataclass whose arrays carry a shared source identity
+        and a view identity derived from the requested dense/diagonal
+        representation. These identities bind metadata and are not checksums
+        of computed output values. Its contained arrays remain mutable. Inputs
+        and products are eagerly materialized; dense observation covariance
+        has quadratic observation-space storage. Full PSD eigendiagnostics are
+        skipped above 512 rows to avoid adding cubic observation-space work.
+        Input attrs/units are not propagated: product attrs are replaced by
```
brendan-m-murphy: The API explicitly discards attrs/units, but #493 requires a units contract. For dimensionless scaling, `H U` and `H B Pi.T` inherit the sensitivity units and `H B H.T` has squared observation units. Please define and test that model rather than silently replacing unit metadata.

openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,1890 @@
...
+            "Basis grid dimensions must equal the covariance spatial dimensions; "
+            f"{basis_operator.meta.grid_dims!r} is incompatible with {native_dims!r}"
+        )
+    sensitivity = _validated_sensitivity(
+        native_sensitivity,
+        native_dims=native_dims,
+        observation_dim=observation_dim,
+        covariance=covariance,
+    )
+    batch_size = int(observation_batch_size)
```
brendan-m-murphy, line 273: `int(...)` silently accepts booleans and strings and truncates floats such as `1.9`, despite the public contract specifying an integer. Please require an integral non-Boolean value before checking positivity.

openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,1890 @@
...
+    """
+    restriction = projection.restriction
+    if not isinstance(projection.strategy, str) or not projection.strategy:
+        raise ValueError("Projection strategy must return a non-empty strategy identifier")
+    expected_restriction_dims = {state_dim, *native_dims}
+    if set(restriction.dims) != expected_restriction_dims or len(restriction.dims) != len(
+        expected_restriction_dims
+    ):
+        raise ValueError("Projection strategy restriction has invalid labelled dimensions")
+    restriction = _densify(restriction).transpose(state_dim, *native_dims)
```
brendan-m-murphy: This eagerly densifies every custom restriction. That is at odds with OPE-31's explicit bounded-memory sparse-Pi requirement for the 361 x 96,768 production shape. The projection boundary should preserve sparse/Dask structure and densify only explicit RHS blocks at the chosen execution boundary.

openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,1890 @@
...
+    missing = [label for label in state_sources if label not in set(native_sources.tolist())]
+    if missing:
+        raise ValueError(f"Basis state sources are absent from native sensitivity: {missing!r}")
+    base = basis_matrix.transpose(*basis_grid_dims, state_dim)
+    values = np.asarray(base.values)
+    source_mask = native_sources[:, np.newaxis] == state_sources[np.newaxis, :]
+    expanded = values[np.newaxis, ...] * source_mask.reshape(
+        (native_sources.size, *([1] * len(basis_grid_dims)), state_sources.size)
+    )
+    coords: dict[str, xr.DataArray] = {
```
brendan-m-murphy: This manual reconstruction keeps dimension and state-only coordinates but drops native auxiliary metadata such as a 2-D grid coordinate, source-axis metadata, or scalar context. A labelled xarray broadcast can construct the source mask while retaining coordinates whose dimensions are subsets of the native axes.

brendan-m-murphy (commented):
Follow-up after independent code-map, correctness, architecture/YAGNI, and scientific-xarray reviews.

I think this needs revision before merge, although the numerical kernel and dense-oracle tests look sound. The roadmap check changes the YAGNI conclusion: OPE-18/#566 consumes these product blocks, and OPE-31 is the concrete second projection strategy, so the in-memory artifact and strategy seam are needed. Durable product serialization/deserialization is also needed, but Brendan has clarified that it belongs to a later stage; I am updating Linear to reflect that boundary. PR #582's covariance-configuration serialization remains a separate, useful reproducibility boundary.

The existing review threads are all actionable:

- Basis grouping, native covariance classes, and source covariance blocks are distinct concepts. Their intended relationship should be established once upstream; this kernel should consume a canonical labelled native layout.
- Multisource expansion to full native U belongs on BasisOperator or a basis-side adapter. The covariance-products layer should not infer an undeclared source_dim convention.
- The source mask and coordinate reconstruction should use labelled xarray broadcasting/alignment.
- Full-H materialization before batching is an architectural blocker. Either materialize explicitly in the pipeline and make this a clearly eager kernel, or preserve lazy orchestration through real RHS blocks.
- The JSON label helper overlaps existing codecs and is lossy.
- Square-matrix axis handling needs one shared repository convention, with numerical diagnostics performed once.

A simpler boundary would be:

1. basis-side adapter → canonical full-native U;
2. numerical product kernel → B, canonical H, and one coherent projection;
3. later artifact-I/O stage → identities, provenance, schema, DataTree/NetCDF persistence.

For the bucket-preserving strategy, with Q = B^-1 U, G = U.T Q, and C = G^-1, we already know Pi = C Q.T, B Pi.T = U C, C_alpha = C, and H B Pi.T = (H U) C. Using those constructed identities avoids an extra covariance application, coherence recomputation, and repeated SPD diagnostics.

Additional concrete issues are inline.

openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,1890 @@
...
+    reinterpret them as a two-dimensional coordinate.
+
+    Args:
+        coordinate: One-dimensional row coordinate to copy.
+        column_dim: Destination column dimension.
+
+    Returns:
+        A plain one-dimensional column coordinate preserving attributes.
+    """
+    source_values = list(coordinate.values)
```
brendan-m-murphy: This encoding is not required by xarray and is not type-safe. If any label is a tuple, the loop calls `list(value)` for every label: mixed tuple/scalar indexes raise `TypeError`, while mixed strings are converted to character lists. MultiIndex tuples also become JSON strings, so row and column axes no longer have equivalent typed selection/alignment. A renamed typed column MultiIndex works with the existing MultiIndex serializer.

openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,1890 @@
...
+        leading_dims=native_dims,
+    )
+    b_restriction_transpose = covariance.apply(restriction_transpose)
+
+    state_covariance = xr.dot(
+        projection.restriction,
+        b_restriction_transpose,
+        dim=list(native_dims),
+    ).transpose(state_dim, state_column_dim)
+    _validate_positive_definite(np.asarray(state_covariance.values), "C_alpha")
```
brendan-m-murphy: `_validate_positive_definite()` computes the full eigenspectrum here, and `_with_matrix_diagnostics()` immediately repeats the symmetry/eigenvalue calculation. Please validate/diagnose this matrix once and reuse the result.

openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,578 @@
...
+        )
+        state_covariance = with_square_matrix_diagnostics(
+            state_covariance.rename("state_covariance").assign_attrs(units="1"),
+            mathematical_name="C_alpha",
+            require_positive_definite=True,
+        )
+    else:
+        state_covariance = projection.state_covariance.rename("state_covariance")
+
+    if projection.state_covariance is None or projection.covariance_restriction_transpose is None:
```
brendan-m-murphy: [P1] Supplying both optional cached products skips this check entirely, and neither cache is independently established from Pi and B. I reproduced a structural strategy returning valid Pi/U plus zero-filled state_covariance and B Pi.T: the kernel accepted zero C_alpha and zero H B Pi.T, even retaining stale positive-definite attrs. Supplying only a consistently scaled B Pi.T can also pass the current coherence equation because the derived C_alpha receives the same scale. Please keep these optimized values private to the built-in strategy, or independently validate B Pi.T = B @ Pi.T and C_alpha = Pi @ B Pi.T, including labels, finiteness, laziness, symmetry, and positive definiteness.

openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,578 @@
...
+        native_dims=native_dims,
+        state_dim=state_dim,
+    )
+    projection = _validated_projection(
+        projection,
+        native_reference=sensitivity,
+        native_dims=native_dims,
+        state_dim=state_dim,
+    )
+    state_column_dim = matrix_column_dim(state_dim, projection.restriction.dims)
```
brendan-m-murphy: [P2] The generated state-column name only avoids projection dimensions, not the observation dimension or other output coordinates. With state_dim='state' and the otherwise valid observation_dim='state_cov', this also chooses 'state_cov'; xr.dot then treats two logical axes as one and the cross-covariance transpose fails with `ValueError: axes don't match array`. Auxiliary coordinates named state_cov can likewise be displaced. Choose matrix column names against the union of all relevant input/output dimensions and coordinate names, and add collision regressions.

openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,578 @@
...
+            },
+            attrs=unit_attrs,
+            name="native_observation_covariance",
+        )
+        return with_square_matrix_diagnostics(
+            result,
+            mathematical_name="H B H.T",
+            maximum_eigen_diagnostic_size=MAX_DENSE_EIGEN_DIAGNOSTIC_SIZE,
+        )
+    result = xr.DataArray(
```
brendan-m-murphy, line 534: [P2] The diagonal path returns without the finite-value validation used by `with_square_matrix_diagnostics`. With finite H=1e308 and identity covariance I get `[inf]`, `minimum_diagonal=inf`, and `diagonal_nonnegative=True`; the equivalent dense request correctly raises. Reject non-finite and materially negative diagonal results before attaching diagnostics.

openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,578 @@
...
+    sensitivity = _validated_sensitivity(
+        native_sensitivity,
+        native_dims=native_dims,
+        observation_dim=observation_dim,
+    )
+    prolongation = _validated_prolongation(basis_prolongation, native_dims=native_dims, state_dim=state_dim)
+    _validate_exact_native_coordinates(
+        prolongation, sensitivity, native_dims=native_dims, role="prolongation"
+    )
+    projection = (strategy or PreserveBucketProlongation()).projection(
```
brendan-m-murphy: [P2] Boolean fallback changes scientific policy for a valid false-valued structural strategy. I reproduced a strategy implementing `__bool__` as false whose `projection()` was never called; the default strategy was silently used. Please branch explicitly on `strategy is None`.

openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,578 @@
...
+    restriction: xr.DataArray,
+    *,
+    native_dims: tuple[str, ...],
+    state_dim: str,
+    column_dim: str,
+    rhs_block_size: int,
+) -> xr.DataArray:
+    """Apply covariance to explicit dense retained-state RHS blocks."""
+    blocks: list[xr.DataArray] = []
+    for start in range(0, restriction.sizes[state_dim], rhs_block_size):
```
brendan-m-murphy, line 379: [P2] Slicing before `.compute()` does not by itself bound a Dask restriction when its producer has one chunk spanning the full state axis. With batch size 1 and a two-state delayed single chunk, the full producer executed four times across these covariance blocks and the later Pi-B-Pi blocks. Either require/rechunk the state axis to the promised block size, share a named materialization boundary, or narrow the documentation so it does not claim that this setting bounds custom-restriction materialization.

openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,578 @@
...
+    if set(restriction.dims) != expected or len(restriction.dims) != len(expected):
+        raise ValueError("Projection strategy restriction has invalid labelled dimensions")
+    prolongation = _validated_prolongation(
+        projection.prolongation, native_dims=native_dims, state_dim=state_dim
+    ).assign_attrs({**projection.prolongation.attrs, "units": "1"})
+    for role, array in (("restriction", restriction), ("prolongation", prolongation)):
+        _validate_exact_native_coordinates(array, native_reference, native_dims=native_dims, role=role)
+    if state_dim not in restriction.coords or state_dim not in prolongation.coords:
+        raise ValueError("Projection strategy restriction and prolongation require state labels")
+    if not restriction.get_index(state_dim).equals(prolongation.get_index(state_dim)):
```
brendan-m-murphy: [P2] Matching state indexes are insufficient when state auxiliary coordinates carry semantic grouping. A custom strategy can return restriction and prolongation with the same index but conflicting `group(state)` values; restriction/C_alpha then describe one grouping while prolongation/H_alpha describe another. Require compatible state auxiliary coordinates (and native auxiliaries where retained), or canonicalize them from one authoritative projection primitive.

openghg_inversions/basis/operators.py
```diff
@@ -769,6 +854,65 @@ def source_labels(self) -> tuple[str, ...]:
...
+        Raises:
+            ValueError: If dimensions, source labels, or native indexes are
+                incompatible with this multisource basis.
+        """
+        if len(native_dims) != len(self.meta.grid_dims) + 1 or native_dims[1:] != self.meta.grid_dims:
+            raise ValueError(
+                "A multisource basis requires one native source dimension followed by its "
+                f"grid dimensions; got {native_dims!r}"
+            )
+        native_source_dim = native_dims[0]
```
brendan-m-murphy, line 884: [P2] A configured native source dimension may equal the retained MultiIndex level name. `IndependentSourceCovariance(..., source_dim='source')` and `MultiSourceBucketBasisOperator(..., source_dim='source')` are both accepted, but `native_prolongation` later fails with an xarray AlignmentError because `source` identifies both a native Index and a state-level Index. Reject this collision explicitly with a focused ValueError, or namespace the retained level before broadcasting.

openghg_inversions/_labelled_matrices.py
```diff
@@ -0,0 +1,176 @@
...
+                suffix += 1
+            renamed_levels.append(candidate)
+            reserved.add(candidate)
+            if raw_name is not None:
+                index_level_names.add(str(raw_name))
+        renamed = index.rename(renamed_levels)
+        coordinates = xr.Coordinates.from_pandas_multiindex(renamed, column_dim)
+        result.update({str(name): coordinate for name, coordinate in coordinates.items()})
+    else:
+        result[column_dim] = xr.DataArray(
```
brendan-m-murphy, line 71: [P2] Converting every non-MultiIndex through `to_numpy()` discards specialized pandas index semantics. I reproduced CategoricalIndex becoming a generic Index on covariance axes; PeriodIndex and RangeIndex have the same issue. Preserve/rename the pandas Index object itself and add categorical/period regressions. MultiIndex column reconstruction should also copy row and level-coordinate attrs.

brendan-m-murphy (commented):
Re-review of the rewritten implementation. The earlier architecture comments are addressed and I have resolved those threads: the module is now an in-memory eager kernel, basis operators own canonical native prolongation, persistence is deferred to OPE-40, and the labelled-array/units/batching fixes are in place.

There is still one merge-blocking correctness issue: public strategy-supplied cached products can bypass the scientific invariants and silently return incorrect C_alpha and H B Pi.T. The other inline findings concern output-axis collisions, diagonal overflow, strategy dispatch, custom-Pi chunk guarantees, custom metadata consistency, and source/index naming behavior.

I would revise before merge, primarily by making one projection primitive authoritative and deriving covariance products inside the kernel. Optimized cached products should either be private to the built-in strategy or validated independently against B @ Pi.T and Pi @ B Pi.T.

Verification: all current CI jobs pass. Focused local suites passed (51 covariance/source tests; 67 covariance/basis tests; an independent 83-test basis review), along with Ruff, Pyright, and git diff --check. The earlier hidden-H computation, throwaway covariance action, lossy heterogeneous labels, native auxiliary-coordinate loss, units omission, duplicated eigendiagnostics, and premature product persistence are materially fixed.

openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,782 @@
...
+@dataclass(frozen=True, slots=True)
+class RetainedProjection:
+    """A labelled restriction/prolongation pair selected by one strategy.
+
+    Attributes:
+        restriction: ``Pi`` with dimensions ``(state_dim, *native_dims)``.
+        prolongation: Covariance-natural ``U_*`` with dimensions
+            ``(*native_dims, state_dim)``.
+        strategy: Stable identifier for the scientific projection choice.
+        state_covariance: Optional cached ``C_alpha``. The kernel independently
```
brendan-m-murphy: I would remove these optional public caches rather than strengthen their adversarial validation. External caches are now always recomputed and discarded, while only the exact built-in strategy/private subtype gets the fast path. A simpler contract is for the strategy to declare one authoritative primitive (fixed U or fixed Pi) and for this kernel to derive every covariance product. That removes the cache comparison helpers, exact-type trust channel, duplicated OPE-31 work, and much of the validation that made this module difficult to reason about.

openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,782 @@
...
+        )
+        return with_square_matrix_diagnostics(
+            result,
+            mathematical_name="H B H.T",
+            maximum_eigen_diagnostic_size=MAX_DENSE_EIGEN_DIAGNOSTIC_SIZE,
+        )
+    if not np.all(np.isfinite(values)):
+        raise ValueError("diag(H B H.T) must contain only finite real values")
+    diagonal_scale = max(1.0, float(np.max(np.abs(values))) if values.size else 1.0)
+    diagonal_tolerance = 1e-10 * diagonal_scale
```
brendan-m-murphy: This remaining check can silently accept a clearly negative variance when another observation has a much larger scale. For example, `[1e20, -1]` produces a global tolerance of `1e10`, so `-1` is accepted and marked nonnegative. Please use a per-observation numerical error scale, or a defensible fixed absolute tolerance independent of unrelated diagonal entries. This is an output scientific invariant rather than adversarial-input validation.

brendan-m-murphy (commented):
Final design suggestion after the follow-up fixes: I think this should use a simpler trust model, not accumulate more adversarial validation. Validate the labelled public boundary once, derive internal products centrally, and translate failures from xarray/covariance/numerical operations into helpful domain errors.

Concretely:

- let a projection strategy provide one authoritative primitive—fixed U or fixed Pi;
- let the kernel derive B Pi.T, C_alpha, and the compatible lift, owning all covariance actions and batching;
- remove the optional public cached products and the private exact-type trusted fast path;
- retain dimension/index checks where xarray could otherwise align or contract silently;
- do not compare arbitrary coordinate attrs unless the calculation actually uses them;
- catch natural AlignmentError/TypeError/numerical failures and re-raise concise messages identifying the invalid sensitivity, prolongation, restriction, or covariance;
- test realistic malformed inputs and those messages rather than forged caches or hostile protocol objects.

This would shrink the validation surface while preserving the important scientific safeguards: exact labelled axes, finite real numerical inputs/results, positive-definite retained covariance, and coherent projection algebra. The eight findings from my preceding review are implemented and their threads are now resolved.

One concrete numerical issue remains inline: negative diagonal tolerance should not be relaxed by an unrelated large variance elsewhere in the vector. Aside from that and the strategy simplification, the numerical paths and revised boundaries look sound. Current CI passes; focused local tests, Ruff, Pyright, and diff checks also pass.

openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,671 @@
...
+            covariance.solve(prolongation),
+            row_dim=state_dim,
+            column_dim=state_column_dim,
+            leading_dims=native_dims,
+        )
+        gram = xr.dot(prolongation, precision_prolongation, dim=list(native_dims)).transpose(
+            state_dim, state_column_dim
+        )
+        gram_values = np.asarray(gram.values, dtype=np.float64)
+        _require_symmetric(gram_values, "U.T B^-1 U")
```
brendan-m-murphy, line 136: Because covariance is typed as an InvertibleNativeCovarianceAction, the missing piece is the semantic contract on that protocol: it should guarantee a self-adjoint positive-definite operator and a consistent inverse. Then this kernel can trust symmetry and positivity by construction. Retain _require_symmetric only if it is useful as a cheap diagnostic that turns a violated action contract into a helpful error; it should not be what certifies B.

openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,671 @@
...
+        )
+
+
+@dataclass(frozen=True, slots=True)
+class NativeCovarianceProducts:
+    """Labelled in-memory product blocks induced by one coherent projection."""
+
+    restriction: xr.DataArray
+    prolongation: xr.DataArray
+    state_covariance: xr.DataArray
```
brendan-m-murphy, line 218: After checking xarray extension behavior, I do not think a DataArray subclass or accessor is a sound certificate: subclass propagation is inconsistent, accessors exist on every DataArray, and mutable payloads can invalidate cached state. For this PR, strengthen the covariance-action contract and derive covariance properties algebraically. If OPE-18 later needs a reusable certified dense SPD value, prefer composition around owned data and its Cholesky factor.

openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,671 @@
...
+
+    Attributes:
+        restriction: ``Pi`` with dimensions ``(state_dim, *native_dims)``.
+        prolongation: Covariance-natural ``U_*`` with dimensions
+            ``(*native_dims, state_dim)``.
+        strategy: Stable identifier for the scientific projection choice.
+    """
+
+    restriction: xr.DataArray
+    prolongation: xr.DataArray
```
brendan-m-murphy: Removing the optional caches fixes the trust bypass, but the strategy still returns two redundant authorities: Pi and U_*. Please make Pi the authoritative result. The kernel can derive B Pi.T, C_alpha = Pi B Pi.T, and U_* = B Pi.T C_alpha^-1. The bucket strategy derives Pi from the supplied bucket U, while OPE-31 can supply Pi directly. This removes the coherence check and repeated prolongation validation/materialization.

openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,671 @@
...
+            terms = rhs * b_rhs
+            values[start:stop] = np.asarray(terms.sum(dim=list(native_dims)).values)
+            assert diagonal_error_bounds is not None
+            contraction_magnitudes = np.asarray(
+                abs(terms).sum(dim=list(native_dims)).values,
+                dtype=np.float64,
+            )
+            diagonal_error_bounds[start:stop] = contraction_error_factor * np.maximum(
+                contraction_magnitudes,
+                1.0,
```
brendan-m-murphy: The per-observation scale fixes the earlier cross-observation bug, but flooring it at 1.0 makes the tolerance depend on units. A contraction with local magnitude 4e-16 and value -4e-16 is accepted because the bound is inflated to roughly 7e-15; rescaling H can therefore change rejection into silent clipping. Please derive the bound only from the local contraction magnitude and add a small-scale/scale-invariance regression test.

openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,671 @@
...
+        rhs = to_column_axis(
+            sensitivity.isel({observation_dim: slice(start, stop)}),
+            row_dim=observation_dim,
+            column_dim=column_dim,
+            leading_dims=native_dims,
+        )
+        b_rhs = covariance.apply(rhs)
+        if output == "dense":
+            block = xr.dot(sensitivity, b_rhs, dim=list(native_dims)).transpose(observation_dim, column_dim)
+            values[:, start:stop] = np.asarray(block.values)
```
brendan-m-murphy, line 490: Please validate the covariance result as finite and real before assigning it into values. The destination dtype is based only on sensitivity.dtype, so a structural covariance action returning complex values currently loses its imaginary component here and the resulting real matrix is accepted. This should fail with a helpful covariance-action contract error. The diagonal assignment has the same issue.

openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,671 @@
...
+import xarray as xr
+from xarray.namedarray.pycompat import is_chunked_array
+
+from openghg_inversions._labelled_matrices import (
+    matrix_column_dim,
+    renamed_column_coordinates,
+    to_column_axis,
+    with_square_matrix_diagnostics,
+)
+from openghg_inversions.native_covariance import InvertibleNativeCovarianceAction
```
brendan-m-murphy, line 38: Please make the mathematical action contract explicit here and on NativeCovarianceAction: the base action represents a self-adjoint positive-semidefinite operator on the labelled native space; the invertible action represents a positive-definite operator, and solve is the compatible inverse of apply. With that semantic guarantee, downstream kernels can trust covariance symmetry/positivity by construction and reserve runtime checks for helpful numerical failures.

openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,671 @@
...
+        """Construct the prior-precision-compatible restriction.
+
+        Args:
+            covariance: Invertible labelled action for native covariance ``B``.
+            basis_prolongation: Eager canonical bucket prolongation ``U``.
+            native_dims: Ordered native dimensions.
+            state_dim: Retained-state dimension.
+
+        Returns:
+            The coherent ``(Pi_U, U)`` pair and algebraically known products.
```
brendan-m-murphy: This is stale after removing the cached products: RetainedProjection now contains only Pi, U, and the strategy identifier. Please describe only what is returned; if the projection contract becomes Pi-first, this wording should follow that narrower contract.

openghg_inversions/basis/covariance_products.py
```diff
@@ -0,0 +1,671 @@
...
+        return RetainedProjection(
+            restriction=restriction,
+            prolongation=prolongation,
+            strategy=self.name,
+        )
+
+
+@dataclass(frozen=True, slots=True)
+class NativeCovarianceProducts:
+    """Labelled in-memory product blocks induced by one coherent projection."""
```
brendan-m-murphy: Please document the public field contract here: dimensions and units, C_alpha being numerically PD, dense HBH.T being PSD but potentially singular, and the diagonal view being a nonnegative vector. Also note that the frozen dataclass does not make its contained DataArrays immutable, so diagnostic attrs are snapshots rather than durable certificates.

docs/usage/native_covariance.rst
```diff
@@ -0,0 +1,261 @@
...
+       covariance-natural prolongation :math:`U_*`.
+   * - ``PreserveBucketProlongation``
+     - Current structural implementation of the strategy interface. It fixes
+       :math:`U_* = U_{\mathrm{bucket}}` and derives the compatible
+       :math:`\Pi_U`.
+   * - ``RetainedProjection``
+     - Frozen value dataclass returned by a strategy. It carries
+       :math:`(\Pi, U_*)` and the strategy identifier; it is not a protocol,
+       and its contained xarray objects remain mutable.
+   * - ``project_native_covariance``
```
brendan-m-murphy, line 58: This should say that project_native_covariance trusts the covariance action contract rather than validating B globally. It validates the labelled inputs and constructed products available at this boundary; a matrix-free action cannot generally be certified as SPD by probing these particular right-hand sides.

brendan-m-murphy (commented):
This round is materially simpler: the public derived caches and privileged built-in path are gone, and the diagonal check now uses observation-local information. I still think two correctness issues and one ownership issue should be addressed before merge.

First, the diagonal tolerance is not scale invariant because its local magnitude is floored at 1.0; this can silently accept a relatively large negative covariance after changing units. Second, a complex result returned by a covariance action is assigned into real storage before validation, so its imaginary part can disappear. Both should produce helpful domain errors rather than accumulating more adversarial validation.

The remaining architectural issue is that a strategy still supplies both Pi and U_*. Those are redundant authorities, which forces the kernel to validate their coherence and repeat prolongation validation. Prefer a Pi-first contract: a strategy returns Pi, the kernel forms B Pi.T and C_alpha, and then derives U_* = B Pi.T C_alpha^-1. The bucket-preserving strategy derives Pi from the supplied bucket U; the forthcoming OPE-31 strategy can supply Pi directly.

For covariance validity, please put the guarantee at the covariance-action boundary. NativeCovarianceAction should semantically guarantee a self-adjoint PSD operator; InvertibleNativeCovarianceAction should guarantee SPD behavior and a consistent solve. The numerical kernel can then trust covariance properties by construction and retain only cheap checks that turn natural numerical failures into useful messages. Dense HBH.T is PSD but may be singular; C_alpha is PD. Diagnostic attrs and eigendecompositions are observations, not certificates.

I do not recommend a DataArray subclass or accessor as a certificate. Xarray does not preserve subclasses consistently, every DataArray can expose an accessor, and mutable payloads can stale either mechanism. If OPE-18 later needs a reusable certified dense SPD object, use composition around owned data and its reusable factorization.

Focused verification remains green: 81 covariance/source/basis tests passed locally, Ruff and Pyright passed, and the current GitHub checks are green.

PR #589, opened 2026-08-14
Add explicit legacy fixedbasis compatibility route

### LINKED ISSUE #588: [P0 quick fix] Add an explicit run_hbmcmc.py route to fixedbasisMCMC for legacy outputs
Parent: #587

## Priority

**P0 quick fix — unblock frustrated users while the main `run_rhime` workflow is reworked.**

## User need

Some production users still need the genuine legacy HBMCMC products written by `inferpymc_postprocessouts(...)` and currently invoke `run_hbmcmc.py` from existing INI/SLURM workflows.

The script now translates supported settings into the modern `run_rhime` path. Direct `fixedbasisMCMC(...)` remains suitable for the compatibility run itself, but its deprecated `"hbmcmc"` / `"hbmcmc_postprocessing"` names are currently canonicalized to the modern `make_legacy_hbmcmc_output(...)` adapter. That is not the formatter these users require.

Provide an explicit temporary route through the existing script to the current fixedbasis sampler and the true legacy postprocessor.

## Proposed interface

Add:

```text
python .../run_hbmcmc.py -c config.ini --legacy-fixedbasis
```

The exact flag spelling may change during implementation if argparse conventions require it, but selection must be explicit.

Do **not** change the default. Without the flag, `run_hbmcmc.py` must continue translating to and calling `run_rhime(...)`.

Do **not** repurpose `mcmc_type = "fixed_basis"`; existing configurations already use that value on the modern route.

## Required behavior

With `--legacy-fixedbasis`:

- select `openghg_inversions.hbmcmc.hbmcmc.fixedbasisMCMC` before modern RHIME parameter translation;
- parse the legacy INI vocabulary and retain CLI start/end overrides;
- preserve legacy keys and meanings including `outputpath`, `outputname`, `nit`, `nchain`, `emissions_name`, prior dictionaries, and sampler kwargs;
- for an absent output setting or legacy `"hbmcmc"` / `"hbmcmc_postprocessing"` request, select a genuine legacy-postprocessing mode;
- ensure that mode calls `openghg_inversions.hbmcmc.inversion_pymc.inferpymc_postprocessouts(...)` exactly once;
- do not canonicalize that mode to `output_format="legacy"` or route it through `make_legacy_hbmcmc_output(...)`;
- preserve the historical filename and NetCDF product produced by `inferpymc_postprocessouts(...)`;
- copy the source configuration for provenance as the script does today;
- print a prominent notice that the direct compatibility implementation and legacy formatter were selected;
- propagate errors without silently falling back to `run_rhime`.

The implementation may retain `output_format="legacy"` as the modern compatibility formatter. It must distinguish that mode from the true `inferpymc_postprocessouts` mode.

## Acceptance criteria

- [ ] Parser test: `--legacy-fixedbasis` is explicit and defaults false.
- [ ] Routing test: the flag calls `fixedbasisMCMC` and does not call `run_rhime`.
- [ ] Default-routing test: no flag continues to call `run_rhime`.
- [ ] Argument test: legacy names and CLI date overrides reach `fixedbasisMCMC` unchanged.
- [ ] Output-resolution test: absent, `"hbmcmc"`, and `"hbmcmc_postprocessing"` compatibility requests select the true legacy postprocessor on the opt-in route.
- [ ] Formatter test: `inferpymc_postprocessouts(...)` is called exactly once and `make_legacy_hbmcmc_output(...)` is not called.
- [ ] Product test: returned/saved variables, dimensions, coordinates, values, relevant attributes, units, and historical filename match the direct `inferpymc_postprocessouts(...)` path.
- [ ] Validation/side-effect test: config copying and country-file checks retain their intended order.
- [ ] Failure test: unsupported options fail with a targeted message rather than falling back.
- [ ] Documentation update for INI/SLURM users, clearly marking this as a temporary escape hatch.

## Non-goals

- Making `fixedbasisMCMC` the default or the future architecture.
- Renaming `run_rhime`.
- Reimplementing the release-0.6 sampler.
- Restoring arbitrary unchecked passthrough globally.
- Refactoring the main RHIME pipeline in this quick-fix PR.
- Replacing or removing the modern `output_format="legacy"` adapter.

## Removal gate

Do not remove this escape hatch merely because old configs run through modern RHIME. Removal requires the usability and output gates in #587 plus an explicit decision about users who still require `inferpymc_postprocessouts` products.

Linear task: [OPE-42](https://linear.app/openghg-inversions/issue/OPE-42/p0-quick-fix-route-run-hbmcmcpy-to-fixedbasismcmc-and-inferpymc).

brendan-m-murphy:
## Summary

- add an explicit `--legacy-fixedbasis` escape hatch to `run_hbmcmc.py` while keeping `run_rhime` as the default
- preserve untranslated fixedbasis configuration vocabulary and route legacy output requests to `inferpymc_postprocessouts(...)` exactly once
- keep `output_format="legacy"` on the modern compatibility adapter, validate unsupported options and country files before side effects, and preserve config provenance
- document the temporary INI/SLURM compatibility workflow

## Why

Production users still need the historical HBMCMC NetCDF product and filename emitted by `inferpymc_postprocessouts(...)`. The existing compatibility aliases are intentionally canonicalized to the modern `make_legacy_hbmcmc_output(...)` adapter, so an explicit and isolated route is needed.

## Impact

Existing `run_hbmcmc.py` invocations are unchanged. Users who explicitly pass `--legacy-fixedbasis` get the current fixedbasis sampler plus the genuine historical postprocessor, with errors propagated and no fallback to RHIME.

## Validation

- `HOME=/tmp MPLCONFIGDIR=/tmp .venv/bin/pytest -q tests/test_run_hbmcmc_shim.py tests/test_postprocessing.py` (82 passed)
- `.venv/bin/ruff check` on changed Python/test files
- `.venv/bin/ruff format --check` on changed Python/test files
- `git diff --check`

Closes #588

Linear: OPE-42

### DIFF
--- a/docs/usage/getting_started.rst
+++ b/docs/usage/getting_started.rst
@@ -139,6 +139,12 @@ Method 2: ini file
   ``openghg-inversions run-rhime`` and the RHIME config vocabulary.
   Unsupported fixedbasis-only options now raise targeted errors instead of
   being passed through to ``inferpymc``.
+- Existing INI workflows that temporarily require the historical
+  ``fixedbasisMCMC`` / ``inferpymc_postprocessouts`` product can add
+  ``--legacy-fixedbasis``. This explicit opt-in keeps legacy parameter names;
+  it does not translate them or fall back to RHIME. A missing output format or
+  ``hbmcmc`` / ``hbmcmc_postprocessing`` selects the historical product, while
+  ``output_format="legacy"`` continues to select the modern adapter.
 - A sample ``.ini`` script is at the bottom of this document.
 
 Method 3: as a job on Blue Pebble
@@ -148,6 +154,10 @@ Method 3: as a job on Blue Pebble
   SLURM <https://www.acrc.bris.ac.uk/protected/hpc-docs/job_types/serial.html>`__
   using ``sbatch``.
 - This file will specify:
+
+  - the same ``run_hbmcmc.py`` command as above; add
+    ``--legacy-fixedbasis`` after the script path only when the submitted job
+    must reproduce the historical fixedbasis product;
 
   - the name of the job
   - resources required (number of nodes, number of tasks per node,
--- a/docs/usage/rhime.rst
+++ b/docs/usage/rhime.rst
@@ -847,8 +847,13 @@ compatibility output formats, but new workflows should save and load DataTree
 files. It translates legacy option names to the modern ``run_rhime`` API and
 uses the legacy filename convention. New scripts and new configs should use
 ``openghg-inversions run-rhime`` or ``run_rhime(...)`` directly.
-This compatibility route no longer preserves the exact historical
-``fixedbasisMCMC`` / ``inferpymc`` passthrough behaviour. Use release ``0.6`` or
-earlier if you need the old fixedbasis implementation.
+For temporary reproduction of historical products, pass
+``--legacy-fixedbasis`` to ``run_hbmcmc.py``. This explicit opt-in sends the
+untranslated INI parameters to ``fixedbasisMCMC``; a missing output format or
+the old ``hbmcmc`` / ``hbmcmc_postprocessing`` names select the historical
+``inferpymc_postprocessouts`` product and filename. An explicit
+``output_format="legacy"`` still selects the modern legacy-format adapter.
+The command prints a prominent warning, raises on unsupported options, and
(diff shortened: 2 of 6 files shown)


pearson01 (approved):
Thanks for doing this Brendan, all looks good to me!

PR #591, opened 2026-08-15
[OPE-45] Expose the visible run_rhime orchestration spine

brendan-m-murphy:
## Summary

- flatten the hidden RHIME runner flow into plainly named, supported resolve, prepare, align, materialize, build, sample, and result stages
- keep the standard, multisector, and prepared-input recipes independently readable while reusing the public build/sample/result mechanics
- replace preparation-signature reflection with explicit option ownership and normalize config options once at the public workflow boundary
- preserve canonical xarray/Dask inputs across the explicit PyMC materialization boundary and retain complete-model-builder replay behavior

Implements [OPE-45](https://linear.app/openghg-inversions/issue/OPE-45/p0-w3a-expose-the-visible-run-rhime-orchestration-spine).

## Validation

- `tox -e py310-openghgCur -- -m rhime_contract` — 40 passed
- `tox -e py310-openghgCur -- tests/test_rhime.py tests/test_run_hbmcmc_shim.py` — 217 passed
- `tox -p --parallel-no-spinner` — py310-openghgCur, borrowed-types, and lint passed


openghg_inversions/rhime/runner.py
```diff
@@ -261,87 +353,240 @@ def _validate_multisector_basis_layout(
...
+        prepared: Canonical RHIME inputs borrowed from preparation or replay.
+
+    Returns:
+        A dataset copy whose available model input variables are dense and
+        eager, with all other variables shared from ``prepared``.
+    """
+    timing_start = timer_start()
+    inv_inputs = prepared.inv_inputs
+    names = [name for name in _MODEL_INPUT_VARIABLES if name in inv_inputs]
+    dense_inputs = xr.Dataset({name: to_dense(inv_inputs[name]) for name in names}).compute()
```
brendan-m-murphy: **High — this materializes inputs the selected model may not use.** `_MODEL_INPUT_VARIABLES` includes every aggregation-error representation, so an explicit `aggregation_error_mode="none"` or `"low_rank"` can still compute and densify a dormant `aggregation_error_covariance` merely because it is present. That covariance is O(nmeasure²), making this a potentially large memory/work regression for replay or externally prepared inputs. Please make the selection mode-aware and add a Dask regression with selected plus dormant representations.

Smaller related point: constructing and computing this temporary Dataset also computes attached lazy auxiliary coordinates, but the shallow copy below retains the original lazy coordinates, so those results are discarded and may be computed again during model coordinate registration. That is worth fixing or explicitly avoiding, but it is not as serious as materializing the unused covariance.

openghg_inversions/rhime/runner.py
```diff
@@ -561,8 +945,38 @@ def run_rhime(
...
+        if config_file is not None
+        else dict(kwargs)
+    )
+    setup = resolve_rhime_options(params=params, multisector=False)
+
+    # Preparation visibly follows the scientific order implemented by the
+    # supported coarse stage: retrieve/reload merged data; filter observations
+    # while retaining aligned site metadata; load/fit the basis; construct
+    # sensitivities; and assemble labelled inversion inputs.
+    prepared = prepare_rhime_run_inputs(setup.data_args, multisector=False)
```
brendan-m-murphy: **High — the central OPE-45 outcome is still hidden here.** The executable runner shows only `resolve -> prepare -> align -> materialize -> build -> sample -> result`; retrieval/reload, filtering/site alignment, basis construction, sensitivity construction, and labelled assembly exist only in the preceding comment and inside `prepare_rhime_inputs`. `prepare_rhime_run_inputs` is itself a timed forwarding wrapper around that large preparer. A maintainer opening `run_rhime` therefore still cannot follow the required scientific stages as calls in execution order. Please expose/call the minimum ordinary named preparation stages here (and in the multisector runner) and assert their real ordering, without introducing a new coordinator or framework.

tests/test_rhime.py
```diff
@@ -2043,49 +2099,154 @@ def fail_setup(**kwargs: Any) -> None:
...
+        "with_prepared_rhime_sites",
+        "materialize_rhime_model_inputs",
+        "build_standard_rhime_model",
+        "build_multisector_rhime_model",
+        "sample_rhime_model",
+        "make_standard_rhime_result",
+        "make_multisector_rhime_result",
+    )
+
+    for name in stage_names:
```
brendan-m-murphy, line 2474: **Medium — this proves export identity, not W2.0 composability.** The ordering test replaces every stage with fakes, while this test checks only `getattr`/`__all__`. There is no executable contract that imports the supported stages from `openghg_inversions.rhime` and passes their real handoffs through a complete copied spine without private glue or manually assembled role/output manifests. Please add one deterministic external-style composition test. This need not implement the later OPE-44 example or documentation.

brendan-m-murphy (commented):
**Review verdict: request changes.** GitHub does not allow the connected account to formally request changes on its own pull request, so this is submitted as a comment review.

OPE-45 is close on the post-preparation stages, explicit parameter ownership, replay reuse, and standard/multisector separation. The full W1 contract suite and focused tests are green. Changes are still needed because:

1. The new materialization boundary computes inactive aggregation-error representations and can discard already-computed lazy coordinate results.
2. `run_rhime` still hides retrieval, filtering, basis construction, and labelled assembly behind one coarse preparation call, so the required scientific spine is not visible in executable order.
3. The W2.0 supported surface is exported but not exercised as a composed external-style runner.

The coordinate recomputation is a secondary efficiency concern, not the main blocker in the materialization thread; the potentially unused O(nmeasure²) covariance is the material risk.

openghg_inversions/rhime/runner.py
```diff
@@ -261,87 +550,268 @@ def _validate_multisector_basis_layout(
...
+    "H",
+    "H_bc",
+    "mf",
+    "mf_error",
+    "min_error",
+    "site_indicator",
+)
+
+
+def materialize_rhime_model_inputs(
```
brendan-m-murphy: Could we call this `materialize_pymc_inputs`? “Materialize” advertises the eager/densification and ownership boundary, and “PyMC” is more precise than “model” because complete custom model builders bypass this stage. The RHIME qualifier seems redundant within the RHIME public API.

openghg_inversions/rhime/runner.py
```diff
@@ -261,87 +550,268 @@ def _validate_multisector_basis_layout(
...
+    prepared: RhimePreparedInputs,
+    *,
+    aggregation_error_mode: AggregationErrorMode,
```
brendan-m-murphy, line 577: `AggregationErrorMode` is already a reusable `Literal` type alias, so please keep the alias and list the accepted values (`auto`, `none`, `dense`, `low_rank`, and `diagonal`) with brief semantics in this public docstring.

brendan-m-murphy (commented):
Most of these functions require creating data classes. A data class is fine if options need to be passed around, especially between modules, or if the equivalent in fixedbasisMCMC was a heterogeneous dict. If we can remove any of them, that would simplify the code.

openghg_inversions/rhime/runner.py
brendan-m-murphy: Addressed in 54d50eda: materialization now selects only the aggregation-error representation used by the resolved mode, and the focused Dask test verifies a dormant dense covariance remains lazy. The same update preserves computed auxiliary coordinates in the model-owned copy. Resolving.

openghg_inversions/rhime/runner.py
brendan-m-murphy: Addressed in 54d50eda: both public runners now visibly call retrieve/reload, filter/site alignment, basis construction, sensitivity construction, labelled assembly, alignment, materialization, build, sample, and result stages in execution order. Stage-order tests cover both runners. Resolving.

tests/test_rhime.py
brendan-m-murphy: Addressed in 54d50eda: the new external-style composition test imports the supported RHIME stage surface and passes real handoffs through a complete copied spine without private runner glue or caller-built manifests. Resolving.

openghg_inversions/rhime/runner.py
```diff
@@ -525,11 +1085,63 @@ def run_rhime_from_prepared_inputs(
         multisector=multisector,
     )
 
-    return _execute_prepared_rhime(
+    run_spec = with_prepared_rhime_sites(run_spec, prepared_inputs)
```
brendan-m-murphy, line 1101: **Medium — replay crosses the eager boundary before the named materialization stage.** Earlier in this function, `resolve_aggregation_error(prepared_inputs.inv_inputs, ...)` reads and validates the selected representation. For Dask-backed dense or low-rank inputs this computes the canonical graph before the materialization below, which computes it again for the built-in path. A focused probe observed two executions of the selected covariance graph. Please use non-materializing mode selection for the output-compatibility decision, leave value validation to the materialized build path, and add a Dask replay regression asserting no pre-boundary execution and one execution overall.

openghg_inversions/rhime/runner.py
```diff
@@ -190,24 +220,283 @@ def _apply_output_bundle(result: RhimeResult, bundle: RhimeOutputBundle) -> None
...
+    Args:
+        merged: Filtered data and retained site metadata.
+        basis_functions: Basis used to create ``site_data``.
+        site_data: Per-site observations and sensitivities.
+        data_args: Resolved assembly, minimum-error, and run options.
+
+    Returns:
+        Validated canonical inputs with retained basis and site metadata.
+    """
+    rhime_preparation._set_domain_attrs(site_data, merged.sites, data_args["domain"])
```
brendan-m-murphy: **Medium — this new supported stage mutates caller-owned xarray datasets.** `_set_domain_attrs` assigns `attrs["Domain"]` in place to every supplied `site_data` dataset. The public stage contract does not establish an ownership transfer, and repository guidance treats xarray inputs as borrowed. Please shallow-copy the per-site datasets before attaching metadata (preserving array/Dask sharing), or move the attribute assignment into the stage that already owns the newly created copies. Add a regression asserting the supplied datasets and attrs are unchanged.

brendan-m-murphy (commented):
Follow-up review: the visible standard and multisector spines, mode-aware materialization, and executable public-stage composition test address the earlier three findings; I resolved those threads. Two medium boundary findings remain inline: prepared-input replay eagerly resolves aggregation error before its named materialization boundary, and the public assembly stage mutates caller-owned xarray datasets. With those corrected, this should be ready to approve.\n\nNon-blocking: retain the existing dataclasses; `materialize_rhime_model_inputs` is a better name than `make_rhime_pymc_inputs`; document the five `AggregationErrorMode` values; and prefer a public xarray replacement path over `Dataset._replace` when practical.

openghg_inversions/rhime/runner.py
```diff
@@ -68,7 +77,14 @@
...
 )
 from openghg_inversions.models._rhime_flux import _select_sector_design
-from openghg_inversions.observation_error import resolve_aggregation_error
+from openghg_inversions.observation_error import (
+    AGGREGATION_ERROR_COVARIANCE,
+    AGGREGATION_ERROR_SD,
+    DIAGONAL_RESIDUAL_VARIANCE,
+    LOW_RANK_FACTOR,
+    AggregationErrorMode,
+    _select_aggregation_error_mode,
```
brendan-m-murphy: Why are we importing a private function?

openghg_inversions/rhime/runner.py
brendan-m-murphy, line 75: These are not very concrete are they? Will this be addressed in a later issue?

openghg_inversions/rhime/runner.py
```diff
@@ -502,11 +1068,11 @@ def run_rhime_from_prepared_inputs(
 
     output_spec = run_spec.output
     validate_output_format(output_spec.output_format)
-    aggregation_error = resolve_aggregation_error(
+    aggregation_error_mode = select_aggregation_error_mode(
```
brendan-m-murphy: **Medium — complete-model replay no longer validates the requested aggregation-error representation.** Before this PR, `run_rhime_from_prepared_inputs` called `resolve_aggregation_error` for every route. The new mode-only selector is correct for built-in models, which validate after materialization, but an explicit mode such as `dense` is returned without checking that the covariance exists or is finite/valid. A complete `model_builder` then bypasses both materialization and built-in validation, so inputs that previously failed can now reach building and sampling. Please retain mode-only selection for the built-in/likelihood path, conditionally call `resolve_aggregation_error(...).mode` when `model_builder` is supplied, and add a regression proving an invalid requested representation fails before the custom builder runs.

PR #596, opened 2026-08-17
[OPE-52] Add the low-ceremony run_rhime likelihood seam

brendan-m-murphy:
* **Summary of changes**

Adds the low-ceremony Python-only likelihood seam requested by OPE-52:

- adds an explicit keyword-only `likelihood_builder: RhimeLikelihoodBuilder | None = None` to `run_rhime`;
- exposes the same seam on `run_rhime_multisector`, whose existing contract remains valid because sector roles stay stage-owned and the combined observation mean enters the shared likelihood context;
- uses declared variable roles for custom-likelihood predictive selection and validates output capabilities before sampling;
- rejects non-callable likelihood builders before configuration, acquisition, prepared-input validation, or materialization;
- rejects simultaneous complete-model and likelihood builders before either executes;
- keeps one build/sample/result orchestration path for default and custom likelihoods;
- reports likelihood-specific output compatibility guidance;
- records safe callable identity and JSON-compatible likelihood metadata in `RhimeResult` and serialized provenance;
- keeps callables out of INI parsing, `RhimeRunSpec`, `RhimeModelSpec`, and serialized executable state;
- retains `run_rhime_from_prepared_inputs` as the advanced complete-model/replay boundary;
- adds a preferred one-call executable example using the same project-owned Student-t variation as the retained full copied runner;
- updates the customization and concrete-model documentation.

Default public runner behavior remains unchanged when the seam is omitted. No `model_builder`, strategy class, registry, plugin discovery, or generic orchestration layer is added.

Part of #587.
Linear: [OPE-52](https://linear.app/openghg-inversions/issue/OPE-52/p0-w2a-add-the-low-ceremony-run-rhime-likelihood-strategy-seam).

- [ ] Closes #xxxx (OPE-52 is tracked in Linear; this PR is part of #587)
- [x] Tests added and passing

**Validation**

- `tox -p --parallel-no-spinner`
- `tox -e py310-openghgCur -- -m rhime_contract`
- `tox -e lint`
- focused OPE-52 and review-regression runner, likelihood, output, serialization, config, multisector, and example tests
- Ruff check and format check on changed Python files
- focused Mypy check on the five implementation/example files
- `git diff --check`

The Sphinx build read the updated customization page and all literal includes, then stopped in an unrelated executable grouped-basis notebook because the configured Jupyter kernel has an older OpenGHG without `cf_ureg`. The configured broad Mypy environment also reports existing package-wide typing/stub failures; the focused changed-file check is clean.

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -1,0 +1,14 @@
+/*
+ * PyData Sphinx Theme caps both the page and article widths. That keeps prose
+ * narrow, but it also means widening the browser cannot give long examples
+ * more room and forces unnecessary horizontal scrolling. Let the standard
+ * responsive layout use the available viewport; code blocks retain the
+ * theme's overflow behaviour as a fallback on genuinely narrow screens.
+ */
+.bd-page-width {
+  max-width: none;
+}
+
+.bd-main .bd-content .bd-article-container {
+  max-width: none;
+}
--- a/docs/usage/concrete_rhime_model.rst
+++ b/docs/usage/concrete_rhime_model.rst
@@ -336,9 +336,11 @@ component namespaces are not implemented. They are tracked in
 Alternative models and likelihoods
 ----------------------------------
 
-Direct-Python likelihood and complete-model builders enter only at
-``run_rhime_from_prepared_inputs``. Callables are deliberately not stored on
-``RhimeModelSpec``, so model and run specs remain serializable. There is no
+Direct-Python likelihood builders enter through ``run_rhime``,
+``run_rhime_multisector``, or ``run_rhime_from_prepared_inputs``.
+Complete-model builders remain available only at the prepared-input boundary.
+Callables are never read from configuration or stored on ``RhimeModelSpec`` or
+``RhimeRunSpec``, so model and run specs remain serializable. There is no
 entry-point or config-file plugin registry.
 
 A likelihood builder owns the complete observation component, including its
--- a/docs/usage/concrete_rhime_model.rst
+++ b/docs/usage/concrete_rhime_model.rst
@@ -351,24 +351,25 @@ error policies, aggregation-error mode, and output dimension. This boundary
 avoids a misleading contract in which the runner builds half an error model
 before calling user code.
 
-For the absolute-sigma Gaussian above, no custom modelling function is needed:
+For the absolute-sigma Gaussian above, pass the helper directly to the
+ordinary runner:
 
 .. code-block:: python
 
    from openghg_inversions.rhime import (
        build_absolute_sigma_gaussian_likelihood,
-       run_rhime_from_prepared_inputs,
+       run_rhime,
    )
 
-   result = run_rhime_from_prepared_inputs(
-       prepared_inputs=prepared,
-       run_spec=run_spec,
+   result = run_rhime(
+       config_file="config.ini",
        likelihood_builder=build_absolute_sigma_gaussian_likelihood,
    )
 
-Set ``add_offset=True`` and ``offset_args={"per_site": False}`` on
-``RhimeModelSpec`` to combine it with one global scalar offset. The default
-``per_site=True`` retains the existing site or site-period offset design.
+Pass ``add_offset=True`` and ``offset_args={"per_site": False}`` as Python or
+configuration options to combine it with one global scalar offset. The
+default ``per_site=True`` retains the existing site or site-period offset
+design.
 
 The helper ``build_rhime_observation_state`` is available when only the
 distribution should change. For example, this replaces Normal observations
@@ -382,9 +383,8 @@ scale:
    from openghg_inversions.rhime import (
        RhimeLikelihoodContext,
        RhimeLikelihoodResult,
-       RhimeSampler,
        build_rhime_observation_state,
-       run_rhime_from_prepared_inputs,
+       run_rhime,
    )
 
 
@@ -414,10 +414,8 @@ scale:
        )
 
 
-   result = run_rhime_from_prepared_inputs(
-       prepared_inputs=prepared,
-       run_spec=run_spec,
-       sampler=RhimeSampler(draws=1000, tune=1000, chains=4),
+   result = run_rhime(
+       config_file="config.ini",
        likelihood_builder=student_t_likelihood,
    )
 
--- a/docs/usage/concrete_rhime_model.rst
+++ b/docs/usage/concrete_rhime_model.rst
@@ -371,53 +371,9 @@ configuration options to combine it with one global scalar offset. The
 default ``per_site=True`` retains the existing site or site-period offset
 design.
 
-The helper ``build_rhime_observation_state`` is available when only the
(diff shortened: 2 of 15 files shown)


openghg_inversions/rhime/runner.py
```diff
@@ -1163,13 +1169,21 @@ def run_rhime_from_prepared_inputs(
 def run_rhime(
     *,
     config_file: str | Path | None = None,
+    likelihood_builder: RhimeLikelihoodBuilder | None = None,
```
brendan-m-murphy, line 1216: **[P2] Validate callability before expensive preparation.** A dotted string or other non-callable direct-Python value is not rejected until graph construction, after retrieval, filtering, basis construction, and materialization may have occurred; the eventual error is only `str object is not callable`. Please preflight non-`None` values before acquisition in both ordinary runners, and consistently at the prepared-input boundary.

examples/rhime_customisation/likelihoods.py
```diff
@@ -0,0 +1,19 @@
...
+``likelihood_builder`` is the exported runner seam. Importing this module does
+not build a model, retrieve data, sample, or write outputs.
+"""
+
+from openghg_inversions.rhime import (
+    RhimeLikelihoodBuilder,
+    build_absolute_sigma_gaussian_likelihood,
+)
+
+likelihood_builder: RhimeLikelihoodBuilder = build_absolute_sigma_gaussian_likelihood
```
brendan-m-murphy: **[P2] Show modification, not only built-in selection.** This preferred “project-owned” example only aliases the library’s existing absolute-sigma builder. It proves the seam mechanically but does not teach a scientist how to make a small likelihood change. Please provide a minimal editable builder returning `RhimeLikelihoodResult`, or directly connect this one-call example to the existing Student-t implementation and document the result invariants.

openghg_inversions/rhime/runner.py
```diff
@@ -1222,13 +1238,36 @@ def run_rhime(
         aggregation_error_mode=run_spec.model.aggregation_error_mode,
     )
     build_and_sample_start = timer_start()
-    model_build_result = build_standard_rhime_model(
-        prepared=prepared,
-        model_inputs=model_inputs,
-        run_spec=run_spec,
-    )
-    idata = sample_rhime_model(model_build_result, setup.sampler)
+    if likelihood_builder is None:
```
brendan-m-murphy: **[P2] Keep one orchestration path.** The default/custom branches duplicate build, sample, and result construction here and again in `run_rhime_multisector`. `None` can remain the public signal for the library-owned default while being forwarded through one build call; use `likelihood_builder is not None` only where explicit customization matters (semantic-role selection and provenance). This avoids four paths drifting without hoisting an explicit default callable into the runner, which would complicate provenance and compatibility.

brendan-m-murphy (commented):
The low-ceremony seam is a strong fit for OPE-52, but I recommend a narrow revision before approval. Two related boundary issues could not be anchored to changed lines: the public build stages should reject simultaneous `model_builder` and `likelihood_builder` values (otherwise the model builder silently wins while provenance may record both), and likelihood compatibility failures should not tell users to return a complete-model build result. Along with the inline comments, these are small API-safety, diagnostic, example, and maintainability fixes—not reasons to redesign the seam.

examples/rhime_customisation/likelihoods.py
```diff
@@ -0,0 +1,76 @@
...
+            multivariate Student-t likelihood.
+
+    Notes:
+        On success, this adds RHIME's named observation-state nodes (``Y``,
+        ``error``, ``min_error``, and ``sigma`` or ``epsilon`` as applicable)
+        plus ``student_y`` to the active PyMC model. It performs no sampling or
+        output writes. Unsupported aggregation modes are rejected before any
+        nodes are added.
+    """
+    aggregation_error = resolve_aggregation_error(
```
brendan-m-murphy: **[P2] Avoid resolving unsupported covariance twice.** This calls `resolve_aggregation_error` only to inspect the selected mode, so dense input pays for full covariance validation and an eigenvalue decomposition before being rejected. Accepted diagonal input is then resolved again by `build_rhime_observation_state`. Use `select_aggregation_error_mode` for the cheap preflight and let the state builder perform the single full resolution. If downstream examples should depend only on the RHIME surface, export that selector there rather than teaching a lower-level import.

docs/usage/concrete_rhime_model.rst
```diff
@@ -380,9 +383,8 @@ scale:
    from openghg_inversions.rhime import (
        RhimeLikelihoodContext,
        RhimeLikelihoodResult,
-       RhimeSampler,
        build_rhime_observation_state,
-       run_rhime_from_prepared_inputs,
+       run_rhime,
```
brendan-m-murphy: **[P2] Keep the concrete-model page about the default model.** This second Student-t implementation has already drifted from the tested customization example: it validates aggregation mode after adding PyMC nodes, while the runnable example preflights it. More importantly, this page was introduced to explain the concrete replacement when the compiler removed concrete models; as that decision is reversed, it should not become another customization guide. Prefer showing only the basic default concrete model here. If customization must remain temporarily, include or link the single tested implementation instead of maintaining a second copy.

docs/usage/customising_rhime.rst
```diff
@@ -1,26 +1,85 @@
...
+``run_rhime``. The callable is deliberately unavailable to INI files: it is
+not imported from configuration or stored in a run or model specification.
+
+This example replaces RHIME's Gaussian observation distribution with an
+editable Student-t likelihood while reusing RHIME's current mean and error
+construction. The likelihood is the same one used by the complete copied
+runner below, so the difference between the examples is orchestration ceremony
+rather than scientific behaviour.
+
+The project-owned likelihood selection is one small module:
```
brendan-m-murphy: **[P2] Lead with the low-ceremony call.** The preferred section currently presents the complete 76-line builder and 107-line CLI module before readers see that integration is one `run_rhime(..., likelihood_builder=...)` argument. Start with a minimal direct-Python example that makes the extension seam obvious, then present the editable builder and project/CLI scaffolding as optional detail. That ordering better serves scientists who want to understand the smallest useful change.

docs/_static/custom.css
```diff
@@ -0,0 +1,14 @@
+/*
+ * PyData Sphinx Theme caps both the page and article widths. That keeps prose
+ * narrow, but it also means widening the browser cannot give long examples
+ * more room and forces unnecessary horizontal scrolling. Let the standard
+ * responsive layout use the available viewport; code blocks retain the
+ * theme's overflow behaviour as a fallback on genuinely narrow screens.
+ */
+.bd-page-width {
```
brendan-m-murphy, line 8: **Keep this requested layout improvement unless it causes a concrete problem.** This site-wide width change was explicitly requested and looks like a reasonable general documentation improvement, so please keep it in this PR. Only narrow or revise it if visual checks on representative prose and code-heavy pages reveal an actual readability or responsive-layout regression; its broader scope is intentional here.

PR #597, opened 2026-08-17
Make latest PARIS outputs CF compliant

brendan-m-murphy:
* **Summary of changes**

Implements [OPE-56](https://linear.app/openghg-inversions/issue/OPE-56/investigate-cf-compliant-parisrhime-output-updates) using Fluxie commit [77e6db7](https://github.com/openghg/fluxie/commit/77e6db70291d587ee0558e9efbea95e564b8b0c6) as the reference.

- updates the latest PARIS flux v03 and concentration v04 templates to use valid CF `cell_methods`, attribute-free `time_bnds`, positive-up altitude metadata, and nonempty global comments/references
- orders additional dimensions before time and spatial axes
- replaces repeated covariance dimensions with distinct `country_2` and `sector_2` coordinates while preserving covariance values
- teaches the CDL metadata parser to accept digits in variable and attribute owners
- adds focused and end-to-end regression coverage for in-memory and serialized NetCDF schemas

The root cause was that RHIME's bundled latest templates and output assembly still matched the pre-CF PARIS schema. Existing consumers of `template_version="latest"` must account for the new covariance dimension names/order; legacy-template output is unchanged.

- [x] Tracks OPE-56
- [x] Tests added and passing
- [x] Template metadata documentation updated
- [x] No new requirements

**Validation**

- `tox -p --parallel-no-spinner`
- `.venv/bin/pytest -q -m slow tests/integration/test_multisector_paris_pipeline.py`
- `.venv/bin/pytest -q tests/postprocessing/test_make_paris_outputs.py`
- focused latest PARIS RHIME and fixedbasis regression tests
- Ruff check/format and Pyright on the changed Python module
- `ncgen` compilation of both revised CDL templates


openghg_inversions/postprocessing/make_paris_outputs.py
```diff
@@ -1225,13 +1238,22 @@ def _sector_country_posterior_covariances_kg(
 
 
 def _add_country_covariance(result: xr.Dataset, covariance: np.ndarray, attrs: dict[str, Any]) -> xr.Dataset:
```
brendan-m-murphy, line 1271: Just to check: we cannot actually use this right now? We should track when we will need to include it (e.g. models that include covariance, or later corrected country total calculations)

brendan-m-murphy: Yes — agreed, and the distinction is important. This function is executed today, but the value it writes is only the empirical covariance of country totals reconstructed from the retained-state posterior draws. Because the current basis reduction assigns no uncertainty to the unresolved native state, it is not the complete posterior covariance of native-grid country totals.

That statistical limitation is orthogonal to this PR's CF/schema change: the distinct `country` / `country_2` axes and ordering are still the required representation for whichever covariance semantics we supply.

The coherent-reduction work is tracked by OPE-18 / GitHub #566, and covariance-safe derived outputs broadly by OPE-24 / GitHub #570. I couldn't find an explicit tracker for carrying unresolved-state uncertainty through the country-total functional, so I added [OPE-68](https://linear.app/openghg-inversions/issue/OPE-68/include-unresolved-native-state-uncertainty-in-country-total) as a child of OPE-24, blocked by OPE-18. It covers full posterior country-total moments, a dense native-grid oracle, provenance, and keeping the PARIS stdev/covariance fields coherent.

SutarPrasad (commented):
**CF metadata: use the case-sensitive `Conventions` attribute**

`make_global_attrs()` currently writes the global attribute as lowercase `conventions`, while both revised CDL templates specify `Conventions`. CF uses the case-sensitive global attribute name `Conventions`, so generated files do not advertise their CF version through the standard attribute despite this PR’s CF-compliance objective. Please change the emitted key to `Conventions` and add an assertion for that exact spelling after NetCDF round-tripping.

Satellite inversions are in process will confirm the output soon

brendan-m-murphy:
Addressed in ed8db92a — thank you. The emitted global key is now the case-sensitive `Conventions = "CF-1.8"`, with exact-spelling assertions on both in-memory datasets and raw NetCDF round-trips.

I also checked both complete latest templates and representative emitted flux/concentration files against the [official CF 1.8 conventions](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.8/cf-conventions.html) and IOOS Compliance Checker 6.1.0. That found a few output-side mismatches which are fixed in the same commit:

- xarray was adding `_FillValue` to coordinate variables and `_FillValue` / `coordinates` to `time_bnds` during serialization, even though the templates exclude them;
- emitted flux latitude/longitude remained `float32` rather than the template's declared `double`;
- concentration retained an extra, redundant `site` variable after constructing the template's platform mapping;
- the recommended/template `source` global metadata was missing.

After those fixes, the emitted concentration file passes all CF-1.8 checker tests. The emitted flux file has no checker errors and one warning for the pre-existing `openghg_inversions:non_finite_flux` provenance attribute: the colon does not follow the netCDF naming recommendation. I have not changed that compatibility-sensitive OpenGHG metadata here.

There are three template-level questions worth confirming with the PARIS template authors before changing the format:

1. The concentration v04 template stores observations on `index`, but does not associate longitude, latitude, altitude, or platform labels with the observation variables through `coordinates` (or declare a CF discrete-sampling geometry). [CF Chapter 5](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.8/cf-conventions.html#coordinate-systems) requires auxiliary spatiotemporal coordinates to be associated this way.
2. The flux template's string-valued `country` / `country_2` and `sector` / `sector_2` labels are not named by data-variable `coordinates` attributes as described by [CF section 6.1](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.8/cf-conventions.html#labels). The checker reports the template as passing but throws an internal monotonicity exception on these string axes, so its result is not conclusive for them.
3. `country_fraction` declares `standard_name = "area_fraction"`, whose [standard-name definition](https://cfconventions.org/Data/cf-standard-names/current/build/cf-standard-name-table.html#area_fraction) calls for an `area_type` coordinate. ISO country codes are not CF `area_type` values, so this may need a different representation or no standard name.

I have left those three template semantics unchanged so RHIME continues to match the supplied PARIS schema; they should be resolved with the template owners.

SutarPrasad (approved):

PR #600, opened 2026-08-18
Add modern nested-domain RHIME support

SutarPrasad:
* **Summary of changes**
Core functionality
- Added run_rhime_nested and run_rhime_nested_from_prepared_inputs.
- Retrieves outer and inner footprints/emissions independently.
- Retains each domain’s native grid and basis operator.
- Masks the outer footprint response and prior flux over the inner-domain extent to prevent double counting.
- Aligns inner observations to filtered outer observations:
  - exact timestamps by default;
  - optional nearest matching with an explicit tolerance;
  - unmatched observations raise an error.
- Builds separate H and H_inner sensitivity matrices.
Inversion model
- Added a modern nested PyMC model with separate state vectors:
  - x_outer
  - x_inner
- The forward model is:
  H @ x_outer + H_inner @ x_inner
- Reuses current RHIME boundary-condition, offset, model-error, likelihood, prior, and sampling components.
- Materializes H and H_inner together at the explicit PyMC boundary while preserving borrowed Dask-backed inputs.
Basis handling
- Supports independent inner basis configuration and saved basis files.
- Defaults the inner grid to quadtree when no inner algorithm is specified.
- Supports inner_nbasis.
- When inner_nbasis is omitted, divides the total nbasis budget using the outer/inner sensitivity ratio, bounded to allocate 35–60% to the inner domain.
- Added outer_regions_path, including compatibility with the old outer_region_definition_file spelling.
- Added validation for missing land-sea maps and mismatched grid shapes.
API and documentation
- Exported the nested preparation, model, and runner APIs.
- Added the openghg-inversions run-rhime-nested CLI command.
- Added a nested-domain tutorial covering:
  - surface observations;
  - satellite column observations;
  - INI configuration;
  - prepared-input execution;
  - timestamp alignment;
  - basis allocation and custom outer-region maps.
Output boundary
Nested runs currently require output_format="none". Results retain the complete InferenceData, H, H_inner, and both native basis objects. Existing basic, PARIS, legacy, and InversionOutput writers are rejected because they currently assume a single spatial grid.

- [ ] Tests added and passing


openghg_inversions/rhime/nested.py
```diff
@@ -0,0 +1,880 @@
...
+        time_tolerance=time_tolerance,
+    )
+    native_inner_state_dim = _state_dimension(inner_h, label="Inner")
+    selected_inner = (
+        inner_h.isel(nmeasure=positions)
+        .transpose(native_inner_state_dim, "nmeasure")
+        .rename({native_inner_state_dim: inner_state_dim})
+    )
+    inner_state_coordinate = selected_inner.coords.get(inner_state_dim)
+    combined_dataset = outer.inv_inputs.copy(deep=False)
```
brendan-m-murphy: **[P1] Prepared-input runs can silently double-count the inner domain.** The documented prepared-input workflow combines two ordinary `RhimePreparedInputs` objects, but this copies the ordinary full-domain outer `H` unchanged and only appends `H_inner`. An ordinary outer preparation already aggregates the inner cells, and after aggregation this function no longer has the spatial information needed to remove or verify the overlap. The model therefore produces plausible results from `mu_outer + mu_inner` with the inner emissions represented twice. Please make complementary support a validated property of the handoff before sensitivity aggregation, or reject arbitrary ordinary prepared pairs that cannot prove it.

openghg_inversions/rhime/nested.py
```diff
@@ -0,0 +1,880 @@
...
+
+    fp_all = dict(outer.fp_all)
+    for site in outer.sites:
+        outer_dataset = outer.fp_all[site]
+        inner_dataset = inner.fp_all[site]
+        if not isinstance(outer_dataset, xr.Dataset) or not isinstance(inner_dataset, xr.Dataset):
+            raise TypeError("Modern nested-domain preparation requires per-site xarray Datasets.")
+        if "lat" not in outer_dataset.coords or "lon" not in outer_dataset.coords:
+            raise ValueError(f"Outer-domain dataset for site {site!r} is missing lat/lon coordinates.")
+        site_mask = _extent_mask(
```
brendan-m-murphy: **[P1] The nested handoff does not own one authoritative support mask.** This derives a rectangular mask independently from each site's inner grid, while the retained outer prior flux below receives the union of all site masks. If inner grids differ or are truncated, per-site `fp_x_flux` and the retained outer flux describe different support. That makes the outer sensitivity/basis/prior relationship inconsistent and hides product coverage errors. Please derive and validate one labelled support mask, apply it consistently across sites and retained flux, and add multi-site partial/inconsistent-coverage failures.

openghg_inversions/rhime/nested.py
```diff
@@ -0,0 +1,880 @@
...
+        outer_times = pd.DatetimeIndex(outer_dataset.get_index("time"))
+        inner_times = pd.DatetimeIndex(inner_dataset.get_index("time"))
+        if outer_times.has_duplicates or inner_times.has_duplicates:
+            raise ValueError(f"Nested datasets for site {site!r} contain duplicate timestamps.")
+        order = inner_times.argsort()
+        sorted_times = inner_times.take(order)
+        if tolerance is None:
+            indexer = sorted_times.get_indexer(outer_times)
+        else:
+            indexer = sorted_times.get_indexer(
```
brendan-m-murphy: **[P2] Nearest matching can silently reuse one inner footprint for multiple outer observations.** `DatetimeIndex.get_indexer(method="nearest")` may return the same inner position more than once—for example, two outer times surrounding one inner time within the tolerance. Lines 430–433 then duplicate and relabel that footprint as distinct observations, contrary to the tutorial's no-duplication guarantee. `_inner_measurement_positions` has the same one-to-many behavior. Please centralize the matcher and reject reused or ambiguous inner positions, with a two-outer/one-inner regression test.

openghg_inversions/rhime/nested.py
```diff
@@ -0,0 +1,880 @@
...
+    ):
+        outer_nbasis, allocated_inner_nbasis = _allocate_nested_nbasis(
+            masked_outer,
+            inner_filtered,
+            total_nbasis=outer_args.get("nbasis", 100),
+        )
+        outer_args["nbasis"] = outer_nbasis
+        inner_args["nbasis"] = allocated_inner_nbasis
+
+    outer_prepared = _prepare_one_domain(masked_outer, outer_args)
```
brendan-m-murphy: **[P1] The documented fixed-outer route can mask away the weights it then tries to subdivide.** `masked_outer` has already had the nested extent zeroed, but `fixed_outer_regions_basis` selects the fixed map's maximum-valued inner region as its generation mask. When that region matches the nested footprint extent shown in the tutorial, weighted/quadtree basis generation sees an all-zero cropped field and fails before model construction (or only fits a mismatched fringe). The current nested test stubs `_prepare_one_domain`, so it misses this composition. Please construct the outer basis over the supported complement and add an integration fixture covering nested masking plus `fix_basis_outer_regions=True`.

brendan-m-murphy (commented):
Review notes for the current PR 600 head, posted now so the intrinsic findings are retained while the delivery strategy remains undecided.

I have deliberately scoped the rebase onto current `devel` out of this review. These comments concern behavior in the implementation as it stands. The broader preservation/porting strategy and the need to capture any real-data evidence from this branch are tracked in #666.

The core two-grid equation and much of the labelled preparation are useful. Before treating this branch as a scientific reference or porting its behavior, the four inline correctness concerns below need an explicit resolution and regression evidence.

SutarPrasad:
Creating a new branch as the conflict resolution was messy with lots of new methods created were in the file deleted on devel.

SutarPrasad:
work shifted to https://github.com/openghg/openghg_inversions/pull/668

PR #601, opened 2026-08-18
[OPE-46] Split RHIME preparation into named scientific stages

brendan-m-murphy:
* **Summary of changes**

Moves RHIME retrieval/reload, filtering, basis construction, sensitivity construction, and labelled input construction into a focused `openghg_inversions.rhime.preparation` module while preserving the visible scientific order in both runners. PyMC materialization lives at a separate nearby boundary so the preparation module reads as the scientific transformation of the data.

Adds a Python-only `merged_data` handoff to `run_rhime` and `run_rhime_multisector`, allowing borrowed external or cached merged scientific data to bypass OpenGHG acquisition and resume at filtering. The retrieval stage performs only the scientifically important single-/multi-sector compatibility check; expert-supplied payload structure otherwise follows the existing downstream contracts, and larger caches may contain harmless extra site payloads.

Preserves the existing `RhimePreparedInputs` serialization/reload contract without adding a configuration-snapshot provenance format. Model-bound replay and richer compatibility metadata remain deferred to the planned artifact work.

Documents the external-data boundary and adds tests for external-data bypass, stage order, ownership, layout rejection, lazy materialization, serialization/reload preservation, and existing custom-basis composition.

- [ ] Closes #xxxx (OPE-46; relates to umbrella #587, which remains open)
- [x] Tests added and passing

## Validation

- Focused Python 3.10 current-OpenGHG suite covering RHIME, prepared-input serialization, and custom runners: 287 passed.
- Final external-data simplification subset: 3 passed.
- Focused Ruff: passed.
- Focused mypy on the changed source modules from the initial implementation: passed.
- Full `tox -p --parallel-no-spinner` from the initial ready-to-push branch: `py310-openghgCur` passed the complete suite and `lint` passed. The unrelated `borrowed-types` environment failed because NumPy 2.5 newly deprecates in-place `ndarray.resize`, which pinned mypy 1.19.1 reports in existing `openghg_inversions/borrowed.py:130`.

A slow `test_inversions` run was intentionally skipped after review because this is a mechanical ownership/refactoring change with no numerical-method or HPC runtime boundary change.


openghg_inversions/rhime/preparation.py
```diff
@@ -0,0 +1,399 @@
...
+    data_args: Mapping[str, Any],
+    *,
+    merged: RhimeMergedData,
+    basis_functions: BasisFunctions,
+) -> dict[str, Any]:
+    """Describe the model-independent preparation that owns an assembled cache."""
+    metadata = {
+        name: _json_preparation_value(data_args[name])
+        for name in _PREPARATION_PROVENANCE_OPTIONS
+        if name in data_args
```
brendan-m-murphy: The metadata includes only options present in `data_args`, while the stages independently apply defaults. For example, omitting `basis_algorithm` and `nbasis` executes `weighted`/`100`, but neither value is serialized. Acquisition selectors such as inlet, instrument, platform, footprint model/height, and calibration scale are also absent. That leaves the artifact without enough provenance for the compatibility validation required by OPE-46. Could we resolve defaults once at the runner boundary and record every data-affecting selector?

openghg_inversions/rhime/preparation.py
```diff
@@ -0,0 +1,399 @@
...
+
+    requested_sites = {str(site).upper() for site in data_args["sites"]}
+    unexpected_sites = [site for site in merged_data.sites if site not in requested_sites]
+    if unexpected_sites:
+        raise ValueError(
+            f"External RHIME merged data contains site(s) outside the resolved run: {unexpected_sites!r}."
+        )
+
+    stored_multisector = merged_data.fp_all.get(".split_by_sectors")
+    if stored_multisector is not None and bool(stored_multisector) != multisector:
```
brendan-m-murphy: This validates `site_options` but not the corresponding `fp_all` contents. A `RhimeMergedData` declaring site `TAC` without `fp_all['TAC']` passes here and later fails with a low-level `KeyError` during filtering. The sector marker is optional as well, so an untyped handoff is accepted by either recipe. Please verify every retained site exists and require an explicit compatible sector layout at this owning boundary.

brendan-m-murphy (commented):
The extraction follows the new RHIME locality guidelines well and keeps the scientific stage order visible. I found two cache-boundary issues to address: preparation provenance is not fully resolved, and external merged-data validation does not yet validate the promised contents. The legacy fixedbasis ownership cleanup is tracked separately as OPE-83.

openghg_inversions/rhime/_validation.py
```diff
@@ -0,0 +1,50 @@
...
+    unexpected_sites = [site for site in merged_data.sites if site not in resolved_sites]
+    if unexpected_sites:
+        raise ValueError(
+            f"External RHIME merged data contains site(s) outside the resolved run: {unexpected_sites!r}."
+        )
+
+    retained_sites = set(merged_data.sites)
+    stored_sites = {str(name).upper() for name in merged_data.fp_all if not str(name).startswith(".")}
+    missing_site_data = sorted(retained_sites - stored_sites)
+    undeclared_site_data = sorted(stored_sites - retained_sites)
```
brendan-m-murphy: This validation now feels disproportionate to the expert-facing Python handoff. In particular, rejecting harmless extra site payloads prevents reuse of a larger external cache even though downstream stages iterate only over `merged_data.sites`. My earlier review pushed this boundary too far: please keep only the scientifically important single-/multisector compatibility check, inline it in `retrieve_or_reload_rhime_data`, and let malformed payloads fail naturally. That should remove this module and much of its test code while making the scientific flow easier to follow.

brendan-m-murphy (commented):
One remaining simplification request; I found no numerical or scientific correctness issues in the revised flow.

PR #603, opened 2026-08-18
Add exact coherent Gaussian reduction

brendan-m-murphy:
## Summary of changes

Implements the first exact in-memory coherent Gaussian reduction for OPE-18.

- adds one atomic eager boundary that constructs OPE-17 dense covariance products and reduces the same labelled `B/H/Pi/m` set
- expresses scientific dimension roles with xarray transposition, then uses one documented `xr.align(..., join="exact")` before computation so dot products cannot silently drop unmatched native cells
- materializes related Dask mean/sensitivity payloads together while preserving the pre-compute xarray indexes
- trusts the products constructed locally and keeps the reduction kernel readable as `Pi m`, `H m`, the affine intercept, and the Schur-complement covariance
- relies on static types rather than adding routine runtime type checks
- leaves physical unit conversion/compatibility at the upstream OpenGHG/pint-xarray boundary and propagates prepared unit metadata
- symmetrizes the unresolved covariance for floating-point roundoff; scientific PSD and Gaussian-equivalence identities are covered in tests
- intentionally defers likelihood adapters, LRPD approximation, persistence, and arbitrary reconstruction

This is stacked on draft PR #602. That base PR now contains the uniform OPE-17 validation cleanup, retained-index materialization fix, and repository guidance for reasonable validation, xarray alignment, and Pint unit boundaries.

## Checklist

- [ ] Closes a GitHub issue — OPE-18 remains open for later roadmap slices
- [x] Tests added and passing
- [x] Documentation and tutorials updated
- [x] CHANGELOG updated
- [x] No new requirements

## Validation

- 146 focused coherent-reduction/covariance/native/source/serialization tests pass
- Ruff passes on the feature and covariance stack
- the base PR's full Python 3.10 current-OpenGHG suite passes: 1,329 passed, 12 skipped, 6 xfailed
- default `borrowed-types` retains the known unrelated NumPy 2.5 Mypy deprecation for `ndarray.resize` in unchanged `borrowed.py`
- strict docs renders the new development and coherent-reduction pages, but remains red on existing repository-wide autodoc warnings/errors


docs/usage/coherent_reduction.rst
```diff
@@ -0,0 +1,127 @@
...
+
+   A = HBH^\mathsf{T} - H_\alpha C_\alpha H_\alpha^\mathsf{T}.
+
+All three retained quantities---the prior, effective forward operator, and
+unresolved covariance---come from the same product set.
+
+Preparing the exact result
+--------------------------
+
+Pass the native covariance, canonical basis prolongation, labelled native mean,
```
brendan-m-murphy: Above you mention the projection/restriction operator Pi, but the functions here take the prolongation U. I think we want to accept both eventually... it might be worth mentioning that we can/will accept either and construct the other operator from the given one and the covariance.

docs/usage/coherent_reduction.rst
```diff
@@ -0,0 +1,127 @@
...
+       native_sensitivity=native_sensitivity,
+       observation_dim="observation",
+   )
+
+   retained_mean = reduction.retained_mean
+   retained_covariance = reduction.retained_covariance
+   effective_operator = reduction.effective_observation_operator
+   unresolved_covariance = reduction.unresolved_observation_covariance
+
+The operation is a named eager boundary. Related Dask-backed mean, sensitivity,
```
brendan-m-murphy: 'Named eager boundary' is jargony. Just say that lazy/dask arrays will be computed when you call this function

docs/usage/coherent_reduction.rst
```diff
@@ -0,0 +1,127 @@
...
+and basis-prolongation payloads are densified lazily where necessary and
+materialized together. The covariance product blocks are built and reduced
+locally, so a sensitivity cannot be substituted after projection. The result
+fields are eager xarray arrays with explicit retained and observation labels.
+
+Boundary contract and limits
+----------------------------
+
+The public operation transposes inputs into their declared scientific roles
```
brendan-m-murphy: Be concrete; what is the 'public operation'? This whole section is probably too detailed and jargony. Trim it down substantially. If there is general policy here it should be in a dedicated note. In this document, prioritize information with scientific or performance implications; code boundaries should be described briefly.

docs/usage/native_covariance.rst
```diff
@@ -174,13 +174,12 @@ and the centred conditional observation model
...
-OPE-18 owns that solve-based transformation, unresolved covariance, and the
-single coherent artifact that binds prior, forward, residual, reconstruction,
-and state-treatment information. The OPE-17 product blocks are inputs to that
-work; they must not be described or persisted as though they were already the
-complete artifact. Arbitrary reporting-function products :math:`Q`,
-low-rank-plus-diagonal numerical views, and likelihood integration are also
-outside this API.
+The :doc:`coherent_reduction` operation owns the first exact solve-based
+transformation, centred forward model, and unresolved observation covariance.
+The OPE-17 product blocks are inputs to that work; they must not be described or
```
brendan-m-murphy: Don't reference linear issues here, this is user documentation

PR #605, opened 2026-08-18
[OPE-81] Split standard and multisector RHIME recipes

brendan-m-murphy:
* **Summary of changes**

Splits the standard and multisector RHIME workflows into separate, readable scientific recipe modules.

- `rhime/standard.py` owns the complete single-sector resolve → retrieve/reload → filter → basis → sensitivities → assemble → materialize → build → sample → output process.
- `rhime/multisector.py` owns the corresponding source-resolved process, including basis-layout validation and multisector output choices.
- `rhime/prepared.py` preserves the advanced prepared-input entry point.
- Shared scientific stages now live with parameter resolution, preparation, sampling, and output ownership; `rhime/runner.py` contains compatibility re-exports only.
- Public imports remain compatible, and documentation points scientists to the concrete recipe modules.

This implements OPE-81 and leaves nested-domain and CO₂-family recipes free to consume the shared preparation stages directly.

- [ ] Closes #xxxx (OPE-81 is tracked in Linear rather than a GitHub issue)
- [x] Tests added and passing
- [x] Added any new requirements to `requirements.txt` (no new requirements)

Validation:

- `tox -e py310-openghgCur -- tests/test_prepared_inputs_serialisation.py tests/test_xarray_input_adapter.py tests/integration/test_cookiecutter_rhime_consumer.py tests/test_rhime.py -k 'not api_smoke'`: 349 passed.
- Focused Python 3.12 prepared-input/serialization suite: 120 passed.
- Focused recipe ordering, failure-order, compatibility, and no-write contracts: 19 passed.
- Ruff: passed.
- The two excluded API smoke cases execute both recipes successfully but currently fail on OpenGHG fixture `long_name` drift (`number_of_observations` is now included).
- Configured mypy reaches two existing errors in `postprocessing/_country_codes.py`; borrowed-types reaches an unrelated NumPy 2.5 deprecation diagnostic in `borrowed.py`.


openghg_inversions/rhime/standard.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+def run_rhime(
+    *,
+    config_file: str | Path | None = None,
+    merged_data: RhimeMergedData | None = None,
+    likelihood_builder: RhimeLikelihoodBuilder | None = None,
+    **kwargs: Any,
+) -> RhimeResult:
+    """Run the standard single-sector RHIME scientific recipe.
```
brendan-m-murphy: Could we preserve the detailed public runner docstring that was on `runner.run_rhime`? Most scientific options are still hidden in `**kwargs`, so removing the argument, return, error, and usage documentation makes `help(run_rhime)` and generated API docs substantially less informative. The same applies to `run_rhime_multisector`, where the previous docstring explained source-count and sector-mapping constraints.

openghg_inversions/rhime/standard.py
```diff
@@ -0,0 +1,208 @@
...
+        if config_file is not None
+        else dict(kwargs)
+    )
+    setup = resolve_rhime_options(params=params, multisector=False)
+
+    preparation_start = timer_start()
+    merged = retrieve_or_reload_rhime_data(
+        setup.data_args,
+        multisector=False,
+        **({} if merged_data is None else {"merged_data": merged_data}),
```
brendan-m-murphy: This conditional dictionary unpacking adds indirection to the main scientific recipe. `retrieve_or_reload_rhime_data` already accepts `merged_data=None`, so can this be the direct `merged_data=merged_data` keyword instead? That keeps the optional handoff explicit and removes machinery we do not need. The multisector recipe has the same pattern.

brendan-m-murphy (commented):
Two readability suggestions from a YAGNI-focused review. The explicit recipe split itself looks appropriately simple and local.

PR #606, opened 2026-08-19
[OPE-47] Colocate concrete RHIME recipe models

brendan-m-murphy:
## Summary of changes

- place the standard and multisector PyMC graphs beside their procedural runners
- expose one unambiguous concrete graph builder per recipe:
  - `build_standard_rhime_model(...) -> pm.Model`
  - `build_multisector_rhime_model(...) -> pm.Model`
- name orchestration adapters explicitly as `*_model_result` and keep spec adapters private
- move RHIME specs, likelihoods, and source/sector helpers under `openghg_inversions.rhime`; remove the `models.rhime` path and all `models -> rhime` facades
- construct pollution, boundary, offset, baseline, and completed model means visibly in each recipe before invoking a likelihood
- pass custom likelihoods the completed modelled concentration plus separately named pollution/baseline quantities used by mismatch scaling
- use `PriorArgs` consistently and restore full Google-style API documentation
- delete the unused semantic flux compiler and remove `builder_strategy` from specs, configuration, metadata, docs, and tests
- retain pure source/sector/prior/state-activity resolution as ordinary RHIME helpers

The generalized component-input contract remains deferred to Linear OPE-82. Backend-neutral role documentation remains tracked in OPE-66. PyTensor precision policy comes from merged PR #610; this PR now invokes that policy before importing PyMC through the RHIME package.

Linear: OPE-47  
Part of #587

- [x] Tests added and passing for changed ownership and forward-model composition
- [x] Documentation updated
- [x] No new requirements

## Validation

- SLURM tox job `18626338`: lint passed; Python 3.10 completed 1,341 tests and identified three focused failures in the new import/composition checks
- SLURM tox job `18626423`: all five focused Python 3.10 follow-up tests passed after those fixes
- focused local state, likelihood, observation-error tests: 55 passed
- focused ownership, graph, composition, and extension tests: 34 passed across two runs
- broader affected local suite: 353 passed; three pre-existing environment/fixture failures remain (h5netcdf H5DS read and two OpenGHG `long_name` expectations)
- fresh-process imports pass with no `models -> rhime` facade
- Ruff, Python compilation, and `git diff --check` passed


openghg_inversions/models/rhime.py
```diff
@@ -1,893 +1,35 @@
-"""RHIME model builders.
```
brendan-m-murphy, line 1: I'm not sure we need this facade... this is unreleased code, although other PRs might depend on it (and I suppose inlining the forwarded imports would be simple enough).

A real issue here is that the `models` submodule now depends on the `rhime` submodule.

openghg_inversions/rhime/__init__.py
```diff
@@ -79,10 +53,67 @@
...
+    "requirements_for_model",
     "run_rhime",
     "run_rhime_from_prepared_inputs",
     "run_rhime_multisector",
     "sample_rhime_model",
     "select_aggregation_error_mode",
     "with_prepared_rhime_sites",
 ]
+
+_PUBLIC_MODULES = {
```
brendan-m-murphy: Are we re-exporting model components here?

openghg_inversions/rhime/contracts.py
```diff
@@ -0,0 +1,331 @@
+"""Component-owned input and output contracts beside the RHIME graph.
```
brendan-m-murphy: In retrospect, this should not have been on the same issue as moving the models closer to the runners.

openghg_inversions/rhime/contracts.py
```diff
@@ -0,0 +1,331 @@
...
+
+from __future__ import annotations
+
+from collections.abc import Mapping
+from dataclasses import dataclass
+from typing import Literal, TypeAlias
+
+import xarray as xr
+
+from .model import (
```
brendan-m-murphy: [P1] This import participates in an import-order-dependent cycle. In a fresh process, `import openghg_inversions.rhime.contracts` follows `contracts.py -> rhime.model -> models.__init__ -> models.rhime -> rhime.model`, where the compatibility facade requests names from the partially initialized `rhime.model`. The result is an `ImportError`, and `tests/test_rhime_model_contract.py` cannot collect in isolation. Importing `openghg_inversions.models` first masks the defect, which explains why a broader suite can pass. Please remove the reverse dependency/facade and add a fresh-process import or isolated contract-test check.

openghg_inversions/rhime/model.py
```diff
@@ -0,0 +1,866 @@
...
+            offset_args=offset_args,
+            power=power,
+            likelihood_builder=likelihood_builder,
+        )
+        setattr(model, _LIKELIHOOD_RESULT_ATTR, likelihood_result)
+
+    return model
+
+
+def build_rhime_model(
```
brendan-m-murphy: [P1] This is the actual readable standard graph, while `standard.py::build_standard_rhime_model` is only a delegating wrapper. The current development guidance says a production recipe contains its runner and readable concrete builder, and the model-family plan suggests converging the existing recipes toward one recipe per module when they are next changed substantially. Please keep this graph beside `run_rhime` in `standard.py`; the multisector graph at line 586 should similarly live in `multisector.py`. Shared ordinary components can remain shared.

openghg_inversions/rhime/model.py
```diff
@@ -0,0 +1,866 @@
...
+from dataclasses import dataclass, field
+from typing import Any, cast
+
+import pymc as pm
+import pytensor.tensor as pt
+import xarray as xr
+from pytensor.tensor.variable import TensorVariable
+
+from openghg_inversions.inversion_inputs import DatetimeLike
+from openghg_inversions.models._rhime_compiler import _compile_loop_sum, _FluxPlan
```
brendan-m-murphy: [P1] The compiled strategy has been removed from public configuration and dispatch, but this import and the private compiled builders below keep the compiler in production model ownership. W5 says to preserve useful pure source/sector/state-selection functions independently, then delete the compiler or archive a uniquely useful research snapshot under an explicitly experimental namespace. Please remove this dependency and the compiled assembly/builders from the production recipe.

openghg_inversions/_pymc_config.py
```diff
@@ -2,13 +2,35 @@
...
+    runtime is left untouched because changing process-wide tensor defaults
+    after graph libraries have initialized is ambiguous and unsafe.
+    """
+    pytensor_was_loaded = "pytensor" in sys.modules
     import pytensor
 
     config = cast(Any, pytensor).config
-    config.floatX = "float32"
-    config.warn_float64 = "warn"
+    if pytensor_was_loaded:
```
brendan-m-murphy, line 31: [P2, already separated] The process-wide precision policy has now landed independently in PR #610. PR 606 still contains its earlier copy at the current head, so please rebase onto current `devel` and drop the duplicate precision commit/delta here. My isolated failure applied to the present PR 606 head: running `tests/test_rhime.py` alone imports PyMC first, retains `float64`, and then `test_concrete_and_compiled_standard_models_are_equivalent` assumes `hx` is `float32`. Please re-run that focused test after rebasing; if PR #610 already corrected the assertion, no further precision work is needed in this PR.

brendan-m-murphy (commented):
I think this needs re-scoping before detailed review continues.

The older W5 ownership sketch did allow a shared `rhime/model.py`, which explains the current implementation. However, the newer development guidance is more specific: each production recipe contains its runner and a readable nearby concrete builder, with model-specific composites retained beside that recipe. Since this PR substantially changes the standard and multisector models, I think the standard graph should live with `standard.py` and the multisector graph with `multisector.py`, sharing only ordinary scientific components whose reuse is already demonstrated.

I also think the compatibility facade/re-exports should be removed rather than cementing the `models -> rhime -> models` dependency. Apart from conflicting ownership, that reverse dependency now causes a reproducible fresh-import failure noted inline.

I originally asked for the PyTensor precision fix in this PR because it was blocking CI and I wanted to avoid an extra PR. That was reasonable as a tactical unblock, but it has proved to define process-wide startup semantics and focused-test behavior. Since it is already isolated in commit `7f8a463`, I now think it should be split out and reviewed independently.

Suggested scope:
1. Put the standard and multisector concrete graphs beside their respective recipes.
2. Remove the unreleased `models/rhime.py` facade and reverse re-exports.
3. Remove the compiler from production ownership, or explicitly archive it under an experimental namespace only if there is a documented unique research use.
4. Split the generalized input/output contract work into a follow-up if it cannot remain small and local.
5. Extract the PyTensor precision-policy commit into its own PR.

The contract is currently reconstructed across `model.py`, `models/rhime_likelihood.py`, `contracts.py`, and `_model_building.py`, with a large dynamic import registry added to hold the public surface together. That is more indirection than the simplicity/locality directive calls for, and it does not yet give the graph contract one obvious ownership location.

brendan-m-murphy:
Correction to my review: the PyTensor precision-policy work has already landed independently in PR #610. Item 5 in the review should therefore read “rebase PR 606 onto current `devel` and drop its duplicate precision commit/delta,” not “open another PR.” I have updated the inline precision comment accordingly. The model ownership, circular import, retained compiler, and contract-scope findings are unchanged.

openghg_inversions/rhime/standard.py
```diff
@@ -44,6 +56,114 @@
...
+            inv_inputs["H"],
+            data_name="hx",
+            prior_args=x_prior,
+            var_name="x",
+            output_name="mu",
+            output_dim="nmeasure",
+            compute_deterministic=True,
+            state_activity=state_activity,
+        )
+        likelihood_result = _add_rhime_observation_components(
```
brendan-m-murphy: **Forward-model composition is still hidden behind the likelihood seam.** `_add_rhime_observation_components` constructs `H_bc @ bc` and the offset, and the likelihood builder later sums flux + boundary + offset. Boundary conditions and offsets are baseline/forward-model terms, not likelihood terms; the development guide explicitly separates those responsibilities. This also means a custom likelihood can accidentally omit or alter part of the modeled mean. Please construct the flux and baseline/total mean visibly in this recipe, then pass the completed mean (plus separately named pollution/baseline quantities only where the mismatch-error calculation genuinely needs them) to the likelihood. The same issue exists in the multisector recipe.

openghg_inversions/rhime/standard.py
```diff
@@ -44,6 +56,114 @@
...
+    pollution_events_from_obs: bool = False,
+    no_model_error: bool = False,
+    aggregation_error_mode: AggregationErrorMode = "auto",
+    offset_args: dict | None = None,
+    power: dict | float = 1.99,
+    state_activity: StateActivity | None = None,
+    bc_state_activity: StateActivity | None = None,
+    likelihood_builder: RhimeLikelihoodBuilder | None = None,
+) -> pm.Model:
+    """Build the concrete standard single-sector RHIME model.
```
brendan-m-murphy, line 87: **Please restore full Google-style documentation for the moved APIs.** This public function has 15+ parameters but its former `Args`, `Returns`, and relevant `Raises` sections were reduced to a narrative summary during the move. The same regression affects `build_rhime_model_from_spec`, both corresponding multisector builders, and `RhimeModelSpec`; the complex private `_add_rhime_observation_components` also now has only a one-line docstring despite its large, non-obvious signature. Per project policy, public functions need full Google-style docstrings, and non-trivial private functions need enough argument/return documentation to be usable.

openghg_inversions/rhime/standard.py
```diff
@@ -11,11 +11,23 @@
...
 import arviz as az
+import pymc as pm
 import xarray as xr
 
 from openghg_inversions._timing import log_timing, timer_seconds, timer_start
 from openghg_inversions.inversion_data import RhimeMergedData, RhimePreparedInputs
-from openghg_inversions.models import RhimeLikelihoodBuilder, build_rhime_model_from_spec
+from openghg_inversions.models.components import add_state_linear_component
+from openghg_inversions.models.coords import CoordRegistry, attach_coord_registry
+from openghg_inversions.models.rhime import (
```
brendan-m-murphy: **The ownership move is still incomplete.** The concrete recipe now lives under `rhime`, but it imports RHIME-specific specs, defaults, and several private assembly helpers back from `models.rhime`. That leaves the scientific recipe split across both packages and makes `rhime/specs.py`'s statement that specs live with the builders inaccurate. OPE-47's W5 calls for a cohesive RHIME ownership boundary. A simpler division is: `rhime` owns RHIME specs/defaults and RHIME-specific composition; `models` owns genuinely reusable PyMC primitives, priors, coordinates, and state operations.

openghg_inversions/rhime/__init__.py
```diff
@@ -43,6 +45,8 @@
 from .prepared import run_rhime_from_prepared_inputs
 from .sampling import RhimeSampler, sample_rhime_model
 from .standard import (
+    build_rhime_model,
```
brendan-m-murphy: **Please reduce or clarify the public builder surface.** The package exports `build_rhime_model`, `build_rhime_model_from_spec`, and `build_standard_rhime_model`, plus three similarly named multisector variants; some return `pm.Model` and others return `RhimeModelBuildResult`. That conflicts with the guidance to provide one obvious concrete builder per recipe. Prefer making stage/spec adapters private, or give the graph and orchestration-stage builders names that make their distinct contracts unmistakable.

openghg_inversions/rhime/standard.py
```diff
@@ -44,6 +56,114 @@
 from .specs import RhimeRunSpec
 
 
+def build_rhime_model(
+    inv_inputs: xr.Dataset,
+    *,
+    sigma_alignment: SigmaAlignment,
+    x_prior: dict | None = None,
```
brendan-m-murphy: **Use the existing prior vocabulary consistently.** `models.priors` already defines `PriorArgs = dict[str, Any]`, and the default prior constants use it, while these new public signatures fall back to bare `dict`. Please annotate the prior arguments (and corresponding multisector/spec fields) with `PriorArgs`. I would not expand this PR by inventing a new prior class without concrete validation needs, but reusing the existing alias is a small clarity improvement.

brendan-m-murphy (commented):
This is substantially improved: the concrete standard and multisector recipes are now visible in their recipe modules, the compiler path is gone, and the PyTensor precision change is correctly outside this PR.

I still don't think this is ready to merge. The inline comments cover the main OPE-47 issues: forward-model composition is still hidden inside the likelihood seam, RHIME-specific ownership remains split across `rhime` and `models.rhime`, moved public/private APIs have lost required documentation, the public builder surface is ambiguous, and the existing `PriorArgs` vocabulary is not used consistently.

One additional cleanup is `openghg_inversions/rhime/runner.py`. W4b called it a temporary compatibility surface, but this code is unreleased and the scientist-facing legacy entry point `hbmcmc/run_hbmcmc.py` imports from `openghg_inversions.rhime`, not `rhime.runner`. Repository references are limited to the compatibility module, its tests, and planning/documentation text. Please remove it in this PR unless there is a concrete external dependency that the repository does not show; if such a dependency exists, document it and track retirement separately.

I also found one pre-existing correctness issue while following this path: the default likelihood always creates inferred `sigma`, even when `no_model_error=True`, and then leaves it disconnected from the likelihood. The absolute-sigma implementation already guards its creation. Please either fix this small graph bug here with an assertion that `sigma` is absent, or open an explicit follow-up issue before merge.

This review supersedes my outdated comments on the earlier version of the PR.

openghg_inversions/postprocessing/inversion_output.py
```diff
@@ -230,6 +230,7 @@ def _add_attributes_to_trace_dataset(trace_ds: xr.Dataset, obs_units: str, obs_l
     "model_error": "epsilon",
     "concentration": "y",
     "baseline": "mu_bc",
```
brendan-m-murphy, line 232: This depends on what is included in the baseline. It might include an offset, and in future work, might include flux contributions from "outer regions"

openghg_inversions/postprocessing/make_outputs.py
```diff
@@ -711,24 +711,48 @@ def make_concentration_outputs(
...
+            conc_roles.append(offset_role)
 
     trace = _rename_trace_roles(
         inv_out.trace_dataset(var_roles=conc_roles),
         inv_out,
-        {concentration_role: "y", baseline_role: "mu_bc", offset_role: "offset"},
+        {
+            concentration_role: "y",
+            baseline_role: "mu_bc",
+            boundary_role: "mu_bc",
```
brendan-m-murphy, line 741: This is correct, but the baseline can be different if there are other terms; I'm not sure if we have considered how to represent this in the outputs. For the "basic" outputs, we can define a convention; for "paris" outputs, we need to consult with others.

openghg_inversions/rhime/_model_building.py
```diff
@@ -69,19 +123,36 @@ def validated_custom_model_build(
...
     if not isinstance(result, RhimeModelBuildResult):
         raise TypeError(
             "A RHIME model builder must return `RhimeModelBuildResult`; "
             f"got {type(result).__name__}."
         )
-    validate_model_build_result(result, context=context)
     return result
 
 
 def validate_likelihood_builder(likelihood_builder: object | None) -> None:
```
brendan-m-murphy: Why do we need to validate a component we created?

openghg_inversions/rhime/standard.py
```diff
@@ -41,18 +54,292 @@
...
+    DEFAULT_BC_PRIOR,
+    DEFAULT_OFFSET_PRIOR,
+    DEFAULT_SIGMA_PRIOR,
+    DEFAULT_X_PRIOR,
+    RhimeModelSpec,
+    RhimeRunSpec,
+)
+
+
+def _prepare_builder_priors(
```
brendan-m-murphy: I would inline the helper, although perhaps we need to make this change at a config parsing boundary. Setting defaults is probably not a great idea...

openghg_inversions/rhime/standard.py
```diff
@@ -41,18 +54,292 @@
...
+            if baseline_mean is None:
+                baseline_mean = offset
+            elif preserve_legacy_likelihood:
+                baseline_mean = baseline_mean + offset
+            else:
+                baseline_mean = pm.Deterministic(
+                    "mu_baseline",
+                    baseline_mean + offset,
+                    dims="nmeasure",
+                )
```
brendan-m-murphy: **[P1] Avoid persisting the complete baseline.** `mu_baseline` duplicates the already retained observation-length `mu_bc` and `offset` vectors for every chain and draw. This can add hundreds of MB to a realistic trace. Please keep the sum as an unnamed expression for likelihood construction and reconstruct or label the complete baseline at the output boundary. The same issue occurs in `rhime/multisector.py` lines 314–318.

openghg_inversions/models/additive_sigma.py
```diff
@@ -0,0 +1,204 @@
...
+
+    Returns:
+        The observed Gaussian variable, named ``y``. The total marginal error
+        scale is also recorded in the active model as ``epsilon``.
+
+    Raises:
+        ValueError: If the observation or aggregation-error inputs are
+            inconsistent with ``output_dim``.
+    """
+    del pollution_mean, pollution_event_baseline, power, pollution_events_from_obs
```
brendan-m-murphy: **[P1] Separate the reusable additive-sigma component from its RHIME adapter.** This public models-owned likelihood requires four pollution-event parameters and then discards them here. A CO₂ or other non-RHIME model would have to fabricate scientifically meaningless values. Please give the reusable component a signature containing only its mathematical inputs and place a small callback-shaped adapter at the RHIME recipe seam.

openghg_inversions/rhime/multisector.py
```diff
@@ -42,15 +72,333 @@
...
+                if baseline_mean is None
+                else pm.Deterministic(
+                    "mu_baseline",
+                    baseline_mean + offset,
+                    dims="nmeasure",
+                )
+            )
+        modelled_mean = pollution_mean if baseline_mean is None else pollution_mean + baseline_mean
+
+        likelihood_component = likelihood_builder or build_pollution_event_gaussian_likelihood
```
brendan-m-murphy: **[P2] Select the default only when the builder is `None`.** The public contract accepts arbitrary callable objects, including valid callables with a false truth value. With `or`, such a builder is silently ignored. Please use the same explicit `is None` selection as the standard recipe.

openghg_inversions/rhime/__init__.py
```diff
@@ -11,19 +11,24 @@
...
+
+# PyTensor precision must be selected before importing likelihoods, recipes, or
+# any other module that imports PyMC.
+from openghg_inversions._pymc_config import configure_pytensor
+
+configure_pytensor()
+
 from openghg_inversions.observation_error import select_aggregation_error_mode
+from openghg_inversions.models.additive_sigma import add_additive_sigma_gaussian_likelihood
+from openghg_inversions.models.pollution_event import build_pollution_event_gaussian_likelihood
```
brendan-m-murphy: **[P2] Remove the remaining import facade.** These aggregation/model components already have direct owner modules, but re-exporting them from `rhime` creates duplicate ownership and preserves the facade pattern this re-scope was meant to remove. Please have examples and users import the components from their owning modules.

CHANGELOG.md
```diff
@@ -83,8 +83,8 @@
   inputs or a complete model.
   Custom builders declare semantic variable roles, supported output formats,
   and serializable provenance metadata, which are validated before sampling
-  and preserved in inversion outputs. Built-in concrete, compiled, and
-  multisector models use the same result contract, while the sampler resolves
+  and preserved in inversion outputs. Built-in standard and multisector
+  models use the same result contract, while the sampler resolves
   posterior predictives by semantic role. Likelihood results now retain
   serializable metadata and the runner persists custom likelihood-builder
   identity. An opt-in absolute-sigma Gaussian implements observation variance
```
brendan-m-murphy: **[P2] Update this entry to match the final API.** Likelihood builders no longer return result records declaring roles, formats, and metadata; they create canonical `y`/`epsilon` variables, while the complete-model builder retains the result contract. This also still calls the restored component “absolute-sigma”; the current name is “additive sigma.”

docs/reference/openghg_inversions.models.rst
```diff
@@ -11,6 +11,5 @@ openghg\_inversions.models
 .. toctree::
    :maxdepth: 4
 
-   openghg_inversions.models.rhime
-   openghg_inversions.models.rhime_likelihood
+   openghg_inversions.models.additive_sigma
```
brendan-m-murphy, line 14: **[P2] Document the actual supported import surface.** The generated customization example imports the public `models.pollution_event` module, but this reference toctree only adds `additive_sigma`. The usage page also claims that the generated example uses only names from `openghg_inversions.rhime`, although it imports `models.pollution_event`, `observation_error`, and `sigma`. Please update that statement and add reference coverage for the supported direct imports.

brendan-m-murphy (commented):
Review of `5f89b9a2`. I have left six findings inline. One further documentation finding cannot be anchored because the file is not changed in this PR:

**[P2] Update the active model-family roadmap after removing the compiler.** `docs/plans/rhime_model_family_expansion.md` lines 300–306 still instruct future work to extract helpers from the deleted `models/_rhime_flux.py` before removing the compiler. Please record that extraction/removal as complete and point follow-up work to the current `models/_flux.py` and recipe ownership.

Compatibility checks found no regression at the `run_hbmcmc.py` boundary: aggregation error remains opt-in, modern PEFO uses the complete baseline, and the legacy path retains its boundary-only behavior.

Issue #607, opened 2026-08-19
Define component-level numerical precision and accumulation policy


brendan-m-murphy:
## Problem

OpenGHG Inversions currently uses PyTensor's process-wide `floatX` setting to keep PyMC model data and graphs in float32. This avoids silently promoting float32 footprints when observations or errors arrive from preparation as float64, which can approximately double model-memory use.

Some CO₂ and verification-games operations need float64 accumulation for adequate numerical accuracy. A process-wide float64 graph is a useful escape hatch, but it is too coarse as the only solution because it also promotes arrays and components that do not need the extra precision.

PR #606 establishes the immediate startup policy:

- a fresh OpenGHG Inversions PyMC/RHIME process defaults to float32;
- an explicit `PYTENSOR_FLAGS=floatX=...` selection is honored;
- importing OpenGHG Inversions does not mutate an already-imported PyTensor/PyMC runtime;
- graph precision is selected before imports and is not a `RhimeModelSpec` option.

## Future work

- [ ] Inventory numerically sensitive operations for CO₂ and verification-games workflows, including forward products, sector sums, likelihood reductions, covariance calculations, reconstruction, and output statistics.
- [ ] Distinguish storage dtype, PyMC tensor dtype, reduction accumulator dtype, and serialized output dtype in the numerical-ownership guidance.
- [ ] Evaluate local float64 accumulation while retaining float32 stored inputs and ordinary graph tensors.
- [ ] Add small component-owned precision options only where scientific evidence requires them; avoid a generic execution manifest or compiler policy.
- [ ] Determine which PyTensor operations support a wider accumulator without materializing full float64 copies, and benchmark alternatives where matrix products require promotion.
- [ ] Add representative numerical-error and peak-memory benchmarks for nested 6 km and CO₂-family cases.
- [ ] Record effective graph and accumulation precision in run/output provenance.
- [ ] Improve script, CLI, notebook, and HPC ergonomics for selecting process precision before PyTensor/PyMC import, including clear restart requirements.
- [ ] Define behavior for supported CPU/JAX/other sampling backends.

## Acceptance evidence

- Float32 remains the measured default for production RHIME memory use.
- Sensitive verification-games calculations meet documented accuracy tolerances.
- Mixed-precision choices are local, explicit, tested, and visible in provenance.
- Tests cover fresh-process defaults, explicit float64 selection, already-initialized runtimes, and representative accumulation accuracy.
[...]

PR #610, opened 2026-08-19
Respect explicit PyTensor process precision

brendan-m-murphy:
## Summary of changes

- retain the memory-conscious PyTensor float32 default only when OpenGHG Inversions owns a fresh process runtime
- honor explicit `PYTENSOR_FLAGS` selections such as `floatX=float64`
- leave already-imported PyTensor/PyMC runtimes unchanged, making notebook and host-process ownership predictable
- document the process-startup requirement and add isolated subprocess tests for all three cases

This separates process-wide precision policy from the RHIME ownership work in PR #606. Component-level mixed-precision accumulation, numerical evidence, provenance, and launcher ergonomics remain tracked in #607.

## Why

Float32 model inputs avoid promoting stored float32 footprints when observations or errors arrive as float64, which can substantially increase inversion memory. Some CO₂ and verification-games calculations require wider accumulation. A process-wide float64 mode remains available, while future work can promote only scientifically sensitive operations.

- [x] Tests added and passing
- [x] Documentation updated
- [x] No new requirements

## Validation

- `tests/test_pymc_config.py` — 3 passed in an isolated existing environment
- Ruff passed for the changed Python paths
- `git diff --check` passed

Related: #607, #606.

### DIFF
--- a/docs/usage/concrete_rhime_model.rst
+++ b/docs/usage/concrete_rhime_model.rst
@@ -75,6 +75,27 @@ Source/design resolution and the boundary, offset, error, and likelihood
 components remain shared. This keeps parity meaningful and avoids independent
 copies silently drifting while compiler extensions are developed.
 
+PyTensor precision at process startup
+-------------------------------------
+
+A fresh OpenGHG Inversions PyMC or RHIME process defaults PyTensor to
+``float32``. This keeps float32 footprint sensitivities from being promoted
+when observations or errors arrive from preparation as float64. Because
+PyTensor configuration is process-wide, graph precision is not a
+``RhimeModelSpec`` option and must be selected before importing PyTensor,
+PyMC, or OpenGHG Inversions.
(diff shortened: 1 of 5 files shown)


docs/usage/concrete_rhime_model.rst
```diff
@@ -75,6 +75,27 @@ Source/design resolution and the boundary, offset, error, and likelihood
 components remain shared. This keeps parity meaningful and avoids independent
 copies silently drifting while compiler extensions are developed.
 
+PyTensor precision at process startup
```
brendan-m-murphy: This is not the right location for this info. This page is meant to be a simple introduction to how the models are built. It is not for generic "running rhime" info.

brendan-m-murphy (commented):
The `configure_pytensor` function is not used anywhere. Where are we intending to use it?

PR #616, opened 2026-08-19
[OPE-82] Expose explicit RHIME scientific inputs

brendan-m-murphy:
## Summary

- replace opaque RHIME dataset plumbing in built-in model recipes with explicit labelled scientific arrays and cohesive `AggregationError` / `SigmaAlignment` values
- keep required-input declarations beside their owning flux, baseline, observation-error, and sigma components, with owner-specific validation
- make `registered_model()` / `CoordRegistry` an enforced labelled-model invariant and register fixed aggregation-error arrays as inspectable PyMC data
- materialize selected lazy aggregation payloads once at their named boundary, preserving labels and rejecting ambiguous direct-builder values
- materialize sigma/site wiring only when the selected likelihood owns it, preserving the built-in legacy compatibility exception
- expose the scientifically general `flux_sensitivity` name and a custom-likelihood seam with common arrays explicit and model-specific options in `likelihood_kwargs`
- validate, detach, and persist `likelihood_kwargs` with callable provenance in results and saved inversion output
- preserve an externally supplied cached sensitivity graph and provenance until the named flux boundary, with no retrieval and no unrelated computation
- document the durable prepared-data inventory, coordinate ownership, model-only wiring, and customization contract

## Stack

#615 (OPE-83) has merged. GitHub retargeted this PR to `devel`; this branch remains the OPE-82 follow-up on top of the OPE-83 content.

## Review follow-up

The latest review round is addressed in commit `1540bfaa`:

- the installed additive-sigma likelihood now works through the ordinary and prepared runner seams, deriving sigma alignment from labelled observations and accepting JSON-compatible component options
- direct dense `AggregationError` values enforce symmetry and positive-semidefiniteness, while resolved values avoid repeated eigendecomposition
- `registered_model(coords=...)` sanitizes PyMC coordinates and seeds the scientific registry, including xarray auxiliary coordinates
- dense covariance column labels and all `AggregationError` representation invariants are enforced
- low-rank shared Dask payloads materialize together once; public direct-builder values must already be eager
- aggregation marginal/covariance/factor/diagonal arrays are registered and preserved in `constant_data`
- redundant pairwise coordinate checks were removed in favor of mandatory component registration
- custom likelihoods no longer inherit unused built-in sigma wiring
- non-empty orphaned `likelihood_kwargs` are rejected and valid options round-trip in output metadata
- complete custom models must use the registered-model boundary

## Validation

- `ruff check <all changed Python paths>` — passed
- `git diff --check` — passed
- latest focused covariance/coordinates/likelihood/prepared-runner set — 54 passed
- copied-runner and cookiecutter integrations — 13 passed
- prepared-input serialization and xarray adapter — 116 passed
- broader `tests/models tests/test_observation_error.py tests/test_rhime.py` — 390 passed; 6 reproduced baseline/environment failures
- independent scientific/ownership review — no blockers after final direct-builder checks

Known baseline failures are three float32-assumption tests under the local PyTensor configuration, the h5netcdf `H5DSget_num_scales` reopen case, and two API-smoke cases affected by upstream OpenGHG `long_name` drift; both default and custom likelihood paths reproduce the latter.


openghg_inversions/observation_error.py
```diff
@@ -129,6 +136,120 @@ def validate_observation_error_inputs(
...
+) -> None:
+    """Validate labelled aggregation-error products against observations."""
+    arrays = (
+        (AGGREGATION_ERROR_COVARIANCE, aggregation_error.covariance),
+        (LOW_RANK_FACTOR, aggregation_error.factor),
+        (DIAGONAL_RESIDUAL_VARIANCE, aggregation_error.diagonal_variance),
+    )
+    for name, array in arrays:
+        if array is not None:
+            validate_observation_alignment(
```
brendan-m-murphy: [P1] Please validate the dense covariance's second observation axis as well as its `nmeasure` row axis. A covariance with rows `['A', 'B']` and `nmeasure_cov=['B', 'A']` is currently accepted because numeric shape and positional symmetry pass; the likelihood later converts the covariance with `.values`, silently applying columns to the wrong observations. Both axes should be normalized or exactly aligned to the observation ordering before labels are discarded.

openghg_inversions/observation_error.py
```diff
@@ -129,6 +136,120 @@ def validate_observation_error_inputs(
...
+    if output_dim not in array.dims:
+        raise ValueError(f"{owner} input {input_name!r} has no {output_dim!r} dimension.")
+    reference_coordinates = {
+        name: coordinate
+        for name, coordinate in observations.coords.items()
+        if coordinate.dims == (output_dim,)
+    }
+    for name, reference in reference_coordinates.items():
+        candidate = array.coords.get(name)
+        if candidate is None or not reference.identical(candidate):
```
brendan-m-murphy: [P2] `DataArray.identical()` makes coordinate attributes part of scientific alignment, and this loop also requires every observation-aligned auxiliary coordinate to exist on every input. That rejects independently prepared arrays with the same exact `nmeasure` index but harmless metadata differences, or a cached sensitivity that lacks an observation-only coordinate such as `qc_flag`. Please use exact indexed alignment/value equality for observation order and validate only auxiliary coordinates that the component actually consumes.

openghg_inversions/rhime/standard.py
```diff
@@ -64,10 +70,78 @@
...
+            f"`aggregation_error_mode={model_spec.aggregation_error_mode!r}`"
+        ),
+    )
+    names.extend(aggregation_names)
+    return tuple(names)
+
+
 def build_standard_rhime_model(
-    inv_inputs: xr.Dataset,
+    emissions_sensitivity: xr.DataArray,
```
brendan-m-murphy: Could we call this `flux_sensitivity` rather than `emissions_sensitivity` throughout the new explicit-input API? “Emissions” implies a positive flux, so it is less general for sensitivities representing uptake or otherwise signed fluxes. `flux_sensitivity` describes the quantity without imposing that sign convention.

brendan-m-murphy (commented):
Two coordinate-contract issues and one scientific naming request are noted inline. I also checked the clarified compatibility boundary: `run_hbmcmc.py` itself is unchanged, and I found no inferpymc/run_hbmcmc graph or numerical regression.

openghg_inversions/observation_error.py
```diff
@@ -129,6 +136,120 @@ def validate_observation_error_inputs(
...
+            raise ValueError(
+                f"{owner} input {input_name!r} has incompatible observation coordinate {name!r}."
+            )
+
+
+def validate_aggregation_error_alignment(
+    observations: xr.DataArray,
+    aggregation_error: AggregationError,
```
brendan-m-murphy: [P1] Separate from the existing column-axis thread, this accepts a manually constructed `AggregationError` whose `marginal_variance` disagrees with its covariance/factor, or whose mode and payload are inconsistent. That is now a public direct-builder input: `epsilon` and the error floor use `marginal_variance`, while `y` uses the covariance payload, so one accepted object can describe two different error models. Please provide one validated construction path or enforce these representation invariants here. A coherent-reduction path need only retain cheap shape/coordinate/consistency checks; it should not repeat the producer's expensive numerical work, while general external/prepared inputs should retain full validation.

openghg_inversions/rhime/standard.py
```diff
@@ -64,10 +70,78 @@
...
+        _STANDARD_FLUX_INPUT_NAMES,
+        owner="Standard flux component",
+    )
+    _require_component_inputs(
+        prepared,
+        OBSERVATION_ERROR_INPUT_NAMES,
+        owner="Observation-error component",
+    )
+    _require_component_inputs(
+        prepared,
```
brendan-m-murphy: [P2] Please select sigma-alignment input only when the selected model actually uses it. Both recipes currently require/materialize `site_indicator`, both result stages construct `SigmaAlignment`, and the likelihood validates it even when `no_model_error=True`. That makes a disabled component impose model-only input and frequency/alignment requirements. Keep `site_indicator` when offsets need it, and preserve the standard `preserve_legacy_likelihood=True` exception because that compatibility graph intentionally retains unused sigma; modern fixed-error standard and multisector graphs should not select sigma-only wiring.

openghg_inversions/rhime/standard.py
```diff
@@ -64,10 +70,78 @@
...
+    *,
+    owner: str,
+) -> None:
+    """Fail before materialization with the selected component named."""
+    missing = [name for name in names if name not in prepared.inv_inputs]
+    if missing:
+        raise ValueError(f"{owner} requires prepared input(s) {missing!r}.")
+
+
+def standard_model_input_names(prepared: RhimePreparedInputs, model_spec: RhimeModelSpec) -> tuple[str, ...]:
```
brendan-m-murphy: [P2] This newly public, root-exported function needs the full Google-style docstring required for public APIs, not only a summary. The same pass is needed for `multisector_model_input_names`, `materialize_pymc_inputs`, `aggregation_error_input_names`, and `validate_aggregation_error_alignment`: document Args, Returns, and applicable Raises.

brendan-m-murphy (commented):
The explicit-input direction is substantially right: the ordinary builders now expose named scientific arrays, the broad context is confined to the advanced whole-model seam, the from-spec duplication is gone, and additive-sigma plus the legacy PEFO/HBMCMC behavior are retained.

I have left three distinct items inline: resolved aggregation-error representation consistency, conditional selection of sigma-only wiring, and full docstrings for the newly public APIs.

I did not repeat the existing review's dense-covariance column-axis finding. The broader minimum-error and boundary-sensitivity preparation consolidations are now tracked separately as OPE-92 and OPE-93 rather than being added to this PR.

openghg_inversions/rhime/standard.py
```diff
@@ -122,10 +200,31 @@ def build_standard_rhime_model(
     sigma_prior = dict(DEFAULT_SIGMA_PRIOR if sigma_prior is None else sigma_prior)
     offset_prior = dict(DEFAULT_OFFSET_PRIOR if offset_prior is None else offset_prior)
 
+    validate_observation_alignment(
```
brendan-m-murphy: Please remove this recipe-level alignment block, and the equivalent block in `multisector.py`. Each consuming component registers its labelled inputs through `add_model_data` → `add_coords` → `CoordRegistry`, which already rejects conflicting shared dimension and auxiliary-coordinate values. I confirmed that, with these calls disabled, a reversed `nmeasure` order still fails during component registration. Keep validation of scientific invariants and inputs that do not cross this registration boundary—notably aggregation covariance, whose two observation axes both need explicit validation.

docs/development/rhime_model_development.rst
```diff
@@ -159,6 +159,79 @@ It is not a caller-authored manifest and must not determine execution order.
 Ordinary runners should not construct or thread a requirements object through
 the pipeline merely to satisfy a framework contract.
 
+Prepared-input inventory and ownership
+--------------------------------------
+
+``RhimePreparedInputs`` is the durable, backend-neutral labelled-data
+boundary. It may retain project-specific arrays that no current PyMC recipe
```
brendan-m-murphy, line 178: Please add the coordinate-ownership rule to this guidance in the current PR; it is small and directly relevant to the boundary OPE-82 establishes, so I would not split it into a Linear issue. Suggested substance: component registration (`add_model_data`/`add_coords`/`CoordRegistry`) owns cross-component coordinate consistency, and recipes should not repeat pairwise alignment checks for data that components register. Explicit validation remains appropriate for scientific invariants and for arrays that are not registered with PyMC, such as aggregation covariance.

brendan-m-murphy (commented):
Follow-up on coordinate ownership: component data registration already owns cross-component coordinate consistency. The recipe-level checks should be removed, and this ownership rule should be stated in the model-development guidance as part of this PR.

openghg_inversions/rhime/standard.py
```diff
@@ -191,12 +298,14 @@ def build_standard_rhime_model(
                 pollution_events_from_obs=pollution_events_from_obs,
                 no_model_error=no_model_error,
                 retain_unused_sigma=preserve_legacy_likelihood,
-                aggregation_error_mode=aggregation_error_mode,
                 output_dim="nmeasure",
             )
         else:
             likelihood = likelihood_builder(
```
brendan-m-murphy, line 319: This call makes the custom-builder contract effectively identical to the default likelihood's full signature, so every custom likelihood must accept sigma and pollution-event policy even when it does not use them. Please keep the common scientific inputs explicit (`observations` through `pollution_event_baseline`, plus `output_dim`) and accept a `likelihood_kwargs` mapping for the remainder, expanding it here with `**likelihood_kwargs`. The built-in recipe can assemble its mapping explicitly from `sigma_alignment`, `sigma_prior`, `power`, `pollution_events_from_obs`, and `no_model_error`; the default likelihood should retain those as required named parameters, so Python validates them naturally. This avoids both making all default-specific arguments optional and hiding the common scientific data behind an opaque object.

brendan-m-murphy (commented):
The likelihood extension seam should separate the common scientific inputs from options specific to the built-in pollution-event likelihood.

openghg_inversions/rhime/standard.py
```diff
@@ -488,14 +590,19 @@ def run_rhime(
...
+            preserve_legacy_likelihood=preserve_legacy_likelihood,
+        ),
     )
     build_and_sample_start = timer_start()
     model_build_result = build_standard_rhime_model_result(
         prepared=prepared,
         model_inputs=model_inputs,
         run_spec=run_spec,
         likelihood_builder=likelihood_builder,
+        likelihood_kwargs=likelihood_kwargs,
```
brendan-m-murphy, line 619: [P1] `likelihood_kwargs` reaches model construction here but is not forwarded to `make_standard_rhime_result` or persisted with the likelihood-builder provenance (and multisector behaves the same). These values can change the scientific model: runs with the same Student-t callable but `degrees_of_freedom=4` and `=30` currently produce identical result and saved builder metadata. Please validate/copy a JSON-compatible mapping, include it in `RhimeResult.output_metadata` and persisted builder metadata, and cover the output round-trip.

openghg_inversions/rhime/standard.py
```diff
@@ -405,6 +504,7 @@ def run_rhime(
     config_file: str | Path | None = None,
     merged_data: RhimeMergedData | None = None,
     likelihood_builder: RhimeLikelihoodBuilder | None = None,
+    likelihood_kwargs: Mapping[str, Any] | None = None,
```
brendan-m-murphy, line 520: [P2] Reject non-empty `likelihood_kwargs` unless a `likelihood_builder` is active. At present `run_rhime(likelihood_kwargs={"degrees_of_freedom": 7})` completes with the default Gaussian and silently discards the option, so an omitted or misspelled builder can change the intended scientific model without an error. Apply the same ownership check in the multisector runner and reject these options on the prepared complete-`model_builder` route as well.

openghg_inversions/rhime/standard.py
```diff
@@ -304,26 +361,68 @@ def build_standard_rhime_model_result(
...
+        )
         result = validated_custom_model_build(model_builder, context=builder_context)
         validate_model_build_result(result, context=builder_context)
     else:
-        model = _build_standard_rhime_model_from_spec(
+        model_spec = run_spec.model
+        if len(model_spec.sectors) != 1:
+            raise ValueError("Standard RHIME model specs must include exactly one sector.")
+        sector = model_spec.sectors[0]
+        sigma_alignment = (
```
brendan-m-murphy, line 383: [P2] The custom branch no longer receives `SigmaAlignment`, but input selection and this result stage still construct it whenever the model spec has the default `no_model_error=False`. Consequently the documented fixed-error custom likelihood still requires/materializes `site_indicator` and can fail before invocation, while a custom additive-sigma likelihood cannot use the alignment that RHIME constructed. Please make the contract consistent: either gate built-in sigma input selection/construction on `likelihood_builder is None` (retaining the legacy exception), or keep preparation-derived runtime alignment available at the likelihood seam separately from caller-supplied `likelihood_kwargs`. An ordinary caller cannot construct it before RHIME filters and prepares observations. The multisector path has the same issue.

openghg_inversions/observation_error.py
```diff
@@ -129,6 +136,255 @@ def validate_observation_error_inputs(
...
+                AGGREGATION_ERROR_COVARIANCE,
+                covariance,
+                owner=f"{owner} input",
+            )
+        )
+    elif aggregation_error.mode == "low_rank":
+        factor = aggregation_error.factor
+        diagonal_variance = aggregation_error.diagonal_variance
+        assert factor is not None and diagonal_variance is not None
+        factor_values = _numeric_finite(LOW_RANK_FACTOR, factor, owner=f"{owner} input")
```
brendan-m-murphy: [P2] This purportedly cheap consumer check materializes the low-rank factor and residual separately through `.values`. With both derived from one Dask source, this validation alone executes the shared source twice, leaves both fields lazy, and they are converted again by the likelihood. The ordinary runner masks this by pre-materializing, but the public `resolve_aggregation_error(lazy_dataset)`/direct-component path reproduces it. Check structural dimensions before payload access and either require an eager validated `AggregationError` here or materialize the related payloads together once and pass the eager representation onward.

brendan-m-murphy (commented):
The recent update addresses the earlier coordinate, naming, conditional-sigma, documentation, and likelihood-signature comments. Four follow-ups remain around likelihood option ownership/provenance and the eager numerical boundary; they are noted inline. Exact-head observation-error, run_hbmcmc, and inferpymc selections passed, and changed-path Ruff passed.

brendan-m-murphy (commented):
Focused follow-up on making coordinate/data registration the actual model invariant. These comments preserve the scientific covariance checks while consolidating repeated observation-axis alignment at the boundary that owns it.

docs/development/rhime_model_development.rst
```diff
@@ -159,6 +159,87 @@ It is not a caller-authored manifest and must not determine execution order.
 Ordinary runners should not construct or thread a requirements object through
 the pipeline merely to satisfy a framework contract.
 
+Component registration owns cross-component coordinate consistency. Data
```
brendan-m-murphy, line 163: [P2] Please make this an enforced model invariant rather than only a convention. There are only three in-tree `pm.Model()` construction sites, and each immediately attaches a `CoordRegistry`. A small factory such as `registered_model() -> pm.Model` would let those sites use `with registered_model() as model:`, after which `add_coords()` can fail clearly if a direct/custom model has no registry; the complete-model boundary should enforce the same invariant. This preserves ordinary callable components while making the rule below true for every model. The earlier statement that registries are “optional conveniences” should be updated accordingly.

openghg_inversions/models/pollution_event.py
```diff
@@ -93,25 +97,54 @@ def build_pollution_event_error(
...
-    observation_error = add_model_data(data["mf_error"].transpose(output_dim), "error")
-    minimum_error = add_model_data(data["min_error"].transpose(output_dim), "min_error")
+    needs_sigma = not no_model_error or retain_unused_sigma
+    if needs_sigma:
+        if sigma_alignment is None:
+            raise ValueError(
+                "Pollution-event likelihood requires `sigma_alignment` when "
+                "model error is enabled or unused sigma is retained."
+            )
+        validate_observation_alignment(
```
brendan-m-murphy: [P2] Once registry attachment is mandatory, please remove these pairwise coordinate checks and their additive-sigma equivalents. `observations`, `observation_error`, and `minimum_error` are registered through `add_model_data`, while `SigmaAlignment` indices are registered by `add_sigma_component`; registration should own their observation-axis compatibility. `validate_observation_error_arrays` should retain its numerical/scientific checks, but not repeat coordinate alignment. This is the redundant likelihood-chain alignment that the new ownership rule is meant to eliminate.

openghg_inversions/models/pollution_event.py
```diff
@@ -127,15 +160,15 @@ def build_pollution_event_error(
         exponent = (
             parse_prior("power", dict(power)) if isinstance(power, Mapping) else power
         )
-        raw_independent_variance = observation_error**2 + pt.pow(
+        raw_independent_variance = reported_error**2 + pt.pow(
             pollution_event * sigma,
             exponent,
         )
         aggregation_marginal_variance = pt.as_tensor_variable(
```
brendan-m-murphy: [P1] “Any data used by the model must be registered” also needs to cover aggregation inputs. The marginal variance here, and covariance/factor/diagonal arrays downstream, are embedded from `.values` as anonymous tensor constants, so they bypass both `CoordRegistry` and `InferenceData.constant_data`. Please register the labelled aggregation data through `add_model_data` and use the registered representation in the graph. The registry can then own observation-row alignment; retain the explicit dense `nmeasure_cov`-column equality check because that second axis has a distinct dimension name, along with the covariance invariants owned by `AggregationError`.

brendan-m-murphy: I checked this against the repository’s pinned PyMC 5.26.1. PyMC handles the matrix directly as registered data: `pm.Data("aggregation_error_covariance", covariance, dims=("nmeasure", "nmeasure_cov"))`, while `pm.MvNormal(..., dims="nmeasure")` labels the distribution output. Both covariance-axis coordinates are preserved in `InferenceData.constant_data`, so registration itself should not be deferred. The two distinct dimension names are needed for xarray/ArviZ; PyMC does not infer that they denote the same ordered observation set. The explicit equality/order check therefore remains a scientific invariant owned by `AggregationError`. The broader two-axis identity policy is already tracked in [OPE-13](https://linear.app/openghg-inversions/issue/OPE-13/complete-covariance-alignment-and-lrpd-positivity-correction-573). See also the [PyMC `Data` dimensions/coordinates API](https://www.pymc.io/projects/docs/en/v5.24.0/api/generated/pymc.Data.html).

openghg_inversions/rhime/likelihoods.py
```diff
@@ -11,23 +11,24 @@
 from openghg_inversions.models.additive_sigma import (
     add_additive_sigma_gaussian_likelihood,
 )
-from openghg_inversions.observation_error import AggregationErrorMode
+from openghg_inversions.observation_error import AggregationError
 from openghg_inversions.sigma import SigmaAlignment
 
 
 def additive_sigma_likelihood_builder(
```
brendan-m-murphy, line 21: [P1] Keep the installed additive-sigma likelihood usable through the ordinary runner seam. `RhimeLikelihoodBuilder` and the recipe call site supply only the common arrays/means/output dimension, but this callable additionally requires `sigma_alignment`, `sigma_prior`, `power`, `pollution_events_from_obs`, and `no_model_error`; `run_rhime(likelihood_builder=additive_sigma_likelihood_builder)` therefore raises `TypeError`. The tests invoke this adapter directly, not through the runner, and the docs now explicitly demote it to copied recipes. That conflicts with the agreed requirement to retain a first-class installed additive-sigma likelihood selectable through the ordinary seam. Please either make this callable satisfy the actual protocol and own/derive its sigma preparation from labelled observations plus JSON-compatible options, or add another explicit supported built-in selection—but do not leave the installed adapter unusable by `run_rhime`.

openghg_inversions/observation_error.py
```diff
@@ -129,6 +155,264 @@ def validate_observation_error_inputs(
...
+            raise ValueError(
+                f"{owner} input {AGGREGATION_ERROR_COVARIANCE!r} has "
+                f"incompatible observation coordinate {covariance_dim!r}."
+            )
+        _require_eager_payload(
+            covariance,
+            AGGREGATION_ERROR_COVARIANCE,
+            owner=owner,
+        )
+        expected_marginal = np.diag(
```
brendan-m-murphy: [P1] Direct dense `AggregationError` values still bypass the covariance’s symmetry and positive-semidefinite invariants. For example, a correctly labelled `[[1, 2], [2, 1]]` covariance with `marginal_variance=[1, 1]` passes this consumer check even though it is indefinite. `resolve_aggregation_error()` rejects it, but the public concrete builders accept directly constructed `AggregationError` values and this follow-up claims all representation invariants are enforced. Please put these invariants in `AggregationError` construction/a factory, or perform the complete dense check at this direct-public boundary, with direct-value regression tests.

openghg_inversions/models/coords.py
```diff
@@ -218,6 +218,21 @@ def get_coord_registry(model: pm.Model) -> CoordRegistry | None:
...
+    """Construct a PyMC model with RHIME coordinate tracking attached.
+
+    Args:
+        *args: Positional arguments forwarded to :class:`pymc.Model`.
+        **kwargs: Keyword arguments forwarded to :class:`pymc.Model`.
+
+    Returns:
+        A new PyMC model with an empty :class:`CoordRegistry` attached.
+    """
+    model = pm.Model(*args, **kwargs)
```
brendan-m-murphy: [P2] `registered_model(coords=...)` forwards coordinates to PyMC and then attaches an empty registry, so those constructor coordinates are absent from `CoordRegistry.original_coords`; complete-model validation currently checks only that a registry object exists. A custom model can therefore pass the new invariant while rich coordinates supplied at construction are unavailable to restoration. Please either seed the registry from the scientific `coords` while passing sanitized coordinates to PyMC, or forbid constructor coordinates and require `add_coords()`. Validation should distinguish an attached-but-empty registry from coverage of the labelled model dimensions that need restoration. The output-side audit is now tracked explicitly in OPE-48 if the broader coverage check is deferred.

brendan-m-murphy (commented):
Quick final pass found two remaining blockers against the agreed OPE-82/additive-likelihood criteria, plus one coordinate-registry completeness gap now tracked in OPE-48.

brendan-m-murphy:
Downstream integration audit at exact head `1540bfaa96edea2d16c9a633e08b5fe8bd3b6420`: I found no remaining substantive blocker for the CO2/CO2–O2 stack. The latest commit enforces AggregationError representation/symmetry/PSD/second-axis invariants, seeds and requires the coordinate registry, registers aggregation payloads, validates and persists `likelihood_kwargs`, rejects orphan options, keeps custom likelihoods free of built-in sigma wiring, and exposes the installed additive-sigma builder through the common labelled-input protocol. Exact-head Actions run `32311040484` is green.

This head is the base of OGI #617 and #618; the completed verification-games two-site/one-week deterministic and sampled acceptance found no downstream violation of these invariants. The remaining unresolved review-thread state appears administrative (the requested changes are present in `1540bfaa`), so I have left thread resolution to the reviewer.

openghg_inversions/observation_error.py
```diff
@@ -40,14 +42,56 @@ class AggregationError:
...
+    scale = max(float(np.max(np.abs(values))), 1.0)
+    tolerance = 1e-10 * scale
+    if not np.allclose(values, values.T, rtol=1e-10, atol=tolerance):
+        raise ValueError(f"{owner} must be symmetric.")
+    if float(np.linalg.eigvalsh(values).min()) < -tolerance:
+        raise ValueError(f"{owner} must be positive semidefinite.")
 
 
-def _numeric_finite(name: str, array: xr.DataArray) -> np.ndarray:
+def _numeric_finite(
```
brendan-m-murphy: This is almost certainly exists in another module.

PR #619, opened 2026-08-20
OPE-92 Consolidate minimum-error preparation

brendan-m-murphy:
* **Summary of changes**

Consolidate scalar, per-site, residual, and percentile minimum-error preparation in a typed `MinimumError` value.

- align per-site values through the labelled observation `site` coordinate
- validate finite, non-negative values and record method/site/unit provenance
- preserve resolved `xr.DataArray` inputs for RHIME/PyMC and analytic consumers
- remove superseded minimum-error alignment helpers

Linear: [OPE-92](https://linear.app/openghg-inversions/issue/OPE-92/consolidate-minimum-error-preparation-in-a-typed-minimumerror-value)

- [ ] Closes #xxxx — N/A; tracked in Linear as OPE-92
- [x] Tests added and passing
- [ ] Documentation and tutorials updated/added — not required; public API docstrings added
- [x] Added a Towncrier news fragment in `newsfragments/`
- [ ] Added any new requirements to `requirements.txt` — no new requirements

Checks:
- `uv run pytest tests/test_inversion_inputs.py -q`: 22 passed, 1 deselected
- targeted RHIME/xarray/fixed-basis minimum-error tests: 7 passed, 363 deselected
- `uv run ruff check openghg_inversions/model_error.py openghg_inversions/inversion_inputs.py tests/test_inversion_inputs.py`
- `git diff --check`
- Slurm Mypy job 18645378 reached the existing repository baseline: 197 errors in 56 files; no errors in the new `model_error.py`.

### DIFF
--- a/openghg_inversions/model_error.py
+++ b/openghg_inversions/model_error.py
@@ -9,9 +9,9 @@ import numpy as np
 import xarray as xr
 
 
-@dataclass(frozen=True, init=False)
+@dataclass(frozen=True, eq=False, slots=True)
 class MinimumError:
-    """Validated, observation-aligned minimum model error.
+    """Observation-aligned minimum model error.
 
     Attributes:
         values: Minimum error labelled by observation.
@@ -24,26 +24,6 @@ class MinimumError:
     method: str
     by_site: bool
     sites: tuple[str, ...]
-
-    def __init__(self) -> None:
-        """Reject construction outside :meth:`prepare`."""
-        raise TypeError("Use MinimumError.prepare() to construct a validated value.")
-
-    @classmethod
-    def _from_validated(
-        cls,
-        values: xr.DataArray,
-        method: str,
-        by_site: bool,
-        sites: tuple[str, ...],
-    ) -> "MinimumError":
-        """Construct an instance after :meth:`prepare` validates its fields."""
-        result = object.__new__(cls)
-        object.__setattr__(result, "values", values)
-        object.__setattr__(result, "method", method)
-        object.__setattr__(result, "by_site", by_site)
-        object.__setattr__(result, "sites", sites)
-        return result
 
     @classmethod
     def prepare(
@@ -75,14 +55,8 @@ class MinimumError:
         if not isinstance(by_site, bool):
             raise ValueError(f"Minimum-error 'by_site' must be a boolean, got {type(by_site).__name__}.")
 
-        site_coord = observations.coords.get("site")
-        sites = (
-            tuple(dict.fromkeys(str(value) for value in site_coord.values))
-            if site_coord is not None
-            else ()
-        )
-        selected = {site: fp_data[site] for site in sites if site in fp_data}
-        missing_data = [site for site in sites if site not in selected]
+        site_coord: xr.DataArray | None = None
+        sites: tuple[str, ...] = ()
         method: str
         varies_by_site = False
 
@@ -93,6 +67,10 @@ class MinimumError:
             source = value
             method = "scalar"
         elif isinstance(value, Mapping):
+            site_coord = observations.coords.get("site")
+            if site_coord is None:
+                raise ValueError("Per-site min_error values require a labelled site coordinate.")
+            sites = tuple(dict.fromkeys(str(site) for site in site_coord.values))
             missing = [site for site in sites if site not in value]
             if missing:
                 raise ValueError(f"min_error mapping is missing values for site(s): {missing}")
@@ -100,6 +78,12 @@ class MinimumError:
             method = "per_site"
             varies_by_site = True
         elif value in ("residual", "percentile"):
+            site_coord = observations.coords.get("site")
+            if site_coord is None:
+                raise ValueError("Calculated min_error values require a labelled site coordinate.")
+            sites = tuple(dict.fromkeys(str(site) for site in site_coord.values))
+            selected = {site: fp_data[site] for site in sites if site in fp_data}
+            missing_data = [site for site in sites if site not in selected]
             if missing_data:
                 raise ValueError(f"Minimum-error calculation is missing fp_data for site(s): {missing_data}")
             method = value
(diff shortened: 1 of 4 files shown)


brendan-m-murphy (commented):
The consolidation is an improvement: `MinimumError` gives a real scientific/model-data concept one clear name and puts the supported preparation paths together. I think two issues should be addressed before merge:

1. **Preserve floating-point minimum errors** — `openghg_inversions/model_error.py:113`

   `xr.full_like(observations.mf, ...)` inherits the observation dtype. Integer-valued `mf` is accepted by this preparation path, so `min_error=0.5` becomes an integer array of zeros and silently changes the scientific model. Please construct the result with an explicit floating dtype and add a regression test using integer observations.

2. **Make `MinimumError` genuinely prepare-only** — `openghg_inversions/model_error.py:12`

   The generated public dataclass constructor bypasses the invariants enforced by `.prepare()`; callers can directly create negative/non-finite values, arbitrary dimensions, or inconsistent provenance. Please prevent direct public initialization and use an internal construction path after validation.

The desired API is prepare-only, matching the revised boundary-alignment direction. Keep the assignment into the dataset explicit in `make_inv_inputs` for this PR. Moving model-owned construction out of inversion-input preparation is separate follow-up [OPE-105](https://linear.app/openghg-inversions/issue/OPE-105/move-model-owned-alignment-construction-out-of-inversion-input).

The earlier NetCDF boolean-attribute issue is fixed at the current head.

brendan-m-murphy:
Addressed both review points in `862d3e14`:

- Scalar minimum errors now use an explicit floating dtype, so integer-valued observations cannot truncate fractional floors. The regression test uses integer Dask observations and verifies `0.5` is preserved.
- Direct `MinimumError()` construction is rejected; `MinimumError.prepare()` now returns an internally constructed, validated value.

Validation:
- Focused tests: 11 passed
- Ruff: passed
- `git diff --check`: passed
- Focused Python 3.10/current OpenGHG tox on Slurm: job `18646804` passed

openghg_inversions/model_error.py
```diff
@@ -1,12 +1,146 @@
...
+    method: str
+    by_site: bool
+    sites: tuple[str, ...]
+
+    def __init__(self) -> None:
+        """Reject construction outside :meth:`prepare`."""
+        raise TypeError("Use MinimumError.prepare() to construct a validated value.")
+
+    @classmethod
+    def _from_validated(
```
brendan-m-murphy: I think we have mostly rejected this pattern. Check the dev docs.

openghg_inversions/model_error.py
```diff
@@ -1,12 +1,146 @@
...
+                are invalid.
+        """
+        if "mf" not in observations:
+            raise ValueError("Minimum-error preparation requires an 'mf' observation array.")
+        if not isinstance(by_site, bool):
+            raise ValueError(f"Minimum-error 'by_site' must be a boolean, got {type(by_site).__name__}.")
+
+        site_coord = observations.coords.get("site")
+        sites = (
+            tuple(dict.fromkeys(str(value) for value in site_coord.values))
```
brendan-m-murphy: **[P2] Do not execute site metadata for input forms that do not use it.** This reads `site_coord.values` before the input-form branch, so even a scalar minimum error materializes a potentially Dask-backed auxiliary `site` coordinate solely to record unused site provenance. That conflicts with the guidance to treat auxiliary coordinates as potentially lazy and to inspect them only at a named boundary. Determine the input form first and derive site labels only for mapping/residual/percentile paths that require them; if per-site alignment must materialize an indexer, make that boundary explicit.

openghg_inversions/model_error.py
```diff
@@ -1,12 +1,146 @@
...
+                coords={"site": np.asarray(sites)},
+                dims="site",
+            )
+            data = per_site.sel(site=site_coord)
+        else:
+            if source.size != 1:
+                raise ValueError("A non-site minimum error must contain exactly one value.")
+            data = xr.full_like(observations.mf, float(source.reshape(-1)[0]), dtype=float)
+
+        data = data.rename("min_error").assign_attrs(
```
brendan-m-murphy: **[P2] Replace rather than merge observation attributes.** `xr.full_like(observations.mf, ...)` preserves the `mf` attributes, and `assign_attrs` merges this provenance into them. Scalar minimum error can therefore inherit scientifically incorrect metadata such as the observation `standard_name` or calibration attributes, while the per-site path does not. Construct a fresh minimum-error attribute mapping (or clear the inherited attrs first) and add a regression test with non-unit observation metadata.

brendan-m-murphy (commented):
Re-review of the current head against the RHIME and xarray guidance.

The earlier floating-dtype and NetCDF-provenance problems are fixed. I agree with the existing inline comment that the new `init=False` / `object.__setattr__` construction is one of the dataclass antipatterns explicitly called out in the numerical ownership guidance, so I have not duplicated that thread.

Two additional issues remain below. Both come from doing more implicit work than the value actually needs. OPE-84 may eventually help with declarative structural schemas, but it would not address these execution-boundary or scientific-metadata problems.

openghg_inversions/model_error.py
brendan-m-murphy: Concretely, I would undo the construction-enforcement machinery introduced in response to the earlier review:

```python
@dataclass(frozen=True, eq=False, slots=True)
class MinimumError:
    values: xr.DataArray
    method: str
    by_site: bool
    sites: tuple[str, ...]
```

Keep `.prepare(...)` as the normal supported configuration/calculation path, and end it with:

```python
return cls(data, method, varies_by_site, sites)
```

Delete `init=False`, the rejecting `__init__`, `_from_validated`, all `object.__new__` / `object.__setattr__` calls, and the test requiring direct construction to fail. Also remove “Validated” from the class-level promise unless every construction path actually guarantees that.

This follows the documented policy: package-created values are trusted; callers who directly construct custom values are responsible for them. The finite/non-negative checks in `.prepare` remain because they define scientific policy. If preventing direct construction is genuinely required later, this should be an ordinary class with an honest initializer—not a dataclass whose generated model is disabled.

openghg_inversions/model_error.py
brendan-m-murphy: Fixed in `9014a963`: minimum-error attrs are now replaced with a fresh provenance mapping rather than merged with `mf` attrs. The regression test includes `standard_name` and calibration metadata and verifies neither leaks into `min_error`.

openghg_inversions/model_error.py
brendan-m-murphy: Fixed in `9014a963`: scalar inputs now branch before reading the auxiliary `site` coordinate. Site labels are materialized only for mapping or calculated methods that require per-site selection/alignment; the regression test verifies scalar preparation executes no Dask tasks.

openghg_inversions/model_error.py
brendan-m-murphy: Fixed in `9014a963`: restored the ordinary `@dataclass(frozen=True, eq=False, slots=True)` field model, removed the custom construction machinery and prepare-only test, and kept scientific-policy validation in `.prepare()`.

PR #620, opened 2026-08-20
Consolidate linear PyMC components

brendan-m-murphy:
* **Summary of changes**

Consolidates the plain and state-aware linear PyMC paths around one `add_linear_component(...)` API.

- Introduces `prepare_linear_sensitivity` as the owning boundary for exact-zero column removal while retaining the full labelled scientific-state mapping.
- Preserves the established `add_linear_component(...) -> LinearComponentResult` contract; the unified result exposes data, latent, state, output, and resolved activity.
- Adds `apply_linear_sensitivity` for shared or correlated states that need another prepared forward operator without another prior.
- Migrates standard, multisector, legacy, and Ramsden model construction to the consolidated path.
- Namespaces pruned Ramsden backend axes per sector and channel while preserving the shared scientific state dimension.
- Updates focused tests, documentation, and the OPE-94 Towncrier fragment.

Linear: [OPE-94](https://linear.app/openghg-inversions/issue/OPE-94/consolidate-plain-and-state-aware-linear-pymc-components)

- [ ] Closes #xxxx (No corresponding GitHub issue; tracked in Linear as OPE-94)
- [x] Tests added and passing
- [x] No new requirements

**Validation**

- `uv run pytest tests/models/test_components.py tests/models/test_state_activity.py tests/experimental/test_ramsden2022.py tests/test_inferpymc.py tests/test_rhime.py -q -k 'state_activity or zero_columns or multisector_model or multisector_model_namespaces or inferpymc or ramsden'` — 89 passed
- Ruff on all changed Python files
- `git diff --check`
- SLURM tox job `18651421`: `py310-openghgCur`, `py310-openghgPrev`, and `py310-openghgDev` all passed

### DIFF
--- a/docs/usage/concrete_rhime_model.rst
+++ b/docs/usage/concrete_rhime_model.rst
@@ -165,13 +165,13 @@ The important default model-data and deterministic names are:
      - Role
      - Canonical input
    * - ``hx``
-     - Flux design data
+     - Flux sensitivity data
      - ``H``
    * - ``mu``
      - Flux contribution
      - ``hx @ x``
    * - ``hbc``
-     - Boundary design data
+     - Boundary sensitivity data
      - ``H_bc``
    * - ``mu_bc``
      - Boundary contribution
@@ -204,7 +204,7 @@ helpers:
 
    from openghg_inversions.models import (
        add_linear_component,
-       prepare_linear_design,
+       prepare_linear_sensitivity,
        registered_model,
    )
    from openghg_inversions.models.likelihoods import add_gaussian_observation_likelihood
@@ -231,12 +231,12 @@ helpers:
        frequency=None,
        per_site=True,
    )
-   flux_design = prepare_linear_design(inv_inputs["H"])
-   boundary_design = prepare_linear_design(inv_inputs["H_bc"])
+   flux_sensitivity = prepare_linear_sensitivity(inv_inputs["H"])
+   boundary_sensitivity = prepare_linear_sensitivity(inv_inputs["H_bc"])
 
    with registered_model() as model:
        flux = add_linear_component(
-           flux_design,
+           flux_sensitivity,
            data_name="hx",
            prior_args=x_prior,
            var_name="x",
@@ -244,7 +244,7 @@ helpers:
            output_dim="nmeasure",
        )
        boundary = add_linear_component(
-           boundary_design,
+           boundary_sensitivity,
            data_name="hbc",
            prior_args=bc_prior,
            var_name="bc",
@@ -282,12 +282,12 @@ model developer needs to change graph construction directly. It does not
 automatically participate in the complete ``run_rhime`` output pipeline; that
 pipeline still selects one of the built-in model builders.
 
-``prepare_linear_design`` is the single owning boundary for exact-zero column
-inspection. It removes those columns from the backend design while retaining
+``prepare_linear_sensitivity`` is the single owning boundary for exact-zero column
+inspection. It removes those columns from the backend sensitivity while retaining
 their full labelled-state mapping; ``state_activity=None`` therefore samples
 every retained state. An explicit ``StateActivity`` fixes or groups scientific
 states but cannot restore a structurally absent column. For shared or
-correlated states, use ``apply_linear_design`` to apply another prepared
+correlated states, use ``apply_linear_sensitivity`` to apply another prepared
 forward operator without constructing a second prior.
 
 Multisector model
@@ -303,7 +303,7 @@ state and forward contribution:
    \mu &= \sum_s \mu_s.
 
 If the normalized PyMC suffix for sector ``s`` is ``ff``, its variables are
-``x_ff`` and ``mu_ff`` and its design data is ``hx_ff``. Source values select
+``x_ff`` and ``mu_ff`` and its sensitivity data is ``hx_ff``. Source values select
 the corresponding ``H`` slices; sector names provide model identities. They
 are not required to be the same strings.
 
--- a/docs/usage/concrete_rhime_model.rst
+++ b/docs/usage/concrete_rhime_model.rst
@@ -204,6 +204,7 @@ helpers:
 
    from openghg_inversions.models import (
        add_linear_component,
+       prepare_linear_design,
        registered_model,
    )
    from openghg_inversions.models.likelihoods import add_gaussian_observation_likelihood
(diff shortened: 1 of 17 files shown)


brendan-m-murphy (commented):
This is the better design direction for OPE-94 once two remaining issues are fixed:

1. **Namespace each sector's retained design dimension** — `openghg_inversions/rhime/multisector.py:261`

   After exact-zero column removal, multiple shared-basis sectors can each register `region_retained`. If different columns are removed for different sources, those coordinates are incompatible and model construction fails. Please namespace each sector's retained backend dimension and add a two-source test in which the removed columns differ.

2. **Perform finite validation and exact-zero detection in one Dask execution** — `openghg_inversions/models/state_activity.py:219-222`

   The two independent `.compute()` calls traverse the same Dask-backed design twice during preparation. Please compute both reductions together and add an execution-counting test showing one preparation inspection while the retained payload remains lazy until the PyMC materialization boundary.

Keeping `LinearComponentResult` is appropriate in this PR because removing it would be an implicit API change; [OPE-104](https://linear.app/openghg-inversions/issue/OPE-104/reconsider-and-standardize-model-component-return-contracts) owns the deliberate return-contract decision.

One clarification for that later decision: the multigas model does not share an observation-aligned `H`. The gases can share a latent state and underlying raw footprint, but each gas has a separately observation-aligned prepared design. The readable recipe is therefore one shared state followed by applying each gas-specific design, rather than relying on `result.state` as an incidental handoff.

docs/usage/concrete_rhime_model.rst
```diff
@@ -230,18 +231,20 @@ helpers:
        frequency=None,
        per_site=True,
    )
+   flux_design = prepare_linear_design(inv_inputs["H"])
+   boundary_design = prepare_linear_design(inv_inputs["H_bc"])
 
    with registered_model() as model:
        flux = add_linear_component(
-           inv_inputs["H"],
+           flux_design,
```
brendan-m-murphy: We don't use the term "design matrix" we use "sensitivity matrix". Update any relevant names to use "sensitivity" instead of "design".

openghg_inversions/experimental/ramsden2022/model.py
```diff
@@ -762,58 +782,81 @@ def build_ramsden_model(
...
+        (
+            sector,
+            designs,
+            prepare_linear_design(
+                _rename_observation_axis(designs.primary, primary_suffix),
+                output_dim=primary_dim,
+            ),
+            None
+            if designs.tracer is None
+            else prepare_linear_design(
```
brendan-m-murphy: **[P1] Namespace each Ramsden retained axis per forward term.** Primary, tracer, and different sectors can remove different exact-zero columns, but every pruned operator currently receives the same backend dimension such as `region_retained`. The next `apply_linear_design` then fails when the retained labels differ. I reproduced this with primary fossil `west` zero and tracer fossil `east` zero: `ValueError: Conflicting coord registration for 'region_retained': values differ.` Namespace only the retained backend axis by sector and channel, keep the full scientific state dimension shared, and add a regression with different zero patterns.

tests/models/test_state_activity.py
```diff
@@ -126,6 +127,60 @@ def test_detect_zero_sensitivity_accepts_a_named_output_dimension() -> None:
...
+        with registered_model():
+            add_linear_component(
+                prepared,
+                data_name="hx",
+                prior_args={"pdf": "normal", "mu": 1.0, "sigma": 0.2},
+                var_name="x",
+                output_name="mu",
+            )
+
+        assert len(executions) > 1
```
brendan-m-murphy: **[P3] Prove exactly one PyMC materialization.** `> 1` only proves that some later Dask execution occurred; it would also pass if registration executed the graph several times. Assert `len(executions) == 2` so the test proves one preparation inspection followed by one PyMC materialization. I measured the current implementation at exactly `1 -> 2`, so this should be a test-only tightening.

brendan-m-murphy (commented):
Follow-up review: the earlier multisector retained-axis conflict and duplicated Dask inspection are fixed.

One documentation correction is also needed in `docs/usage/rhime.rst`: the legacy `inferpymc` path no longer samples every flux state when `H` has exact-zero columns. Observation-space predictions are unchanged because those columns contribute zero, but the full posterior `x` is not identical: formerly unidentified entries change from prior draws to fixed reconstructed values, which may also affect derived flux products. Please document that distinction. I am treating the pruning itself as the intentional OPE-94 change, not requesting legacy code vendoring here.

The remaining actionable code and acceptance-test findings are inline.

openghg_inversions/experimental/ramsden2022/model.py
```diff
@@ -628,31 +636,70 @@ def _sum_terms(terms: list[TensorVariable], *, name: str, dim: str) -> TensorVar
...
+    suffix: str,
+) -> PreparedLinearSensitivity:
+    """Namespace only a structurally retained backend state axis."""
+    retained_dim = next(
+        str(dim) for dim in prepared.sensitivity.dims if dim != prepared.output_dim
+    )
+    if retained_dim == prepared.state_dim:
+        return prepared
+    return PreparedLinearSensitivity(
+        sensitivity=prepared.sensitivity.rename({retained_dim: f"{retained_dim}_{suffix}"}),
```
brendan-m-murphy: **[P1] Namespace retained auxiliary coordinates with their axis.** `prepare_linear_sensitivity` renames every auxiliary coordinate spanning a pruned state axis to names such as `basis_group_retained`, but this helper then namespaces only the retained dimension. If primary and tracer retain different columns while carrying `basis_group`, the second component registers the same auxiliary-coordinate name against a different dimension and model construction fails with `ValueError: Conflicting coord registration for 'basis_group_retained': shape mismatch.` Please namespace all auxiliary coordinates spanning `retained_dim` with the same sector/channel suffix. Reusing `_namespace_sector_state_coords(..., namespace_state_dim=True)` is the smallest existing solution. Add `basis_group` to the different-zero-mask regression so it exercises the scientific metadata we need to preserve. The general ownership rule is tracked separately in #624.

brendan-m-murphy (commented):
Final pass: the earlier naming, inferpymc documentation, Dask execution-count, and simple retained-axis findings are fixed, and CI is green. One realistic auxiliary-coordinate case remains inline. The broader coordinate-namespace design is now tracked in #624; this PR only needs the local propagation fix.

openghg_inversions/experimental/ramsden2022/model.py
brendan-m-murphy: Fixed in `14a0d9b5`. `_namespace_retained_sensitivity` now reuses `_namespace_sector_state_coords(..., namespace_state_dim=True)`, so the retained dimension and every auxiliary coordinate spanning it receive the same sector/channel suffix while the full scientific state remains shared. The differing-zero-mask Ramsden regression now carries `basis_group` and asserts the independently retained methane/ethane metadata. Validation: Ramsden suite `16 passed` locally, Ruff clean, and the previously failing OpenGHG-development tox test passed in targeted SLURM job `18652553`.

PR #621, opened 2026-08-20
OPE-93: Consolidate boundary sensitivity preparation

brendan-m-murphy:
## Summary

- add a typed `BoundaryAlignment` value that owns labelled period expansion, observation alignment, provenance, invariants, and installation into gathered inputs
- reuse the existing frequency-index calculation shared with sigma alignment
- move satellite-column boundary scaling into the same boundary-sensitivity owner
- remove the superseded `_transform_bc_freq` and `transform_bc` helpers
- preserve the concrete model boundary as the resolved labelled `H_bc` array

## Validation

- boundary, inversion-input, xarray-adapter, and fixed-basis preparation suite: 127 passed, 6 expected xfails
- focused class and satellite-scaling checks: 21 passed
- Ruff and `git diff --check`: passed
- remote whole-repository `tox -e type`: blocked by the existing baseline (198 errors across 57 files, primarily missing third-party stubs and unrelated modules)

## Checklist

- [ ] Closes a GitHub issue (tracked in Linear as OPE-93)
- [x] Tests added and passing
- [ ] Documentation/tutorial changes are not required; public Google-style API docstrings are included
- [x] Added `newsfragments/OPE-93.feature.md`
- [x] No new requirements

Linear: OPE-93

### DIFF
--- a/openghg_inversions/boundary_sensitivity.py
+++ b/openghg_inversions/boundary_sensitivity.py
@@ -12,21 +12,21 @@ from openghg_inversions.array_ops import get_xr_dummies
 from openghg_inversions.inversion_inputs import DatetimeLike, make_freq_indicator
 
 __all__ = [
-    "BoundarySensitivity",
+    "BoundaryAlignment",
     "scale_satellite_boundary_sensitivity",
 ]
 
 
 @dataclass(frozen=True, eq=False, slots=True)
-class BoundarySensitivity:
-    """Resolved sampled boundary sensitivity with labelled state invariants.
+class BoundaryAlignment:
+    """Resolved alignment of sampled boundary states to fitted periods.
 
     Args:
         data: Prepared ``H_bc`` with a ``bc_region`` MultiIndex whose levels
             are ``bc_curtain`` and ``bc_period``, plus one observation axis.
 
     Raises:
-        ValueError: If dimensions, state labels, or eager values are invalid.
+        ValueError: If dimensions or state labels are invalid.
     """
 
     data: xr.DataArray
@@ -46,8 +46,6 @@ class BoundarySensitivity:
             raise ValueError(
                 "Resolved boundary sensitivity states must be labelled by 'bc_curtain' and 'bc_period'."
             )
-        if not hasattr(data.data, "__dask_graph__") and not np.isfinite(data.values).all():
-            raise ValueError("Resolved boundary sensitivity must contain only finite values.")
         object.__setattr__(self, "data", data.rename("H_bc"))
 
     @classmethod
@@ -58,7 +56,7 @@ class BoundarySensitivity:
         frequency: str | None = None,
         anchor_time: DatetimeLike | None = None,
         observation_labels: xr.DataArray | None = None,
-    ) -> BoundarySensitivity:
+    ) -> BoundaryAlignment:
         """Align and expand raw boundary sensitivity over fitted periods.
 
         Args:
@@ -123,34 +121,6 @@ class BoundarySensitivity:
             }
         )
         return cls(data)
-
-    def install(self, inputs: xr.Dataset) -> xr.Dataset:
-        """Return gathered inputs containing this boundary sensitivity.
-
-        Args:
-            inputs: Gathered inversion inputs whose observation axis is
-                ``nmeasure``. The dataset is borrowed and is not mutated.
-
-        Returns:
-            A shallow dataset handoff containing ``H_bc`` with its labelled
-            boundary-state coordinate and provenance.
-
-        Raises:
-            ValueError: If this value is not aligned along ``nmeasure``.
-        """
-        if "nmeasure" not in self.data.dims:
-            raise ValueError("Gathered boundary sensitivity must be aligned along 'nmeasure'.")
-        data = self.data.transpose("bc_region", "nmeasure")
-        result = inputs.drop_dims("bc_region") if "bc_region" in inputs.dims else inputs.copy(deep=False)
-        result["H_bc"] = xr.DataArray(
-            data.data,
-            dims=("bc_region", "nmeasure"),
-            coords={"bc_region": data["bc_region"]},
-            name="H_bc",
-            attrs=data.attrs,
-        )
-        return result
-
 
 def scale_satellite_boundary_sensitivity(
     inputs: xr.Dataset,
--- a/openghg_inversions/boundary_sensitivity.py
+++ b/openghg_inversions/boundary_sensitivity.py
@@ -73,16 +73,16 @@ class BoundaryAlignment:
             sensitivity = sensitivity.sel({observation_dim: labels})
 
         time = sensitivity["time"].transpose(observation_dim)
-        # Dummy construction below is eager, so materialize a lazy auxiliary
-        # time coordinate once rather than executing its graph for each lookup.
-        if hasattr(time.data, "__dask_graph__"):
-            time = time.compute()
-            sensitivity = sensitivity.assign_coords(time=(observation_dim, time.data))
-        period = (
-            make_freq_indicator(time, frequency, anchor_time=anchor_time)
-            if frequency is not None
-            else xr.zeros_like(time, dtype=int)
-        ).rename("bc_period")
+        if frequency is not None:
+            # Dummy construction below is eager, so materialize a lazy auxiliary
+            # time coordinate once rather than executing its graph for each lookup.
+            if hasattr(time.data, "__dask_graph__"):
+                time = time.compute()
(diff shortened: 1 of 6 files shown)


brendan-m-murphy (commented):
The consolidation is an improvement, but one released-behaviour regression and two API-shape changes should be addressed before merge.

1. **Preserve existing NaN row cleanup**

   `BoundarySensitivity.__post_init__` rejects eager non-finite values before `make_inv_inputs` reaches `_drop_nan_and_compute`. Previously, observations with NaNs in `H_bc` were dropped; now eager input raises while equivalent Dask-backed input bypasses the check and is later dropped. Please preserve the existing cleanup behavior and make eager and lazy inputs follow the same semantics.

2. **Rename the value to `BoundaryAlignment`**

   `H_bc` is already the prepared scientific boundary-sensitivity matrix after basis functions are applied. This value describes how boundary-condition scaling factors are partitioned into optimization periods at the requested frequency. `BoundaryAlignment` is concise and parallels `SigmaAlignment`.

3. **Remove `.install()`**

   Keep replacement/assignment of `H_bc` explicit in `make_inv_inputs`, so the transformation and dataset plumbing remain visible at the call site. This also gives `BoundaryAlignment` and `MinimumError` consistent prepare-only APIs.

Reuse the existing shared frequency/period helper; no generic public alignment framework is needed here. Possible future reuse—including correlated or pyramid-weighted boundary periods and temporally correlated state vectors—is recorded in [GitHub #622](https://github.com/openghg/openghg_inversions/issues/622).

Moving construction out of `make_inv_inputs` and beside the consuming model recipe is the intended longer-term ownership direction, but it should not block this PR; it is tracked separately in [OPE-105](https://linear.app/openghg-inversions/issue/OPE-105/move-model-owned-alignment-construction-out-of-inversion-input).

openghg_inversions/inversion_data/preparation.py
```diff
@@ -925,66 +926,19 @@ def _warn_for_nan_inputs(inv_inputs: xr.Dataset, *, use_bc: bool) -> None:
...
-    try:
-        values = list(platform)
-    except TypeError:
-        return {site: None for site in sites}
-    if len(values) != len(sites):
-        return {site: None for site in sites}
-    return {site: None if value is None else str(value) for site, value in zip(sites, values, strict=True)}
-
-
 def _scale_satellite_bc_sensitivity_to_column_signal(
```
brendan-m-murphy: Why keep this wrapper? Its name is better than the name of the function it wraps, so keep the name and import it. (Or just keep the function here.)

openghg_inversions/boundary_sensitivity.py
```diff
@@ -0,0 +1,168 @@
...
+            observation_labels: Optional authoritative observation coordinate;
+                values are selected and reordered to these exact labels.
+
+        Returns:
+            Prepared boundary sensitivity carrying resolved state invariants.
+
+        Raises:
+            ValueError: If dimensions or observation labels are invalid.
+        """
+        observation_dim = next((dim for dim in ("time", "nmeasure") if dim in sensitivity.dims), None)
```
brendan-m-murphy: This function does a lot of validation for dims and coordinates. See the dev guidance: we want to defer to xarray whenever possible. These checks seem like they should either apply to all data with an observation_dim or should not be needed here. Also, the general policy is to trust data we create and make users responsible for custom data.

brendan-m-murphy: I would say, at a minimum, move this validation code into a helper method. Code that has scientific or modelling content should make that content as clear as possible. Validation obscures this.

openghg_inversions/boundary_sensitivity.py
```diff
@@ -0,0 +1,168 @@
...
+            missing = requested[~np.isin(requested, available)]
+            if missing.size:
+                raise ValueError(f"Boundary sensitivity is missing observation label(s): {missing.tolist()!r}.")
+            sensitivity = sensitivity.sel({observation_dim: requested})
+
+        time = sensitivity.coords.get("time")
+        if time is None or time.dims != (observation_dim,):
+            raise ValueError("Boundary sensitivity requires an observation-aligned 'time' coordinate.")
+        period = (
+            make_freq_indicator(time, frequency, anchor_time=anchor_time)
```
brendan-m-murphy: **[P2] Avoid executing a lazy auxiliary time coordinate twice.** For a fixed frequency with no explicit anchor, `make_freq_indicator` reads `time.min().values` and then `time.values`. When `time` is an auxiliary Dask-backed coordinate on `nmeasure`, `BoundaryAlignment.prepare` executes the same graph twice while leaving `H_bc` lazy; a delayed-coordinate reproduction confirms two executions. Materialize the coordinate once at an explicit/documented boundary, or keep the period calculation lazy until one coordinated materialization.

openghg_inversions/boundary_sensitivity.py
```diff
@@ -0,0 +1,168 @@
...
+        expanded = (sensitivity.rename(bc_region="bc_curtain") * mask).stack(
+            bc_region=("bc_curtain", "bc_period")
+        )
+        data = expanded.transpose("bc_region", observation_dim).rename("H_bc")
+        data.attrs = dict(sensitivity.attrs)
+        data.attrs.update(
+            {
+                "boundary_sensitivity_preparation": "sampled-period-expansion",
+                "bc_frequency": frequency or "none",
+                "bc_anchor_time": "first-observation" if anchor_time is None else str(anchor_time),
```
brendan-m-murphy: **[P2] Record the effective anchor, not the raw argument.** Monthly alignment ignores `anchor_time` but this attribute records it; the production path always passes `start_date`, so monthly artifacts claim an anchor that did not determine their periods. For fixed-frequency alignment without an explicit anchor, the calculation uses `time.min()` but records `first-observation`, which is false when observations are unsorted. Record the actual effective anchor/policy, and use a clear not-applicable value for modes that do not use one.

brendan-m-murphy (commented):
Re-review of the current head against the RHIME and xarray guidance.

The earlier NaN-row regression is fixed, and the rename to `BoundaryAlignment` plus removal of `.install()` are good changes. I agree with the existing threads about deleting the redundant scaling wrapper and letting xarray own labelled selection/alignment rather than surrounding the scientific operation with bespoke dimension checks.

Two additional execution/provenance issues remain below. OPE-84 is relevant only as a future pilot for mechanical schemas; these issues need ordinary xarray/execution-boundary fixes in this PR.

openghg_inversions/inversion_data/preparation.py
brendan-m-murphy: A concrete minimal fix is to keep the moved implementation but give it the clearer name, then import it directly under the existing preparation name:

```python
from openghg_inversions.boundary_sensitivity import (
    scale_satellite_boundary_sensitivity_to_column_signal
    as _scale_satellite_bc_sensitivity_to_column_signal,
)
```

Delete the forwarding function at lines 929–941. Rename the implementation/export in `boundary_sensitivity.py` to `scale_satellite_boundary_sensitivity_to_column_signal` (or retain the shorter `bc` spelling consistently). Existing callers and monkeypatch points can continue using the imported preparation-module name, but there is only one real function and one docstring. Keeping the implementation here instead is also acceptable; the important instruction is to remove the wrapper.

openghg_inversions/boundary_sensitivity.py
brendan-m-murphy: Suggested direct simplification:

```python
observation_dim = "time" if "time" in sensitivity.dims else "nmeasure"
sensitivity = sensitivity.transpose("bc_region", observation_dim)

if observation_labels is not None:
    labels = observation_labels.transpose(observation_dim)
    sensitivity = sensitivity.sel({observation_dim: labels})

time = sensitivity["time"].transpose(observation_dim)
```

Then proceed directly with period construction, dummy creation, multiplication, `stack`, and final `transpose`.

Please delete the manual `set(dims)` comparison, the coordinate-dimension/uniqueness loop, the `np.isin` missing-label calculation, and the custom missing-label branch. Exact `transpose` rejects missing/extra dimensions; `.sel` owns label lookup, ordering, missing labels, and index requirements; `stack` constructs the resolved boundary-state MultiIndex.

If direct `BoundaryAlignment(data)` construction must guarantee one defining invariant, keep only a small final check in `__post_init__` that the resolved `bc_region` index has levels `bc_curtain` and `bc_period`. Do not recheck every property of data just constructed by `.prepare`. This keeps the scientific operation readable and leaves future mechanical schema work to OPE-84.

SutarPrasad (approved):

PR #623, opened 2026-08-20
OPE-48: Simplify RHIME result and output orchestration

brendan-m-murphy:
* **Summary of changes**

Expose the canonical RHIME build → sample → result → requested-products sequence for standard, multisector, and prepared-input replay paths.

- Remove the one-use `RhimeOutputBundle` application layer.
- Consume the model build result's scientific-role and output-capability contract directly.
- Add multisector trace persistence and preserve output filenames/provenance.
- Reject unsupported multisector output modes before sampling and filesystem side effects.
- Keep rich prepared inputs and basis provenance as the durable owner where `InferenceData.constant_data` is insufficient.

~This is a stacked PR targeting `codex/ope-93-boundary-sensitivity` while OPE-93 / #621 awaits final approval.~

- [ ] Closes #xxxx (Linear: OPE-48)
- [x] Tests added and passing
- [ ] Added any new requirements to `requirements.txt` (none required)

Validation:
- Focused orchestration, replay, persistence, coordinate, and integration tests pass.
- Ruff and `git diff --check` pass.
- Broad suite: 426 passed, 3 failed, 3 deselected. One HDF backend failure passed immediately in isolation; the remaining two are pre-existing OpenGHG observation `long_name` expectation mismatches.


openghg_inversions/rhime/builders.py
```diff
@@ -141,6 +141,27 @@ def __call__(self, context: RhimeModelBuilderContext, /) -> RhimeModelBuildResul
         ...
 
 
+def validate_requested_output(
```
brendan-m-murphy: Why isn't this a method on RhimeModelBuildResult?

openghg_inversions/rhime/builders.py
```diff
@@ -155,18 +176,26 @@ def validate_model_build_result(
...
         raise ValueError(
-            f"Custom RHIME model builder does not declare output_format={output_format!r} compatible. "
-            f"Declared formats: {list(result.supported_output_formats)!r}. Use output_format='none' or "
-            "return a build result that explicitly supports the requested RHIME output contract."
+            "A custom RHIME model has output-role dimensions absent from its `CoordRegistry`: "
+            f"{unregistered_dims!r}. Register their scientific coordinates with "
+            "`registered_model(coords=...)` or `add_coords()`."
         )
+    output_format = context.run_spec.output.output_format
+    validate_requested_output(result, output_format)
```
brendan-m-murphy: Isn't this the cheaper check to run first?

openghg_inversions/rhime/outputs.py
```diff
@@ -456,41 +412,24 @@ def _make_multisector_flux_diagnostics(
...
-
-    Returns:
-        Generated inversion output, sector products, and save metadata.
     """
+    output_spec = result.output_spec
+    run_spec = result.run_spec
+    model_spec = result.model_spec
     with timed("rhime.output.inversion_output_create", output_format=output_spec.output_format):
         inv_out_for_outputs = _make_inversion_output(
+            result=result,
```
brendan-m-murphy, line 434: `output_format="none"` is documented as sampling-only, but this constructs an `InversionOutput` and sector diagnostics before inspecting the format. A successful sample can therefore still fail in unrequested postprocessing, and the new test only hides that by stubbing the diagnostics helper. Please return early for `none`, as the standard path does. The same locality rule suggests only constructing sector diagnostics for a format that requests them (currently `paris`), rather than making `inv_out` pay for an extra product.

openghg_inversions/rhime/builders.py
```diff
@@ -155,18 +176,26 @@ def validate_model_build_result(
...
-    if get_coord_registry(result.model) is None:
+    registry = get_coord_registry(result.model)
+    if registry is None:
         raise ValueError(
             "A custom RHIME model must carry a `CoordRegistry`; construct it "
             "with `registered_model()`."
         )
-    output_format = context.run_spec.output.output_format
-    if output_format not in result.supported_output_formats:
+    role_dims = {
```
brendan-m-murphy: This added completeness check makes `CoordRegistry.original_coords` the authority for whether a PyMC dimension is valid. It is not: a model can use ordinary `model.add_coord(...)`, in which case ArviZ already carries the coordinate and no RHIME restoration is needed. Even if custom models must carry a registry, please delete this `role_dims`/`unregistered_dims` preflight (and its bespoke rejection test); registration should enable restoration, not become a second coordinate-validity system.

openghg_inversions/rhime/multisector.py
```diff
@@ -598,6 +603,7 @@ def build_multisector_rhime_model_result(
             multisector=True,
             input_names=tuple(str(name) for name in prepared.inv_inputs.data_vars),
         )
+        validate_requested_output(result, run_spec.output.output_format)
```
brendan-m-murphy: This built-in capability check cannot usefully fail: the result was just constructed with the built-in multisector capability tuple, while `basic`/`legacy` incompatibility is already rejected by output-spec validation (and again at the top of this function). Please remove this call and its standard-run counterpart; retain capability validation only where a separately supplied build result can disagree with the request.

openghg_inversions/rhime/outputs.py
```diff
@@ -543,7 +478,7 @@ def make_multisector_output_bundle(
             )
             flux_outs = paris_flux_output(
                 inv_out_for_outputs,
-                country_file=country_file,
+                country_file=output_spec.country_file,
```
brendan-m-murphy: The latest PARIS flux product is already sector-aware: it contains total and per-sector flux/country outputs, so the nearby “flux total output” note understates what this call produces. What is missing is the total concentration product, which is particularly useful for verification-games. Real observations generally have no sector-level split, so concentration should remain total-only here; any optional per-sector modelled-concentration companion can remain separate (#511). Please attach/write total concentration rather than omitting it. Also, because this branch is advertised as supporting `paris`, resolve `latest` as the multisector default rather than silently producing no PARIS dataset when `template_version` is omitted.

tests/test_rhime.py
```diff
@@ -2179,7 +2207,27 @@ def test_standard_model_preserves_reparameterized_prior_and_forward_equation(
             var_names=var_names,
             random_seed=402,
         )
+    registry = models.get_coord_registry(model)
+    assert registry is not None
+    models.restore_inferencedata_coords(prior, registry)
     prior_dataset = cast(Any, prior).prior
+    constant_data = cast(Any, prior).constant_data
+
+    assert {
```
brendan-m-murphy: These exact `constant_data` contents, dimension, MultiIndex, and dropped-attrs assertions freeze an unrelated serialization design into this orchestration PR. Whether `InferenceData.constant_data` is the best record for postprocessing/serialization is still something to investigate, not a contract established here. Please remove or defer this block and keep this test focused on the graph/forward-equation behavior it owns.

brendan-m-murphy (commented):
Further review, focusing on correctness and keeping validation at the boundary that owns the input.

One unanchored simplification: `builders.py:209-245` preflights a speculative union of built-in variable roles rather than the inputs consumed by each concrete output (for example, multisector PARIS flux does not consume every observation/error role listed there). Please remove that union and let `InversionOutput` or the selected output helper report a missing value at the access boundary.

openghg_inversions/rhime/builders.py
brendan-m-murphy: Moved this onto `RhimeModelBuildResult.validate_requested_output()` in 431d88ac.

openghg_inversions/rhime/builders.py
brendan-m-murphy: Yes. The requested-output capability check now runs first in `validate_model_build_result()` (431d88ac).

openghg_inversions/rhime/outputs.py
brendan-m-murphy: Fixed in 431d88ac: `none` returns before constructing `InversionOutput`, diagnostics, or trace output, and diagnostics are now created only for PARIS.

openghg_inversions/rhime/builders.py
brendan-m-murphy: Removed the `role_dims`/`unregistered_dims` preflight and its bespoke rejection test in 431d88ac. The registry is no longer treated as coordinate authority.

openghg_inversions/rhime/multisector.py
brendan-m-murphy: Removed both built-in capability checks in 431d88ac. Capability validation remains at the separately supplied custom build-result boundary.

tests/test_rhime.py
brendan-m-murphy: Removed the unrelated `constant_data` assertions in 431d88ac; the test is again limited to graph/forward-equation behavior.

openghg_inversions/rhime/outputs.py
brendan-m-murphy: Implemented in 431d88ac: multisector PARIS now defaults to `latest`, retains sector-aware flux, and attaches/writes the total concentration product. The real multisector PARIS integration test passes.

brendan-m-murphy:
Addressed the unanchored validation note in 431d88ac as well: the speculative union of required output roles has been removed, so each selected output helper now owns reporting missing values at its access boundary.

openghg_inversions/rhime/builders.py
```diff
@@ -152,21 +161,17 @@ def validate_model_build_result(
         result: Complete model build result to validate.
         context: Prepared inputs and run settings for the active model build.
     Raises:
-        ValueError: If the requested output is unsupported or declared
-            variable roles are incomplete or refer to absent variables.
+        ValueError: If the requested output is unsupported or a declared
+            variable role refers to an absent variable.
     """
+    output_format = context.run_spec.output.output_format
+    result.validate_requested_output(output_format)
```
brendan-m-murphy, line 171: For `output_format="none"`, the runner only samples and `RhimeModelBuildResult.__post_init__()` already guarantees that `none` is supported. The simplest fix is to return immediately after reading `output_format` when it is `none`, before this method call and the registry/name preflights. That lets a sampling-only custom builder use an ordinary `pm.Model`; registry validation and coordinate-restoration guarantees remain required for RHIME postprocessing formats.

tests/integration/test_multisector_paris_pipeline.py
```diff
@@ -199,15 +199,18 @@ def test_multisector_rhime_pipeline_writes_latest_paris_flux_schema(
...
+    assert result.output_metadata["paris_concentration_path"] == str(expected_concentration_path)
     assert result.output_metadata["sector_flux_diagnostics_path"] == str(expected_diagnostics_path)
     assert expected_flux_path.is_file()
+    assert expected_concentration_path.is_file()
     assert expected_diagnostics_path.is_file()
     assert "inversion_output_path" not in result.output_metadata
-    assert "paris_concentration" not in result.outputs
-    assert "paris_concentration_path" not in result.output_metadata
-    assert not list(tmp_path.glob("*concentration*.nc"))
+    assert result.outputs["paris_concentration"].attrs["paris_concentration_template_version"] == "v04"
```
brendan-m-murphy: Because total concentration is the important verification-games addition, existence plus one version attribute is too weak: an empty, mis-dimensioned, or all-NaN product could pass. Please reopen the real concentration file and add a few direct assertions for `mf_observed`, `mf_prior`, `mf_posterior`, their expected index/platform dimensions, and finite values. No new fixture or schema helper is needed.

brendan-m-murphy (commented):
Quick follow-up review. The previous substantive findings are fixed; these are the remaining small simplifications/coverage improvements.

tests/test_rhime.py
```diff
@@ -7029,27 +7042,89 @@ def test_make_multisector_output_bundle_returns_modern_inv_out() -> None:
...
+        output_spec,
+        split_by_sectors=True,
+    )
+    prepared = RhimePreparedInputs(
+        inv_inputs=_minimal_output_inv_inputs(),
+        basis_functions=_fake_basis_functions(),
+        site_metadata=_prepared_site_metadata(),
+    )
+    result = _result_for_outputs(run_spec, _minimal_output_idata(), model_spec=model_spec)
+    monkeypatch.setattr(
```
brendan-m-murphy: This stub is now dead setup: unexpected kwargs are rejected before `_make_multisector_flux_diagnostics()` is called. Please delete the monkeypatch and the now-unused `monkeypatch` argument from this test. Net: about 5 lines.

openghg_inversions/rhime/builders.py
brendan-m-murphy: Implemented in 5ce25586: `output_format='none'` now returns before capability, registry, and variable-name preflights. The existing custom-builder test now uses an ordinary `pm.Model` and passes through the sampling-only path.

tests/integration/test_multisector_paris_pipeline.py
brendan-m-murphy: Strengthened in 5ce25586: the real NetCDF is reopened and checks 24 `index` entries, one `platform`, exact dimensions for `mf_observed`, `mf_prior`, and `mf_posterior`, and finite values. The slow integration test passes.

tests/test_rhime.py
brendan-m-murphy: Removed the dead monkeypatch and unused test argument in 5ce25586.

PR #631, opened 2026-08-21
Skip satellite BC scaling when max levels match

SutarPrasad:
## Summary

- leave satellite boundary-condition sensitivities unscaled when observation and footprint `max_level` values match
- retain the compatibility scaling only when `max_level` metadata is missing or inconsistent
- pass aligned observation and footprint `max_level` metadata through RHIME and legacy preparation

## Tests

- `pytest -q tests/test_rhime.py -k 'satellite_bc_sensitivity or preparation_uses_platform_for_sites_retained_after_filtering or assemble_rhime_inputs_preserves_borrowed'`
- `pytest -q tests/test_boundary_sensitivity.py tests/test_full_inversion.py::test_full_satellite_inversion`
- `ruff check` on changed Python paths
- `git diff --check`

### DIFF
--- a/newsfragments/+.bugfix.md
+++ b/newsfragments/+.bugfix.md
@@ -1,0 +1,1 @@
+Do not scale satellite boundary-condition sensitivities when observation and footprint `max_level` values agree; retain the compatibility workaround only for missing or inconsistent vertical extents.
--- a/openghg_inversions/boundary_sensitivity.py
+++ b/openghg_inversions/boundary_sensitivity.py
@@ -109,42 +109,56 @@ def scale_satellite_boundary_sensitivity_to_column_signal(
     *,
     sites: Sequence[str],
     platform: Sequence[str | None],
+    observation_max_level: Sequence[int | None],
+    footprint_max_level: Sequence[int | None],
 ) -> xr.Dataset:
-    """Scale satellite boundary sensitivity into corrected-column space.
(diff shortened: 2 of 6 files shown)


brendan-m-murphy (changes requested):
Thanks. Please rebase this before it can be merged: the current head is conflicting and `newsfragments/+.bugfix.md` still contains conflict markers.

More importantly, the new `footprint_max_level` is read from `fp_data[site].attrs` after the footprint and observations have been combined. OpenGHG merges footprint attrs and then observation attrs, and column observations carry `max_level`; this can therefore compare the observation level with itself and skip the compatibility scaling even when the footprint extent differs. Please carry verified footprint-level provenance through the merge (or retain scaling), and add a regression where observation max_level is 3 and footprint max_level is 17, which must still scale H_bc.

Also preserve compatibility for the exported scaler: making the two new keyword-only arguments required breaks existing callers. Finally, rebase/split the large unrelated branch delta before the next review.

brendan-m-murphy (approved):
Basic re-review of the rebuilt title-sized change found no correctness issues. The footprint max_level is preserved before the dataset-attribute merge, all preparation paths forward explicit observation/footprint provenance, and the scaler retains its compatible three-argument usage while making the new provenance inputs optional. Focused tests and Ruff passed in the implementation worktree. Residual environment caveat: a separate lean-worktree review could not collect pytest because its installed OpenGHG lacks cf_ureg.

PR #618, opened 2026-08-19
OPE-77: Add the linked CO₂/O₂ model handoff

brendan-m-murphy:
## Stack

- #616 — explicit scientific inputs and enforced coordinate registry
- #617 — activity-aware correlated state plus concrete CO₂ recipe
- This PR — linked CO₂/O₂ preparation, model, runner, config, and tests
- VG consumer/acceptance: openghg/verification-games#44

## Summary

- preserve separate, potentially unequal CO₂ and O₂ observation axes until validated joint preparation
- require shared GPP/TER/FF states and tracer-specific CO₂/O₂ ocean states
- assemble four labelled aggregation-covariance blocks, including the CO₂–O₂ cross block
- carry tracer, site, time, source role, state, row-unit, and covariance-column-unit metadata
- build from explicit arrays, `CorrelatedLognormalPrior`, `AggregationError`, and `registered_model()`
- reuse #617's activity-aware component so inactive states stay visible and fixed exactly
- require labelled fixed independent error rather than silently inferring mismatch
- persist roles, units, coordinates, and preparation provenance in InferenceData/NetCDF

## Scientific model

```text
x_active ~ correlated arithmetic-moment LogNormal
x = restore_active_and_exact_fixed_values(x_active)
mu = prior_forward + H_alpha @ (x - prior_state)
R = A_joint + diag(independent_error_sd²)
(co2, o2) ~ MVN(mu, R)
```

The UOB prototype config documents fixed 1 ppm per channel, no inferred mismatch, NumPyro, and `target_accept=0.95`. The public resolver does not yet consume those linked-channel sections.

## Validation

OGI:

- linked suite plus amended CO₂ config regression: 8 passed
- shared activity regressions: passed
- two-site/seven-day runner + NetCDF test and real partial gathered MultiIndex regression
- Ruff lint/format and `git diff --check`: passed

Real VG acceptance against this exact head:

- UOB BASE + ATEN, MHD/TAC, 12–18 July 2021; 70 rows/tracer/scenario
- all 365 public states retained: 282 active, 83 fixed exactly
- design, active prior covariance, and full cross-tracer aggregation covariance differences: 0
- prior-forward max difference: `8.88e-16 ppm`
- 1000 tune + 1000 draws, four NumPyro chains: zero divergences
- max R-hat 1.00/1.01; min bulk ESS 4680/3752
- roles, row/covariance units, labels, preparation provenance, and exact inactive values survive OGI save/load
[...]

Issue #622, opened 2026-08-20
Explore reusable observation-period partitioning for latent-state alignments


brendan-m-murphy:
## Motivation

`SigmaAlignment` and the proposed `BoundaryAlignment` both need to partition observations into optimization periods. The existing shared helper is sufficient for the current implementations, so no new public abstraction is needed now.

Record this as a possible extension point if more production consumers need the same partition semantics.

## Possible applications

- Align another site- or time-varying latent variable to observations.
- Represent boundary-condition scaling factors optimized in successive periods.
- Add fixed correlation between boundary-condition periods, for example with pyramid-shaped averaging weights rather than a dummy/indicator matrix.
- Support temporal correlation between state vectors, coordinated with the separate work already progressing in that area.

## Revisit when

At least one additional production use case lands, or the current helper can no longer express the required period mapping without duplicated semantics.

## Design constraints

- Keep scientist-facing concepts such as `SigmaAlignment` and `BoundaryAlignment` concrete.
- Prefer one small shared helper over a generic alignment framework.
- Make the distinction between observation-to-period assignment and any correlation/averaging model explicit.
- Do not pull basis-function ownership into this work.

brendan-m-murphy:
Closing as not planned for now. The issue already concluded that the existing period-index helper is sufficient, and #621 reused it for BoundaryAlignment without adding a general framework. A focused issue can be opened if another production consumer exposes a concrete missing contract.

Issue #415, opened 2026-04-29
Define and write a serialisable RHIME run bundle


brendan-m-murphy:
## Goal

Make modern RHIME runs reproducible from saved artifacts.

## Tasks

- Define required artifacts: config, run spec, canonical inputs, basis objects/operators, model spec,
  sampling settings, package versions, and output settings.
- Implement a writer for a first run-bundle format.
- Implement a reader sufficient to reproduce post-processing without reloading raw object-store data.
- Record component data needed to reconstruct derived variables.

## Acceptance criteria

- A completed modern RHIME run can write a run bundle.
- A run bundle can be loaded to reproduce post-processing.
- Package versions and key provenance are recorded.

PR #632, opened 2026-08-21
OPE-77: Add explicit CO₂/O₂ shared-state RHIME support

brendan-m-murphy:
## Summary

Replaces closed stacked PR #618 after its merged base branch was deleted.

- add explicit CO₂/O₂ preparation, shared-state model construction, and a prepared-input replay runner using `registered_model()` and labelled scientific-array contracts
- preserve unequal and staggered CO₂/O₂ observation axes with a gathered `(species, channel_observation)` index, shared GPP/TER/FF states, species-specific ocean states, all four covariance blocks, row-specific units, inactive values, and provenance metadata
- express the coherent reduction as one affine signal, with the fixed prior contribution plus one gathered joint sensitivity applied once to the shared state
- embed signed, source-resolved O₂/CO₂ ratios in the O₂ operator when representable, while recording an explicit reason when native spatial ratios cannot be reduced honestly
- preserve gathered observation and covariance identities through the existing CF and InferenceData serialization boundaries
- keep materialization at named ownership boundaries, including one shared Dask computation for related prepared inputs; the replay seam trusts prepared scientific arrays and validates only the externally supplied error labels, units, and values
- rename the CO₂-only and CO₂/O₂ modules for clarity, document the shared-state mathematics, and add a Towncrier fragment

The complete public CO₂/O₂ acquisition/output runner remains intentionally reserved; this PR exposes `run_rhime_co2_o2_from_prepared_inputs` as the reviewed replay seam.

## Validation

- broader affected CO₂, CO₂/O₂, array, serialization, component, and Ramsden suite: **206 passed, 1 deselected**
- focused changed-area suite: **134 passed**
- Ruff on changed Python paths: passed
- `git diff --check`: passed

## Checklist

- [ ] Closes a GitHub issue (tracked in Linear OPE-77 and VG integration issue openghg/verification-games#35)
- [x] Tests added and passing
- [x] Documentation updated
- [x] Towncrier news fragment added
- [x] No new requirements

Linear: OPE-77


openghg_inversions/rhime/co2/__init__.py
```diff
@@ -1,11 +1,22 @@
-"""CO2-only RHIME recipe."""
+"""Public CO2-family RHIME recipes."""
 
 from .model import build_co2_rhime_model, co2_prior_forward_mean
 from .runner import co2_model_input_names, run_rhime_co2
+from .linked_model import (
+    build_linked_co2_o2_model,
+    linked_co2_o2_prior_forward_mean,
+)
+from .linked_preparation import Co2O2PreparedInputs, prepare_linked_co2_o2_inputs
```
brendan-m-murphy: Use "co2_o2" instead of "linked".

openghg_inversions/rhime/co2/linked_model.py
```diff
@@ -0,0 +1,224 @@
...
+        output_dim=output_dim,
+    )
+
+    with registered_model() as model:
+        state = add_correlated_lognormal_state_with_activity(
+            activity,
+            retained_prior,
+            var_name="flux_scaling",
+        ).state
+        operator = add_model_data(
```
brendan-m-murphy: I agree. This changes the reuse assessment: CO₂/O₂ and Ramsden provide two concrete examples of the same component:

\[
\mu_{\text{linked}} = H_{\text{linked}}(r \odot x)
\]

where `x` is an existing state and `r` may be absent, scalar, state-resolved, fixed, or inferred.

The existing `apply_linear_sensitivity()` is already the numerical kernel. A minimal public component could wrap it:

```python
def add_linked_linear_component(
    prepared,
    state,
    /,
    *,
    state_multiplier=None,
    data_name,
    output_name,
):
    linked_state = state if state_multiplier is None else state * state_multiplier
    return apply_linear_sensitivity(
        prepared,
        linked_state,
        data_name=data_name,
        output_name=output_name,
    )
```

The important ownership split is:

- `add_linked_linear_component` owns registering and applying the linked sensitivity.
- The recipe owns constructing the scientifically named multiplier—oxidation ratio, emission ratio, or reference-ratio multiplier.
- Preparation owns provenance declaring whether a fixed reference ratio is already embedded in the sensitivity.

It should not accept prior specifications or an arbitrary transform callable. The caller should visibly construct the ratio:

```python
oxidation_ratio = add_oxidation_ratio(...)
o2_signal = add_linked_linear_component(
    o2_sensitivity,
    flux_scaling,
    state_multiplier=oxidation_ratio,
    data_name="o2_sensitivity",
    output_name="modelled_o2",
)
```

For Ramsden, this replaces the current `apply_linear_sensitivity(..., state * ratio_factor)` expression directly.

This also suggests that CO₂/O₂ preparation should retain separate channel operators until model construction. A pre-stacked joint operator cannot expose a multiplier that applies only to the tracer channel without also scaling the primary channel. Build the CO₂ and O₂ signals separately, then stack them for the joint covariance likelihood.

So I’d distinguish the names as follows:

- `co2_o2_*`: concrete preparation, model, runner, and likelihood.
- `add_linked_linear_component`: reusable graph component beside `add_linear_component`.
- `tracer`: the scientific channel role.
- `state_multiplier`: the generic graph input; recipes give it the meaningful name `oxidation_ratio`, `emission_ratio`, etc.

That is demonstrated reuse, not a speculative generic multi-gas framework.

openghg_inversions/rhime/co2/linked_model.py
```diff
@@ -0,0 +1,224 @@
...
+    output_dim: str = "observation",
+    covariance_dim: str = "observation_cov",
+) -> pm.Model:
+    """Build one affine linked signal and fixed-error joint likelihood.
+
+    All scientific inputs are explicit labelled values. The UOB prototype
+    supplies one ppm as ``independent_error_sd`` for each channel; this builder
+    does not encode that run policy and does not infer a mismatch amplitude.
+    """
+    independent_error = _validate_linked_arrays(
```
brendan-m-murphy: This broad validation should not be mandatory inside the concrete builder. `run_rhime_co2_o2()` forwards arrays created by `prepare_linked_co2_o2_inputs()`, while `CorrelatedLognormalPrior`, `AggregationError`, state-activity resolution, and `registered_model()` / `CoordRegistry` already own their respective contracts. Rechecking dimensions, finite values, exact state roles, and descriptive unit coordinates here makes the model distrust package-created data and forces custom builders through policy checks they should either opt into or own themselves.

Please remove the unconditional `_validate_linked_arrays()` call. If direct/custom pipelines want the complete recipe check, expose it as an optional helper. The one value not created by preparation is `independent_error_sd`; a narrow finite-and-positive standard-deviation check can live at the standard runner boundary that owns that external input. Direct custom model callers remain responsible for correctness.

This cleanup should also remove the unused `aggregation_error` and `covariance_dim` validator arguments. `covariance_dim` is currently exposed as a public builder option but has no effect, so it should not remain as dead configurability.

openghg_inversions/rhime/co2/linked_model.py
```diff
@@ -0,0 +1,224 @@
...
+        effective_observation_operator,
+        state_activity,
+        output_dim=output_dim,
+    )
+    prior_state = xr.where(
+        activity.active,
+        retained_prior.mean,
+        activity.fixed_value,
+    ).rename("prior_flux_scaling")
+    activity_prior_forward = linked_co2_o2_prior_forward_mean(
```
brendan-m-murphy: Resolve state activity once and reuse the resulting `activity` and `prior_state`. `linked_co2_o2_prior_forward_mean()` calls `_resolve_activity()` again, so `detect_zero_sensitivity()` eagerly scans and materializes the full operator twice during one model build. Besides the unnecessary work, this obscures that the prior-state adjustment and sampled state are governed by the same resolved policy.

Either compute the prior-forward adjustment directly from the already-resolved values here, or use a small helper that accepts `ResolvedStateActivity`; do not rerun resolution on a locally constructed intermediate.

openghg_inversions/rhime/co2/co2_o2_preparation.py
```diff
@@ -0,0 +1,277 @@
...
+        raise ValueError("Linked CO2/O2 provenance must be JSON serializable.") from exc
+
+    co2_dim = _axis(co2_observations, "CO2 observations")
+    o2_dim = _axis(o2_observations, "O2 observations")
+    if co2_dim == o2_dim:
+        raise ValueError("CO2 and O2 require distinct pre-stacking dimension names.")
+    _same_axis(co2_observations, co2_prior_forward_mean, "CO2 prior forward mean")
+    _same_axis(o2_observations, o2_prior_forward_mean, "O2 prior forward mean")
+    state_mean, state_dim = _state(retained_prior)
+    _operator(co2_operator, co2_observations, state_mean, "CO2")
```
brendan-m-murphy, line 409: The preparation boundary currently validates the operator dimensions and labels, but not the model-defining tracer-loading invariant. It accepts a nonzero CO₂-operator coefficient for the O₂-only ocean state, or a nonzero O₂ coefficient for the CO₂-only ocean state, and the model will sample that scientifically different graph while still describing those states as tracer-specific.

This is a mandatory check that belongs here: these are independently supplied per-channel operators and `prepare_linked_co2_o2_inputs()` is the first boundary that combines them with the declared `source` / `tracer_scope` roles. Require the cross-tracer ocean loadings to be zero and add one focused negative test for each direction. This is distinct from the redundant downstream validation in the builder: it protects a genuine recipe policy at the owning boundary.

openghg_inversions/rhime/co2/linked_preparation.py
```diff
@@ -0,0 +1,277 @@
...
+    )
+    prior_forward = _stack(
+        co2_prior_forward_mean,
+        o2_prior_forward_mean,
+        co2_units=co2_units,
+        o2_units=o2_units,
+        name="prior_forward_concentration",
+    )
+    operator = xr.DataArray(
+        np.concatenate((co2_operator.values, o2_operator.values), axis=0),
```
brendan-m-murphy: This preparation object is described as backend-neutral, but the implementation eagerly converts related arrays through `.values` / NumPy in `_covariance()`, `_stack()`, and this operator concatenation. For Dask-backed inputs that can schedule shared graphs separately before the PyMC boundary and makes the execution boundary invisible to the runner.

Please keep the channel vectors and operators labelled and lazy with ordinary xarray transpose/alignment/concatenation. Materialize the selected related model inputs together at an explicit runner/PyMC boundary, following the existing `materialize_pymc_inputs()` pattern used by the CO₂ runner. Dense covariance validation may remain at the named `resolve_aggregation_error()` eager boundary, but it should receive the combined labelled covariance rather than arrays already materialized piecemeal.

openghg_inversions/rhime/co2/linked_runner.py
```diff
@@ -0,0 +1,138 @@
...
+                "observation_units per dimensionless flux scale"
+            )
+        if "aggregation_error_covariance" in group:
+            group["aggregation_error_covariance"].attrs["units"] = "observation_units * observation_units_cov"
+        group.attrs["rhime_recipe"] = "co2_o2"
+        setattr(trace, group_name, group)
+    return trace
+
+
+def run_rhime_co2_o2(
```
brendan-m-murphy: This exported function is a prepared-input replay seam, not the complete procedural CO₂/O₂ recipe required by the active model-family plan. It starts after channel preparation, state/covariance construction, and materialization; meanwhile the added `co2_o2.ini` explicitly says the resolver does not consume it and callers must assemble the scientific inputs themselves. The corresponding config test only reparses and asserts those literals, so it cannot detect drift from the actual runner.

Please either add the public top-to-bottom runner/config path that visibly performs preparation → explicit materialization → model construction → sampling/output, or name this function `run_rhime_co2_o2_from_prepared_inputs` and document it as the advanced replay seam until the production runner exists. The public `run_rhime_co2_o2` name should be reserved for the readable complete recipe.

brendan-m-murphy (commented):
The model mathematics are promising, but I am requesting the design changes described inline before treating this as the production CO₂/O₂ recipe. The main changes are:

- introduce the demonstrated reusable linked-linear component for applying a tracer sensitivity—and an explicit optional state multiplier—to an existing state;
- keep CO₂/O₂ recipe names concrete while leaving the linked component general;
- move validation to the boundaries that own it and trust package-created intermediates;
- preserve labelled/lazy arrays until one explicit joint materialization boundary; and
- either provide the complete procedural runner with a runnable config or clearly expose this as a prepared-input replay seam.

The current CI is green; these comments concern scientific readability, customization responsibility, and the production architecture described in the repository guidance.

openghg_inversions/rhime/co2/co2_o2_model.py
```diff
@@ -0,0 +1,214 @@
...
+        prior_state_data = add_model_data(prior_state)
+        prior_forward_data = add_model_data(activity_prior_forward)
+        state_anomaly = state - prior_state_data
+        co2_adjustment = add_linked_linear_component(
+            co2_sensitivity,
+            state_anomaly,
+            data_name="co2_operator",
+            output_name="co2_flux_adjustment",
+        )
+        o2_adjustment = add_linked_linear_component(
```
brendan-m-murphy: **[P2] Make the O₂ oxidation-ratio contract explicit**

The shared component now has the right generic seam—`state_multiplier`—and Ramsden demonstrates it with a recipe-owned emission ratio, but the CO₂/O₂ tracer call still applies the raw `state_anomaly`. That means the oxidation ratio must be hidden inside `o2_operator`, while `prepare_co2_o2_inputs()` exposes no reference-ratio declaration that distinguishes a ratio-free sensitivity from one which already embeds a fixed ratio. An opaque `provenance` mapping does not prevent omission or double application.

Please either construct a scientifically named `oxidation_ratio` in this recipe and pass `state_multiplier=oxidation_ratio`, or make the preparation contract explicitly declare and validate the reference ratio, direction/sign convention, and state/source scope already included in the O₂ operator. The linked component should remain generic; the CO₂/O₂ recipe should own this scientific meaning and persist it in result metadata.

openghg_inversions/rhime/co2/co2_o2_model.py
```diff
@@ -0,0 +1,214 @@
...
+            "modelled_concentration",
+            pt.concatenate(
+                (
+                    prior_forward_data[:nco2] + co2_adjustment,
+                    prior_forward_data[nco2:] + o2_adjustment,
+                )
+            ),
+            dims=output_dim,
+        )
+        observed = add_model_data(observations, "observed_concentration")
```
brendan-m-murphy: **[P2] Move backend data registration into scientific components**

The linked-sensitivity registration is now properly owned by `add_linked_linear_component`, but the concrete graph still exposes four bare `add_model_data` calls, manually concatenates the two channel signals, registers aggregation data separately, and then calls the tensor-level likelihood helper. A scientist has to reconstruct the affine CO₂/O₂ signal and fixed-error joint likelihood from backend plumbing.

Please keep two ordinary, recipe-local composite functions beside this builder: a CO₂/O₂ affine signal component that owns prior-state/prior-forward registration plus channel stacking, and a fixed-error joint-likelihood component that owns observations, independent error, aggregation-error registration, and the Gaussian likelihood. No framework or result dataclass is needed. The concrete builder should read in scientific order as state → CO₂/O₂ signal → joint likelihood.

openghg_inversions/rhime/co2/co2_o2_runner.py
```diff
@@ -0,0 +1,192 @@
...
+from openghg_inversions.rhime.builders import RhimeModelBuildResult
+from openghg_inversions.rhime.sampling import RhimeSampler, sample_rhime_model
+
+from .co2_o2_model import build_co2_o2_model
+from .co2_o2_preparation import Co2O2PreparedInputs
+
+
+_CO2_O2_VARIABLE_ROLES = {
+    "observation": "observed_concentration",
+    "concentration": "modelled_concentration",
```
brendan-m-murphy: **[P2] Map the concentration role to the likelihood variable**

The canonical `concentration` role should identify the observed/posterior-predictive variable `y`, not the deterministic model mean. `add_gaussian_observation_likelihood()` creates `y`, and the standard and CO₂ recipes both map `"concentration": "y"`; they retain a separate role for the mean. Mapping concentration to `modelled_concentration` will make trace metadata and downstream role lookup treat the mean as the sampled concentration.

Change this entry to `"concentration": "y"` and keep `"modelled_concentration": "modelled_concentration"` for the mean. Please add a focused assertion for this role—the current end-to-end test only checks the flux-scaling role.

brendan-m-murphy (commented):
Re-reviewed head `a77fef24`. CI is green, and six original threads are now fully addressed and have been resolved: concrete `co2_o2` naming, builder validation ownership, single state-activity resolution, cross-tracer ocean loading checks, lazy/joint materialization, and the prepared-input replay seam naming.

I left the linked-component thread open because the reusable component is present and Ramsden uses its ratio multiplier, but the CO₂/O₂ recipe still neither supplies an explicit oxidation ratio nor declares the fixed ratio already embedded in its O₂ operator. I added current-line comments for that remaining contract, the still-low-level signal/likelihood composition, and an incorrect `concentration` variable-role mapping.

Before merge, please also refresh the PR title/body: the summary and validation section still describe the previous head and seven-test patch, and the documentation checklist no longer matches the removal of the non-runnable config.

openghg_inversions/rhime/co2/co2_o2_model.py
brendan-m-murphy: The direction/sign/scope metadata is a useful improvement, but I think one simplicity and reproducibility issue remains. `o2_operator_ratio_convention` is mandatory while its type admits exactly one literal and preparation rejects every other value. Every caller must therefore repeat the magic string, but the check cannot establish that the operator actually follows that convention. This is ceremonial validation rather than an owned data invariant. At the same time, `reference_ratio` in result metadata is only the sentence “source/spatially resolved in the paired native O2 flux”; it does not persist the actual source/state-resolved ratios needed to reconstruct the scientific equation. Please remove the one-value argument/check and make the embedded convention an unconditional property of this concrete preparation recipe, while requiring or deriving the actual ratio provenance (values keyed by source/state, including sign/direction) from the input data and persisting it. If the actual ratios are not available at this boundary, state that explicitly in the contract and track it rather than presenting the prose placeholder as reference-ratio provenance. This keeps custom callers responsible for their operator while making package-produced outputs scientifically recoverable.

brendan-m-murphy (commented):
Quick re-review of `f1c4c274`: the scientific component ownership and variable-role changes look good, and CI is green. One non-code cleanup remains before merge: the PR title/body still describe the implementation as the “linked” CO2/O2 handoff, list the old `916db02e` head and seven-test result, and claim documentation was updated although the current file list contains no CO2/O2 recipe documentation. Please refresh the title/summary/validation/checklist to describe the concrete `co2_o2` recipe plus reusable linked-linear component and the current head/test evidence. I left the oxidation-ratio thread open with the remaining simplicity/reproducibility concern.

openghg_inversions/rhime/co2/co2_o2_model.py
```diff
@@ -0,0 +1,255 @@
...
+        co2_operator,
+        o2_operator,
+    )
+    activity = _resolve_activity(co2_sensitivity, state_activity)
+    prior_state = xr.where(
+        activity.active,
+        retained_prior.mean,
+        activity.fixed_value,
+    ).rename("prior_flux_scaling")
+    activity_prior_forward = _co2_o2_prior_forward_mean(
```
brendan-m-murphy: **[P2] Consume the coherent-reduction affine intercept**

This re-centres an affine model that coherent reduction has already centred. `CoherentGaussianReduction` owns `observation_intercept = Hm - H_alpha Pi m`, and the current CO₂ recipe already consumes that product directly as `fixed_prior_contribution + H_alpha @ x`. Here `_co2_o2_prior_forward_mean()` first adds `H(p - m)`, then `add_co2_o2_affine_signal()` adds `H(x - p)`; those steps cancel algebraically to `prior_forward_mean + H(x - m)`.

Please keep the preparation-owned reduction product explicit: carry the joint `observation_intercept` (or consistently named `fixed_prior_contribution`) in `Co2O2PreparedInputs`, then construct `fixed_prior_contribution + concat(H_co2 @ state, H_o2 @ state)`. Exact fixed states already occupy their values in the restored full state vector, so they do not require changing the intercept. That lets us remove the duplicate prior-state centring and state-anomaly registration while keeping CO₂ and CO₂/O₂ on the same scientific equation. The closure-check helper may remain if it mirrors `co2_prior_forward_mean()`, but it should not be needed during graph construction. The CO₂-only model should not call this two-channel helper; it already implements the canonical form. If a genuinely shared affine graph component remains after the inputs are aligned, extract that small operation generically.

openghg_inversions/rhime/co2/__init__.py
```diff
@@ -1,11 +1,22 @@
-"""CO2-only RHIME recipe."""
+"""Public CO2-family RHIME recipes."""
 
 from .model import build_co2_rhime_model, co2_prior_forward_mean
```
brendan-m-murphy: **[P2] Name both peer recipes explicitly**

This package now owns two peer recipes, but the CO₂-only pair is hidden behind generic `model.py` / `runner.py` while the CO₂/O₂ pair is scientifically named. Please rename the former to `co2_model.py` and `co2_runner.py`, yielding the flat symmetric layout `co2_model.py`, `co2_runner.py`, `co2_o2_preparation.py`, `co2_o2_model.py`, and `co2_o2_runner.py`. Public imports from `openghg_inversions.rhime.co2` can remain unchanged, and repository call sites require only the local import updates. No compatibility shims or nested subpackages are needed unless these very recent internal module paths have already been released or documented.

openghg_inversions/rhime/co2/co2_o2_model.py
brendan-m-murphy: The user documentation should make this shared mathematics explicit as part of the same change. Please add a brief **CO₂/O₂ shared-state model** subsection immediately after the CO₂ coherent-reduction section in `docs/usage/concrete_rhime_model.rst`, rather than leaving the equation recoverable only from code.

A compact block form would be scientifically clearest. Partition the retained state as `alpha = [alpha_shared; alpha_co2_ocean; alpha_o2_ocean]`, where the shared block contains GPP, TER, and fossil-fuel states, and write

```text
H_joint = [
  H_co2,shared      H_co2,ocean      0
  H_o2,shared_eff   0                 H_o2,ocean
]
b_joint = [b_co2; b_o2]
mu_joint = b_joint + H_joint @ alpha
```

This is a row-stacked, block-sparse operator acting on one state vector, not two independent block-diagonal models. The likelihood is

```text
[y_co2; y_o2] | alpha ~ Normal(
    mu_joint,
    [[A_co2,co2, A_co2,o2],
     [A_o2,co2,  A_o2,o2]] + diag(s_independent**2),
)
```

Relate these inputs directly to the existing coherent-reduction notation: `C_alpha = Pi B Pi.T`, `H_joint = G B Pi.T C_alpha^-1`, `b_joint = G m - H_joint Pi m`, and `A = G B G.T - H_joint C_alpha H_joint.T`. The implementation may preserve separate channel arrays because ppm and per-meg rows have different units, but all operator and covariance blocks must be products of the same coherent reduction; the off-diagonal `A_co2,o2` block is part of that contract.

Also explain the precise Ramsden relationship and cross-reference `docs/experimental/ramsden2022.rst`: Ramsden writes the tracer term as `H_tracer @ (R * alpha_shared)`. For this CO₂/O₂ recipe the signed, source-resolved oxidation ratio is fixed and already folded into `H_o2,shared_eff`, equivalently `H_o2,shared_eff = H_o2,ratio_free @ diag(R)`. If `R` were sampled, the tracer term would be bilinear and would not be the fixed linear coherent-reduction model implemented here. Cross-reference `docs/usage/coherent_reduction.rst` for the full derivation instead of duplicating it.

openghg_inversions/models/components.py
```diff
@@ -283,6 +283,25 @@ def apply_linear_sensitivity(
     return cast(TensorVariable, output)
 
 
+def add_linked_linear_component(
+    prepared: PreparedLinearSensitivity,
+    state: TensorVariable,
+    /,
+    *,
+    state_multiplier: TensorVariable | np.ndarray | float | None = None,
```
brendan-m-murphy: **[P2] Keep the state/ratio multiplication visible in the recipe**

`state_multiplier` hides the scientifically meaningful `linked_state = state * ratio` operation inside a generic sensitivity wrapper. Whether the complete graph is called linear or bilinear depends on what is conditioned on and on the parameterization of `state`; that classification is less useful here than showing where the ratio enters. Please make `add_linked_linear_component` take the already constructed linked state and limit it to registering/applying the linked sensitivity. A recipe with an explicit ratio should read `linked_state = state * oxidation_ratio` (or `emission_ratio`) immediately before the call. Ramsden should likewise form `state * ratio_factor` in its builder. The current CO₂/O₂ recipe should continue passing the unmodified state because its fixed signed ratios are already embedded in the O₂ operator; multiplying again would be wrong. Add a complete Google-style docstring explaining that existing-state contract and the prepared sensitivity/output ownership.

openghg_inversions/rhime/co2/co2_o2_model.py
```diff
@@ -0,0 +1,221 @@
...
+    fixed_prior_contribution: xr.DataArray,
+    co2_operator: xr.DataArray,
+    o2_operator: xr.DataArray,
+    aggregation_error: AggregationError,
+    retained_prior: CorrelatedLognormalPrior,
+    independent_error_sd: xr.DataArray,
+    state_activity: StateActivity | None = None,
+    output_dim: str = "observation",
+) -> pm.Model:
+    """Build separate CO2/O2 signals and one fixed-error joint likelihood.
```
brendan-m-murphy: **[P2] Give the model functions complete scientific docstrings**

The exported builder and closure helper, plus the two public-by-name composite functions, currently have one-paragraph or one-line docstrings. Please use Google-style Args, Returns, and applicable Raises sections throughout this module. The builder contract needs to state the dimensions and meanings of every labelled array, that fixed_prior_contribution is the coherent intercept, that the O₂ operator already contains fixed signed ratios, that custom direct callers own input correctness, and how the joint covariance and independent error enter the likelihood. Either prefix the two recipe-local composites with _ or document them as supported extension components. Do not hard-code the O₂ units: the current verification-game inputs use ppm for both channels, while real O₂ data may use per-meg delta(O₂/N₂). Describe independent_error_sd as already expressed in each observation row native declared channel units.

openghg_inversions/rhime/co2/co2_o2_model.py
```diff
@@ -0,0 +1,221 @@
...
+    add_gaussian_observation_likelihood,
+)
+from openghg_inversions.observation_error import AggregationError
+
+
+def _prepare_channel_sensitivities(
+    co2_operator: xr.DataArray,
+    o2_operator: xr.DataArray,
+) -> tuple[PreparedLinearSensitivity, PreparedLinearSensitivity]:
+    """Inspect both channels once and retain their independent output axes."""
```
brendan-m-murphy: **[P3] Explain the joint zero-sensitivity preparation**

The concatenate-prepare-split operation is necessary, but this docstring does not reveal why: a state column must be removed only when it is zero in both channels, while each channel keeps its independent observation axis. Please rename this to something like `_prepare_shared_state_channel_sensitivities` and document that invariant with Google-style arguments and return values. Preparing each channel independently would incorrectly drop a state used only by the other channel. The public `co2_o2_prior_forward_mean` can then reuse the prepared joint operator for one labelled dot product rather than rebuilding channel slices and a Python contribution loop; if the loop remains, explain that it exists solely to restore unequal channel axes before stacking.

openghg_inversions/rhime/co2/co2_o2_preparation.py
```diff
@@ -0,0 +1,463 @@
...
+    o2_co2_flux_ratio_unavailable_reason: str | None,
+    co2_aggregation_covariance: xr.DataArray,
+    co2_o2_aggregation_covariance: xr.DataArray,
+    o2_aggregation_covariance: xr.DataArray,
+    retained_prior: CorrelatedLognormalPrior,
+    co2_units: str,
+    o2_units: str,
+    provenance: Mapping[str, Any] | None = None,
+) -> Co2O2PreparedInputs:
+    """Validate coherent-reduction channel products and form one joint likelihood.
```
brendan-m-murphy, line 332: **[P2] Document the complete prepared-input boundary**

This exported function has fifteen scientific inputs but no Google-style `Args`, `Returns`, or `Raises` contract. Please define every vector and matrix dimension, indexed-label requirement, covariance-block orientation, channel-unit convention, and the exactly-one ratio-vector/unavailable-reason rule. For an available ratio, document the required `source` coordinate and `direction`, `sign_convention`, and `provenance` attributes that the function enforces. `Co2O2PreparedInputs` also needs an `Attributes` section and should say callers normally obtain it from this factory. While doing this, rename or split `_validate_ocean_loadings`: it also materializes, validates, and returns ratio values, which contradicts its current name and docstring.

openghg_inversions/rhime/co2/co2_o2_runner.py
```diff
@@ -0,0 +1,201 @@
...
+from openghg_inversions.models import StateActivity
+from openghg_inversions.rhime.builders import RhimeModelBuildResult
+from openghg_inversions.rhime.sampling import RhimeSampler, sample_rhime_model
+
+from .co2_o2_model import build_co2_o2_model
+from .co2_o2_preparation import Co2O2PreparedInputs
+
+
+_CO2_O2_VARIABLE_ROLES = {
+    "observation": "observed_concentration",
```
brendan-m-murphy: **[P2] Point the observation role at data that survives sampling**

Keeping observations as `pm.Data` inside the complete likelihood component is sound: PyMC supports a data container as `observed=`, writes the values to `observed_data["y"]`, and normally omits the redundant value container from `constant_data`. Consequently `observed_concentration` usually does not exist in the returned `InferenceData`, so this role is dangling. Map both `observation` and `concentration` to `y`, and assert that `observed_data["y"]` preserves the prepared values, tracer coordinates, units, and both role annotations. Keep `fixed_independent_error_sd` separately in `constant_data`; unlike the observed RV, it is a fixed graph input. The standard, multisector, additive-sigma, Ramsden, and CO₂/O₂ likelihood components already own error registration consistently, so no broader builder-level registration abstraction is needed.

docs/usage/concrete_rhime_model.rst
```diff
@@ -239,6 +239,95 @@ it. Persist gathered-state traces with
...
+Native paired-flux construction may instead apply spatially resolved ratios
+before footprint convolution, in which case no unique retained-state
+:math:`R` is available. Preparation records that status and its reason rather
+than inventing scalar values; the supplied effective O2 operator remains the
+scientific input.
+
+The retained/source-resolved form is equivalent to the tracer term
+:math:`H_{tracer}(R\mathbin{\odot}\alpha_{shared})` used by the
+:doc:`Ramsden methane/ethane model <../experimental/ramsden2022>`. Ramsden may
+infer its emission ratio; if :math:`R` were sampled here, the tracer term would
```
brendan-m-murphy: **[P3] Describe where the ratio multiplication occurs**

Please avoid making “bilinear rather than linear” the main boundary here. Conditional on a ratio, the tracer operation is linear in the linked state; viewed through a latent parameterization, even an ordinary lognormal linear component is nonlinear in its latent standard normal. The scientifically useful distinction is where multiplication happens. Say that this recipe receives an O₂ operator with the fixed ratio already applied upstream, so the builder passes the shared state directly. If a fixed or inferred ratio were explicit model state, the recipe should visibly form `o2_state = oxidation_ratio * co2_state` and pass that result to `add_linked_linear_component`. This also makes the contrast with the Ramsden builder precise without implying that the linked component itself owns scaling.

openghg_inversions/rhime/co2/co2_o2_runner.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+def run_rhime_co2_o2_from_prepared_inputs(
+    *,
+    prepared_inputs: Co2O2PreparedInputs,
+    independent_error_sd: xr.DataArray,
+    state_activity: StateActivity | None = None,
+    sampler: RhimeSampler | None = None,
+) -> az.InferenceData:
+    """Build and sample the CO2/O2 model from prepared scientific inputs.
```
brendan-m-murphy, line 200: **[P2] Document the replay contract in Google style**

Please add complete `Args`, `Returns`, and `Raises` sections. In particular, `independent_error_sd` must be indexed on the stacked observation axis, already expressed numerically in each row channel units, and finite and positive; `prepared_inputs` should normally come from `prepare_co2_o2_inputs`; and this function begins after acquisition/preparation but owns joint materialization, the narrow external-SD validation, sampling, and trace metadata. The current prose explains the deferred production runner but not the callable contract a custom replay pipeline must satisfy.

docs/usage/concrete_rhime_model.rst
```diff
@@ -239,6 +239,95 @@ it. Persist gathered-state traces with
 :func:`openghg_inversions.serialization.save_inferencedata`, which uses the
 same MultiIndex-safe boundary as standard and multisector RHIME outputs.
 
+CO2/O2 shared-state model
```
brendan-m-murphy, line 242: **[P3] Connect the equations to the public API**

The new mathematical section is useful, but it never links the functions that implement the preparation, graph, and replay boundaries. Add a short paragraph with fully qualified `:func:` links for `prepare_co2_o2_inputs`, `build_co2_o2_model`, and `run_rhime_co2_o2_from_prepared_inputs`, and include the CO₂ subpackage in the generated reference documentation or autodocument those functions here. At present `rhime/co2/__init__.py` exports them, but `docs/reference/openghg_inversions.rhime.rst` omits the subpackage, so users can find the equations but not the released call contracts.

openghg_inversions/rhime/co2/co2_o2_model.py
```diff
@@ -0,0 +1,221 @@
...
+    # Preparation declares that this operator already contains signed,
+    # source-resolved O2:CO2 oxidation ratios; do not multiply them again.
+    o2_signal = add_linked_linear_component(
+        o2_sensitivity,
+        state,
+        data_name="o2_operator",
+        output_name="o2_flux_contribution",
+    )
+    return pm.Deterministic(
+        "modelled_concentration",
```
brendan-m-murphy: **[P2] Encode the coherent affine contract in one small component**

The useful abstraction is not generic tensor addition. It is the operation every coherent-reduction recipe must apply consistently: `modelled = fixed_prior_contribution + linear_contribution`. Please add a narrow ordinary `add_coherent_affine_component` that accepts the labelled fixed contribution, an already-composed linear signal, and the output name; it owns registration of the fixed contribution and always creates the canonical `pm.Deterministic`. It must not accept sensitivities, states, ratios, channel lists, or validation. This recipe should continue to construct the CO₂ and O₂ linked signals and concatenate them visibly before the call; an explicit ratio would likewise remain visible as `tracer_state = ratio * state`, while the present fixed ratio remains embedded in the O₂ operator.

Document the equivalent coherent-reduction forms `mu = mu_prior + H_alpha (x - m_alpha) = (mu_prior - H_alpha m_alpha) + H_alpha x`, so users understand that `fixed_prior_contribution` is the affine intercept rather than the complete prior-forward concentration. Keep the deterministic in the PyMC graph. Whether it is stored in the posterior is sampler policy: `pm.sample(var_names=...)` can omit large observation-aligned deterministics and `pm.compute_deterministics(...)` can recover them once the same model is available or rebuilt.

Please keep the remainder out of PR 632. Applying this component to the existing CO₂ model, revisiting standard/multisector component ownership, selecting trace-retention defaults, and rebuilding omitted deterministics after serialization belong in follow-up issues #397, #414, and #415.

docs/usage/concrete_rhime_model.rst
```diff
@@ -239,6 +239,95 @@ it. Persist gathered-state traces with
...
+   C_\alpha &= \Pi B\Pi^\mathsf{T}, \\
+   H_{joint} &= GB\Pi^\mathsf{T}C_\alpha^{-1}, \\
+   b_{joint} &= Gm - H_{joint}\Pi m, \\
+   A &= GBG^\mathsf{T} - H_{joint}C_\alpha H_{joint}^\mathsf{T}.
+
+In particular, the off-diagonal :math:`A_{CO_2,O_2}` block is part of the
+coherent-reduction contract. See the :doc:`full derivation
+<coherent_reduction>` for its assumptions and limitations.
+
+The implementation preserves separate channel arrays because CO2 rows in ppm
```
brendan-m-murphy: **[P2] Keep channel units data-dependent**

This is too specific: verification-game inputs may express both CO₂ and O₂ numerically in ppm, while real atmospheric O₂ observations may use delta(O₂/N₂) in per meg. Please say that the separate channel arrays preserve each row declared native units and scale; `independent_error_sd` and each covariance row and column must use those corresponding units. If numerical channel scaling or whitening is later needed even for ppm/ppm data, make it a named transformation that is applied consistently to observations, model mean, independent error, and every joint covariance block, while retaining the physical-unit model outputs and provenance. It should not be implied by the unit strings or hidden in this affine component.

openghg_inversions/rhime/co2/co2_o2_runner.py
```diff
@@ -0,0 +1,273 @@
...
+
+from .co2_o2_model import build_co2_o2_model
+from .co2_o2_preparation import Co2O2PreparedInputs
+
+
+_CO2_O2_VARIABLE_ROLES = {
+    "observation": "y",
+    "concentration": "y",
+    "modelled_concentration": "modelled_concentration",
+    "pollution_concentration": "modelled_concentration",
```
brendan-m-murphy: `modelled_concentration` includes `fixed_prior_contribution` as well as both channel flux signals, so mapping `pollution_concentration` to it changes that role's meaning relative to the CO₂ recipe, where the pollution role is flux-only. Role-based consumers will silently receive total modelled concentration. Please remove this alias or expose a joint flux-only deterministic and map the role to that.

openghg_inversions/rhime/co2/co2_o2_preparation.py
```diff
@@ -0,0 +1,561 @@
...
+    fixed_prior_contribution = _stack(
+        co2_intercept,
+        o2_intercept,
+        co2_units=co2_units,
+        o2_units=o2_units,
+        name="fixed_prior_contribution",
+    )
+    fixed_prior_contribution.attrs["mathematical_name"] = "H m - H_alpha Pi m"
+    operator_coords = {name: state_mean[name] for name in ("source", "tracer_scope")}
+    co2_operator = co2_operator.rename("co2_effective_observation_operator").assign_coords(
```
brendan-m-murphy: The operator state index is validated, but an existing `source` or `tracer_scope` coordinate is preserved without being checked against `retained_prior`. An operator with correct state labels but stale auxiliary labels passes preparation and then fails during model coordinate registration. This prepared boundary should either require exact auxiliary-coordinate agreement or canonicalize these coordinates from the prior.

openghg_inversions/rhime/co2/co2_o2_preparation.py
```diff
@@ -0,0 +1,561 @@
...
+
+    def channel(
+        value: xr.DataArray,
+        *,
+        start: int,
+        tracer: str,
+        units: str,
+    ) -> xr.DataArray:
+        original_dim = str(value.dims[0])
+        labels = [str(label) for label in value.indexes[original_dim]]
```
brendan-m-murphy: Stringifying native observation labels is lossy: integers and datetimes change type, MultiIndex structure disappears, and distinct labels such as `1` and `"1"` can collide. Since the joint observation index is replaced with integers, `within_tracer_observation` is the only retained native-label record. Please preserve the original label values losslessly.

openghg_inversions/rhime/co2/co2_o2_runner.py
```diff
@@ -0,0 +1,273 @@
...
+        o2_operator,
+        independent_error_sd,
+    ) = _materialize_co2_o2_pymc_inputs(
+        prepared.observations,
+        prepared.fixed_prior_contribution,
+        prepared.co2_operator,
+        prepared.o2_operator,
+        independent_error_sd,
+    )
+    if not np.isfinite(independent_error_sd.data).all() or (
```
brendan-m-murphy: This eager boundary validates only `independent_error_sd`. NaN observations or a NaN `fixed_prior_contribution` can pass preparation and then fail inside `pm.Data` with an internal `NotImplementedError`. Please validate the materialized observations and affine contribution as finite numeric vectors here so invalid scientific inputs fail deterministically at the owning boundary.

openghg_inversions/rhime/co2/__init__.py
```diff
@@ -1,11 +1,22 @@
-"""CO2-only RHIME recipe."""
+"""Public CO2-family RHIME recipes."""
 
-from .model import build_co2_rhime_model, co2_prior_forward_mean
-from .runner import co2_model_input_names, run_rhime_co2
+from .co2_model import build_co2_rhime_model, co2_prior_forward_mean
```
brendan-m-murphy, line 3: Renaming `model.py` and `runner.py` removes the existing importable submodule paths without compatibility shims. Code using `openghg_inversions.rhime.co2.model` or `from openghg_inversions.rhime.co2 import runner` will now fail even though the package-level functions remain. Please retain forwarding modules, ideally with deprecation warnings.

brendan-m-murphy (commented):
This PR needs changes before it is ready. I found five actionable API, metadata, and validation issues in the current head.

openghg_inversions/rhime/co2/__init__.py
brendan-m-murphy: Thanks. We’re intentionally not adding compatibility shims here because `model.py` and `runner.py` were introduced only in unreleased development work and have never been part of a published package API. This rename is being completed before release, while the package-level public function imports remain unchanged. Forwarding modules and deprecation warnings would create a compatibility contract for paths that no released consumer could have depended on, so this PR keeps the final symmetric module layout directly.

openghg_inversions/rhime/co2/co2_o2_model.py
```diff
@@ -0,0 +1,302 @@
...
+            first and the shared retained-state dimension second.
+        o2_operator: O2 sensitivity with its distinct channel observation
+            dimension first and the same retained-state dimension second.
+
+    Returns:
+        The prepared joint sensitivity followed by prepared CO2 and O2
+        sensitivities. The joint value owns the shared zero-column mask; the
+        channel values retain their original row dimensions and labels.
+    """
+    co2_dim = str(co2_operator.dims[0])
```
brendan-m-murphy: This kind of stacking is usually done with the 'gather-concat' operations from array_ops.py. In this case, the new dimension would be 'species'.

openghg_inversions/rhime/co2/co2_o2_model.py
```diff
@@ -0,0 +1,302 @@
...
+                {joint_dim: range(nco2)}
+            ),
+            o2_operator.rename({o2_dim: joint_dim}).assign_coords(
+                {joint_dim: range(nco2, nco2 + no2)}
+            ),
+        ),
+        dim=joint_dim,
+    )
+    joint = prepare_linear_sensitivity(joint_operator, output_dim=joint_dim)
+    co2_sensitivity = joint.sensitivity.isel({joint_dim: slice(0, nco2)}).rename(
```
brendan-m-murphy: With the gathered coord, this would just be a selection operation

openghg_inversions/rhime/co2/co2_o2_model.py
```diff
@@ -0,0 +1,302 @@
...
+    )
+    o2_sensitivity = o2_sensitivity.assign_coords({o2_dim: o2_operator[o2_dim]})
+    return (
+        joint,
+        PreparedLinearSensitivity(co2_sensitivity, joint.removed, co2_dim),
+        PreparedLinearSensitivity(o2_sensitivity, joint.removed, o2_dim),
+    )
+
+
+def _resolve_activity(
```
brendan-m-murphy: Is this duplicate? It seems generic.

openghg_inversions/rhime/co2/co2_o2_model.py
```diff
@@ -0,0 +1,302 @@
...
+        {joint_dim: co2_dim}
+    )
+    co2_sensitivity = co2_sensitivity.assign_coords({co2_dim: co2_operator[co2_dim]})
+    o2_sensitivity = joint.sensitivity.isel({joint_dim: slice(nco2, None)}).rename(
+        {joint_dim: o2_dim}
+    )
+    o2_sensitivity = o2_sensitivity.assign_coords({o2_dim: o2_operator[o2_dim]})
+    return (
+        joint,
+        PreparedLinearSensitivity(co2_sensitivity, joint.removed, co2_dim),
```
brendan-m-murphy: The last two return values are ignored later; do we need them?

openghg_inversions/rhime/co2/co2_o2_model.py
```diff
@@ -0,0 +1,302 @@
...
+    Raises:
+        ValueError: If the channel operators cannot form one shared-state
+            sensitivity or ``state_activity`` does not align with that state.
+    """
+    joint_sensitivity, _, _ = _prepare_shared_state_channel_sensitivities(
+        co2_operator,
+        o2_operator,
+    )
+    activity = _resolve_activity(joint_sensitivity, state_activity)
+    prior_state = xr.where(
```
brendan-m-murphy: What is happening here? Add inline comments or make code self documenting.

openghg_inversions/rhime/co2/co2_o2_model.py
```diff
@@ -0,0 +1,302 @@
...
+        activity.active,
+        retained_prior.mean,
+        activity.fixed_value,
+    )
+    state_dim = retained_prior.state_dim
+    sensitivity_state_dim = str(joint_sensitivity.sensitivity.dims[1])
```
brendan-m-murphy: This is a common move, but is brittle. Attributes  (global or on a dimension coordinate) could label the state dimension explicitly. A helper function could retrieve a dimension by a common label eg. 'state_dim'. (This is similar to cf_xarray's standard name selection.)

Ideally, we would not need to check this at all in the modeling code.

openghg_inversions/rhime/co2/co2_o2_model.py
```diff
@@ -0,0 +1,302 @@
...
+    retained_state = prior_state.isel({state_dim: joint_sensitivity.retained_indices})
+    if sensitivity_state_dim != state_dim:
+        retained_state = retained_state.rename({state_dim: sensitivity_state_dim})
```
brendan-m-murphy: This is pragmatic, but should not be necessary. If our own data processing guarantees this, then it should be documented as a convention/schema. Ideally, we would use eg xarray-validate to actually encode a schema.

openghg_inversions/rhime/co2/co2_o2_model.py
```diff
@@ -0,0 +1,302 @@
...
+    Returns:
+        A registered PyMC model containing the shared state, separate channel
+        linear signals, coherent affine intercept, and joint Gaussian
+        likelihood.
+
+    Raises:
+        ValueError: If shared-state sensitivity preparation, activity
+            resolution, or registered coordinate alignment fails.
+    """
+    joint_sensitivity, co2_sensitivity, o2_sensitivity = (
```
brendan-m-murphy: Why is this called twice? Is the helper above not used?

brendan-m-murphy (commented):
Check for duplicate code or code that should be shared, reuse the common concat-gather pattern, clarify state dim validation/rename.

openghg_inversions/rhime/co2/co2_o2_model.py
brendan-m-murphy: Confirmed: this is a correctness failure, not only a brittle convention. At current head `2fb0e3c8`, I reproduced it with the gathered `(source, tracer_scope, region_in_source)` state and one column that is exactly zero in both channels. `prepare_linear_sensitivity()` renames the retained dimension and its MultiIndex levels; this code renames only the dimension on `prior_state`. The subsequent `xr.dot()` raises `AlignmentError: cannot align objects on coordinate 'state_retained' because of conflicting indexes`.

The simplest fix does not need attrs, a semantic-dimension resolver, or xarray-validate: evaluate the closure with the full gathered joint operator and the full resolved prior state. Structurally removed columns are exactly zero, so they contribute nothing and the retained-dimension lookup, positional selection, and conditional rename all disappear. Please add a regression combining a gathered MultiIndex state with a jointly zero column.

openghg_inversions/rhime/co2/co2_o2_preparation.py
brendan-m-murphy: The ordinary auxiliary-coordinate case is now canonicalized, but the gathered-state case remains. `pandas.MultiIndex.equals()` ignores level names, so an operator with the correct tuple values but stale names such as `bad_source` / `bad_scope` passes `_operator()`. I reproduced the result at `2fb0e3c8`: preparation then fails later in the intercept `xr.dot()` with an internal xarray `AlignmentError`, before the attempted canonicalization.

Please make the preparation boundary either require the operator MultiIndex level names to match the retained prior exactly or replace the complete operator state index from the prior after validating the values. Add the gathered stale-level-name case alongside the existing ordinary auxiliary-coordinate test so invalid independent inputs fail with the boundary's own `ValueError`.

openghg_inversions/rhime/co2/co2_o2_preparation.py
```diff
@@ -0,0 +1,577 @@
...
+
+    stacked = xr.concat(
+        (
+            channel(co2, start=0, tracer="co2", units=co2_units),
+            channel(o2, start=co2.size, tracer="o2", units=o2_units),
+        ),
+        dim="observation",
+        join="exact",
+    )
+    native_labels = (*co2.indexes[str(co2.dims[0])], *o2.indexes[str(o2.dims[0])])
```
brendan-m-murphy: **[P2] Reuse gathered coordinates and the existing CF MultiIndex codec**

This now round-trips the label types covered by the tests, but it makes a public scientific coordinate opaque and narrows the input contract unintentionally. `_axis` accepts any unique pandas/xarray index, whereas `_encode_observation_label` is a closed codec: common labels such as `Period`, `Timedelta`, and `bytes` now raise `TypeError`. Consumers also need a private decoder to recover the native observation labels.

The repository already has both halves of the standard solution:

- `concat_gather_data_arrays` / `select_gathered_data_array` construct and select the labelled ragged MultiIndex used by basis operators and flux components.
- `encode_cf_multiindexes` / `decode_cf_multiindexes` delegate durable MultiIndex serialization to cf-xarray's CF compression-by-gathering codec.

Please rename each channel's row axis to a common temporary observation dimension, gather under a `species` key level, and select CO2/O2 rows by that labelled level when native channel views are needed. At a durable artifact boundary, encode that named MultiIndex with the existing CF helper. This preserves native labels and unequal channel axes without positional slicing, opaque JSON strings, or a second serialization format.

If native observation indexes may themselves be MultiIndexes, extend/test the shared gather abstraction at that owning layer rather than nesting encoded tuples here. This should also let this PR remove its datetime/Timestamp expansion of the private `tagged_json_v1` covariance-artifact codec unless another artifact owner independently needs it.

openghg_inversions/rhime/co2/co2_o2_model.py
brendan-m-murphy: Agreed after following the gathered-index design through. This helper should return only the joint `PreparedLinearSensitivity`, indexed by `(species, channel_observation)`. The builder can then visibly select `species="co2"` and `species="o2"` with `select_gathered_data_array` immediately before constructing the two named channel signals. The prior-forward preservation calculation needs only the same joint operator and therefore does not need two discarded channel objects. This removes the positional slicing and makes the scientific branch point explicit in the recipe.

openghg_inversions/rhime/co2/co2_o2_model.py
brendan-m-murphy: These are currently two independent API uses, not two calls in one runner execution: the builder prepares the sampled symbolic graph, while `co2_o2_prior_forward_mean` eagerly evaluates `fixed_prior_contribution + H @ resolved_prior_state` for tests/diagnostics. The runner calls only the builder.

That role should be stated without the vague term “closure.” Since the function returns an evaluation rather than comparing or raising, I suggest a verb such as `evaluate_co2_o2_prior_forward_mean`, documented as: “Evaluate the reduced affine model at the resolved prior state so preparation/tests can verify that the expected prior-forward contribution is preserved.” The actual check is then the caller’s comparison with the expected prior-forward field. If it is instead named `check_...`, it should accept that expected field and perform the comparison itself.

With the single gathered return proposed above, both APIs share one clear preparation primitive: the diagnostic dots the gathered operator directly with the resolved prior state; the builder selects the two species and constructs the sampled channel signals.

openghg_inversions/rhime/co2/co2_o2_model.py
brendan-m-murphy: Agreed. This should use `concat_gather_data_arrays` with `key_dim="species"`, one common ragged observation level, and one joint output dimension. The preparation helper should return that single gathered `PreparedLinearSensitivity`; it does not need to construct and return two positional channel slices. This is also consistent with the repository's existing CF compression-by-gathering serialization boundary. Follow-up design guidance is tracked in #638.

openghg_inversions/rhime/co2/co2_o2_model.py
brendan-m-murphy: Agreed, and ordinary xarray selection is the clearest implementation: `joint.sensitivity.sel(species="co2", drop=True)` and the corresponding O2 selection. Xarray removes the selected MultiIndex level and restores the remaining observation index. The resulting channel dimensions may then need explicit CO2/O2 names because PyMC's coordinate registry cannot register unequal channel axes under one global dimension name. My earlier recommendation to use `select_gathered_data_array` here was unnecessarily defensive; that helper's structural validation is useful at a less-trusted boundary, not immediately after our own gather operation.

openghg_inversions/rhime/co2/co2_o2_model.py
brendan-m-murphy: Agreed. This one-line wrapper adds no model-specific ownership: it only forwards `sensitivity.removed` to the generic `resolve_state_activity`. Delete it and call the generic function directly in the two independent APIs.

openghg_inversions/rhime/co2/co2_o2_model.py
brendan-m-murphy: Correction to my earlier reply: the single gathered return remains the right design, but the builder should select its species with ordinary `.sel(species=..., drop=True)`, not the stricter `select_gathered_data_array` helper. The gathered object is package-created and can be trusted. The model then visibly constructs the two channel signals, while the prior-forward preservation evaluation uses the joint gathered operator directly.

openghg_inversions/rhime/co2/co2_o2_model.py
brendan-m-murphy: Agreed. This should state the policy directly, for example `resolved_prior_state = retained_prior.mean.where(activity.active, activity.fixed_value)`, with a short explanation that active states are evaluated at their arithmetic prior means and inactive states at their exact fixed values. The current generic `xr.where` ordering makes the scientific meaning harder to recover.

openghg_inversions/rhime/co2/co2_o2_model.py
brendan-m-murphy: Agreed that this rename should disappear from the model recipe. For the prior-forward preservation evaluation, dotting the full gathered joint operator with the full resolved prior state removes the retained-dimension inference and conditional rename entirely; structurally removed columns are exactly zero. I would not add xarray-validate or a semantic-dimension framework in this concrete PR. The broader convention—when gathered axes are selected into components versus retained for alignment—is now tracked in #638.

openghg_inversions/rhime/co2/co2_o2_model.py
brendan-m-murphy: Additional clarification after separating the two questions: gathered representation and semantic dimension ownership are orthogonal. Issue #638 now records that gather/select operations receive concrete dimension names explicitly and must not infer output or state roles from position, suffixes, or MultiIndex levels. PreparedLinearSensitivity, or the preparation owner constructing it, remains authoritative for those roles.

For this PR, dotting the full gathered operator with the full resolved state is still the right local fix because it removes this lookup altogether. That does not make attrs or an xarray-validate schema necessary in the concrete model.

openghg_inversions/rhime/co2/co2_o2_model.py
brendan-m-murphy: One correction to the proposed follow-up: returning only the joint PreparedLinearSensitivity does not necessarily mean that the builder must select CO2 and O2 into channel-specific prepared views.

If both species share one state and differ only through a known ratio or scaling, preparation can align a species/state factor to the gathered observations, with ones for CO2 and the ratio for O2, and multiply the joint sensitivity before graph construction. This is analogous to the SigmaAlignment pattern, but keyed by species rather than site/period. The model can then apply the single gathered sensitivity to the shared state once.

In this PR the O2 operator already embeds the signed ratios, so even that extra factor is unnecessary: one joint linear application may be sufficient. Species selection remains appropriate only if a channel owns distinct graph structure, state, prior, or a separately required contribution. I added this distinction to #638; #414 tracks the broader prepared component-data pattern and #622 the narrower reusable observation-partition mapping.

brendan-m-murphy: The current head improves this from three returns to two, but the original readability issue remains: the builder now writes `joint_sensitivity, _ = _prepare_shared_state_sensitivity(...)`, while the preservation evaluator needs both values.

The simplest version is for the helper to do only what its gathered-coordinate name says and return one `joint_operator` DataArray. Each independent API can then state its own next step explicitly:

```python
joint_operator = _gather_co2_o2_operator(...)
joint_sensitivity = prepare_linear_sensitivity(joint_operator, output_dim=output_dim)
```

That one-line duplication keeps the two roles visible: the builder needs the prepared sensitivity for graph construction; `evaluate_co2_o2_prior_forward_mean` additionally uses the full operator to verify that the expected prior-forward contribution is preserved. No result wrapper or extra abstraction is needed. Please also put that preservation purpose in the evaluator's opening docstring rather than describing only the calculation.

openghg_inversions/rhime/co2/co2_o2_model.py
```diff
@@ -0,0 +1,229 @@
...
+            data_name="co2_o2_operator",
+            output_name="co2_o2_flux_contribution",
+        )
+        modelled = add_coherent_affine_component(
+            fixed_prior_contribution,
+            joint_signal,
+            output_name="modelled_concentration",
+        )
+
+        observed = add_model_data(observations, "observed_concentration")
```
brendan-m-murphy: **[P2] Keep fixed-error likelihood ownership explicit in this PR**

The latest simplification has moved observation/error registration and aggregation-error plumbing back into four bare backend calls in the concrete builder. The reader has to reconstruct the single concept “joint fixed-error likelihood” from `add_model_data`, `add_aggregation_error_data`, and the tensor-level distribution helper.

For PR #632, please keep one ordinary private function such as `_add_co2_o2_fixed_error_likelihood(observations, modelled, independent_error_sd=..., aggregation_error=..., output_dim=...)` that owns those registrations and calls `add_gaussian_observation_likelihood`. This is a small local readability component, not a new framework or permanent competing abstraction; it lets the builder read as state → gathered affine signal → joint likelihood.

The reusable end state is tracked separately in #639: generalize the existing additive-sigma likelihood so it cleanly supports this fixed-error-only mode, scalar or grouped error conveniences at an appropriate boundary, and **remove `minimum_error` from additive sigma** because that floor belongs only to the historical PEFO=True/False likelihoods. Once that generic component covers the required semantics, replace this local wrapper rather than retaining both implementations. Do not enlarge this PR with that cross-model change.

Issue #17, opened 2023-07-04
Speed improvements for HBMCMC - extra info


aliceramsden:
The table below contains info on the runtimes for various HBMCMC setups, from the code at commit https://github.com/openghg/openghg_inversions/commit/8d2ecf11f93a376e83422ca169f4951869df314e. 

No filtering was applied to the 4-hourly averaged observations.

Trace parameters were set at:
nit = 250000
burn = 50000
tune = 125000

Ideally, the largest model runs (20 sites, 500 basis functions) should be brought down to 8 hours, but this target can definitely be adjusted if needed.

<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=OneNote.File>
<meta name=Generator content="Microsoft OneNote 15">
</head>

<body lang=en-GB style='font-family:Calibri;font-size:11.0pt'>

<div style='direction:ltr'>

Inputs | Num. obs | Run ID | Run time | Memory usage
-- | -- | -- | -- | --
n_basis 50   n_sites = 10 | 1741 | 5411818 | 2H 15mins | 16 GB
n_basis = 100   n_sites = 10 | 1741 | 5411819 | 3H 15mins | 16 GB
n_basis = 500   n_sites = 10 | 1741 | 5418801 | 5H 50mins | 20 GB
  |   |   |   |  
n_basis = 100   n_sites = 15 | 2626 | 5411820 | 4H 45mins | 23 GB
n_basis = 500   n_sites = 15 | 2626 | 5418813 | 12H 50mins | 26 GB
  |   |   |   |  
n_basis = 100   n_sites = 20 | 3377 | 5411821 | 6H 30mins | 30 GB
n_basis = 200   n_sites = 20 | 3377 | 5460436 | 11H 30mins | 30 GB
n_basis = 500   n_sites = 20 | 3377 | 5418855 | 16H 50mins | 31 GB

</div>

</body>

</html>

This links to a larger task in the project overview: https://github.com/ACRG-Bristol/projects/issues/30

brendan-m-murphy:
@aliceramsden are these runs based off the paris no filtering .ini files in the paris_total_methane_inversion directory on BP1?

Do you know if these .ini files just depend on the PR that I reverted? I should be able to add those changes back in soon.

aliceramsden:
[...]

PR #635, opened 2026-08-21
OPE-76: Add explicit CO₂ outer-state treatment

brendan-m-murphy:
## Summary

Adds explicit fixed, Gaussian-marginalized, and inferred CO₂ outer-state treatments to the public OGI model. Optional cross-sector collapse preserves member labels; `basis_group="outer"` is the explicit partition contract and overlap is rejected to prevent double counting.

Fixed and inferred treatments retain the ordinary full-state contract (`outer_flux_scaling`, `outer_sensitivity`, `outer_flux_contribution`). Fixed outer states use the shared `StateActivity` / `add_state_vector` path and create no sampled active latent. Marginalized outer uncertainty is retained as an observation-by-state factor, so structured aggregation errors do not become dense. Zero-H inferred states are fixed at one and the same resolved activity drives graph construction and provenance.

CO₂-only now uses the prepared-linear component path: `flux_scaling`, pruned backend `co2_sensitivity`, `co2_flux_contribution`, and the complete likelihood mean `modelled_concentration`. CO₂/O₂ uses the same sensitivity vocabulary and retains the linked helper only for its genuinely shared flux state.

Atmospheric boundary conditions (`mu_bc`) and outer flux (`outer_flux_contribution`) are separate model components. Grouping them as a baseline is a reporting concern only.

Marginalization is outer-scoped here for context flux outside the scored inversion domain. The API is motivated by PARIS Verification Games with known synthetic flux and transport truth; current evidence selects inferred outer states. General marginalization of weak states in any basis group remains a future shared state-disposition concern.

## Checklist

- [ ] Closes #xxxx (tracked in Linear as OPE-76; no GitHub issue)
- [x] Tests added and passing
- [x] Public API and user documentation added
- [x] Towncrier news fragment added
- [x] No new requirements

## Stack and validation

PR #635 is rebased directly onto `devel`, including merged PR #632. Current head: `9ce4f75`.

- `uv run pytest tests/test_co2*.py tests/test_rhime_co2_o2*.py tests/models/test_components.py tests/models/test_state_activity.py -q` — 107 passed
- A structural-zero regression proves that the public state remains complete, the backend sensitivity is pruned, and the contribution equals the original full sensitivity multiplied by the full state
- A custom observation-dimension regression proves that outer graph/covariance paths consume the retained `PreparedLinearSensitivity` roles
- 50,000-observation regression retains a `50,000 × 2` marginalized factor and never constructs an observation-square covariance
- None/diagonal/low-rank aggregation errors remain structured; already-dense input remains dense
- Singular PSD outer covariance is covered
- Conflicting labeled observation indexes are rejected by existing xarray/CoordRegistry paths; no outer-specific custom-input validator remains
- Coordinate-validation policy is documented in `docs/development/validation_and_xarray.rst`: xarray first, then `CoordRegistry`, with bespoke checks requiring a documented gap and focused regression
- Ruff and `git diff --check` pass
- All review threads have implementation evidence and are resolved

## Downstream evidence

- [openghg/test_inversions#15](https://github.com/openghg/test_inversions/pull/15) is merged and registers the bounded MHD/TAC CO₂ smoke case; [#17](https://github.com/openghg/test_inversions/pull/17) updates it to the final model-variable names.
- Final-contract direct VG evidence used OGI `099a4d7` and VG `8a740d2`: 96 MHD/TAC observations, 8 inner + 24 inferred outer + 4 boundary states, four-chain NumPyro NUTS with 1,000 tune + 1,000 draws, zero divergences, max R-hat 1.00, minimum bulk ESS 3599, exact 0.0 ppm fixed-one closure, and 875,520 KiB peak process RSS. Commit `c7571cf` only deletes a redundant pre-graph index check and does not change the model graph.
- The larger reviewed January run used 640 observations across MHD/TAC/JFJ/CBW with the same 36 states. It passed with zero divergences, max R-hat 1.01, minimum bulk ESS 3295, exact 0.0 ppm fixed-one closure, and 922,276 KiB peak batch MaxRSS. The reviewed adapted-InTEM input contains four sites; no compatible fifth-site artifact currently exists.
- [openghg/verification-games#48](https://github.com/openghg/verification-games/pull/48) contains the parameterized runner and reporting-only baseline grouping. [#49](https://github.com/openghg/verification-games/pull/49) updates the merged CO₂/O₂ consumer to final sensitivity names.

The registered smoke remains a dependency/contract check; the longer VG runs are the scientific recovery evidence.

Linear: OPE-76


openghg_inversions/rhime/co2/outer_regions.py
```diff
@@ -0,0 +1,478 @@
...
+
+
+def add_outer_observation_covariance(
+    aggregation_error: AggregationError,
+    treatment: OuterRegionTreatment,
+) -> AggregationError:
+    """Add exact Gaussian outer-state covariance to fixed observation error."""
+    if treatment.mode != "marginalized":
+        return aggregation_error
+    outer = treatment.observation_covariance
```
brendan-m-murphy: Marginalized outer states always reach this point as an `N_observation × N_observation` covariance, and `.values` eagerly materializes it before every aggregation-error representation is converted to dense. For 50,000 observations that is about 20 GB for one float64 matrix, even when the existing error is diagonal or low-rank. Please retain the outer contribution as a factor such as `H @ chol(Sigma)` and combine it with the existing representation without forcing dense covariance.

openghg_inversions/rhime/co2/co2_model.py
```diff
@@ -124,6 +134,32 @@ def _fixed_mismatch_array(
...
+        return
+    state_dim = next(str(dim) for dim in flux_sensitivity.dims if dim != "nmeasure")
+    if "basis_group" in flux_sensitivity.coords:
+        groups = np.asarray(flux_sensitivity["basis_group"].values).astype(str)
+        if (groups == "outer").any():
+            raise ValueError(
+                "CO2 flux_sensitivity still contains basis_group='outer' states while an "
+                "outer_treatment is supplied; remove them to avoid double counting."
+            )
+    if state_dim in outer_treatment.state_metadata.dims:
```
brendan-m-murphy: This overlap check only runs when the inner state dimension name also exists in `state_metadata`. With the expected `region`/`outer_region` split, the same semantic state can therefore appear in both designs (without `basis_group`) and be sampled and added twice, causing non-identifiability and double counting. Please compare canonical semantic state identity across dimensions, or require explicit metadata proving the partitions are disjoint; comparing bare local labels alone would reject valid partition-local reuse.

openghg_inversions/rhime/co2/outer_regions.py
```diff
@@ -0,0 +1,478 @@
...
+    sensitivity = sensitivity.transpose(observation_dim, state_dim)
+    zero_covariance = _zero_observation_covariance(
+        sensitivity,
+        observation_dim=observation_dim,
+    ).rename("outer_observation_covariance")
+    zero_contribution = xr.zeros_like(sensitivity.isel({state_dim: 0}, drop=True)).rename(
+        "fixed_outer_contribution"
+    )
+
+    metadata_dim = next(iter(metadata.dims))
```
brendan-m-murphy, line 392: `activity` is set solely from the treatment mode before inferred-state zero-sensitivity resolution. A zero-H inferred state is later fixed at one, but `state_metadata.activity` still reports it as active. Please derive this metadata from the same resolved activity policy used to construct `x_outer` so stored scientific provenance matches the actual sampled state.

brendan-m-murphy (commented):
PR #632 needs changes before this stacked PR can be ready; I posted the CO₂/O₂ findings on #632. Separately, the current OPE-76 head has the three outer-state issues below.

brendan-m-murphy (commented):
Two model-boundary questions remain after re-review. Both concern whether state disposition should be a shared linear/flux-state concept rather than an outer-region-specific path.

openghg_inversions/rhime/co2/outer_regions.py
```diff
@@ -0,0 +1,589 @@
...
+        state_dim=state_dim,
+        name="prior_mean",
+    )
+    covariance, _ = _aligned_state_covariance(
+        prior_covariance,
+        sensitivity,
+        state_dim=state_dim,
+    )
+
+    if mode == "marginalized":
```
brendan-m-murphy, line 441: What concrete scientific use requires an outer-specific `marginalized` mode here? There is a genuine general need to marginalize weakly observed basis regions—for example remote islands that cannot sensibly be merged with nearby land—but that is a disposition of a flux state, not inherently an outer-region operation. The current PR documents the Gaussian calculation but not why this capability belongs on the separate outer path or which production input selects it. Please document the intended use and explain the boundary; otherwise, could this wait for a shared state-marginalization operation usable by inner and outer basis groups alike?

openghg_inversions/rhime/co2/outer_regions.py
```diff
@@ -0,0 +1,589 @@
...
+                metadata_dim,
+                np.full(metadata.sizes[metadata_dim], "outer", dtype=object),
+            )
+        )
+    metadata["treatment"] = (
+        metadata_dim,
+        np.full(metadata.sizes[metadata_dim], mode, dtype=object),
+    )
+
+    if mode == "fixed":
```
brendan-m-murphy, line 405: Why is fixed outer state handled differently from a fixed retained flux state? `StateActivity` plus `add_correlated_lognormal_state_with_activity` already keeps fixed entries in the full state vector and applies the ordinary linear sensitivity. This branch instead removes the state and precomputes `H_outer @ fixed_scale` as baseline data. The prediction is algebraically equivalent, but the state/output/provenance contract differs solely because the basis group is outer. Please route fixing through one shared linear-component/state-vector operation (not necessarily `add_linear_component` itself) so inner and outer states use the same mechanism.

openghg_inversions/rhime/co2/outer_regions.py
brendan-m-murphy: Addressed. Marginalized outer uncertainty is now retained as the observation-by-state factor `H @ sqrt(Sigma)`. None, diagonal, and low-rank aggregation errors remain structured; an observation-square covariance is formed only when the caller already supplied dense error. The regression uses 50,000 observations and two outer states and asserts a 50,000 x 2 factor. The affected 231-test selection passes.

openghg_inversions/rhime/co2/co2_model.py
brendan-m-murphy: Addressed with an explicit basis-group partition contract. Inner sensitivity must carry complete state-aligned `basis_group` metadata and cannot contain `outer` when an outer treatment is supplied; every outer treatment state must be labelled `basis_group=outer`. This works across different dimension names without comparing partition-local labels, and the rejection paths are covered by tests.

openghg_inversions/rhime/co2/outer_regions.py
brendan-m-murphy: Addressed. Preparation now resolves activity once and `_set_metadata_activity` derives provenance from that exact `ResolvedStateActivity`. Zero-H inferred states are therefore reported inactive and fixed at one in `x_outer`; direct and collapsed-state mappings are both covered.

openghg_inversions/rhime/co2/outer_regions.py
brendan-m-murphy: Addressed by documenting the scientific boundary. This outer-scoped mode is for flux outside the scored inversion domain when its Gaussian prior uncertainty should propagate without sampling those context states. Upstream preparation selects the mode explicitly. The docs also state that general marginalization of weak states in any basis group belongs in a future shared state-disposition API. Atmospheric BC and outer flux now remain separate model components; grouping them as a baseline is explicitly a reporting concern.

openghg_inversions/rhime/co2/outer_regions.py
brendan-m-murphy: Addressed in `8430162`. Fixed outer treatment now resolves `StateActivity(active=False, fixed_value=scale)`, retains the full sensitivity, and builds `x_outer` with the shared `add_state_vector` path before applying ordinary `h_outer @ x_outer` as `mu_outer`. It no longer precomputes outer flux into baseline data. Tests assert the full fixed state, linear contribution, and absence of a sampled active latent.

openghg_inversions/rhime/co2/outer_regions.py
```diff
@@ -0,0 +1,579 @@
...
+    if isinstance(outer_sensitivity, CollapsedOuterStates):
+        sensitivity = outer_sensitivity.sensitivity
+        metadata = outer_sensitivity.members.copy()
+    else:
+        sensitivity = outer_sensitivity
+        state_dim = _matrix_state_dim(sensitivity, observation_dim)
+        metadata = _state_metadata(sensitivity, state_dim)
+
+    state_dim = _matrix_state_dim(sensitivity, observation_dim)
+    sensitivity = _label_outer_basis_group(
```
brendan-m-murphy, line 387: [P1] Require observation labels before retaining this treatment

`_matrix_state_dim` only requires state labels, so this path also accepts an outer sensitivity whose `nmeasure` dimension has no coordinate. I reproduced the model successfully building with labeled observations `[a, b]` and an unlabeled outer design; `mu_outer` and the marginalized low-rank factor were attached positionally. Because there is no outer observation coordinate for `CoordRegistry` to compare, reordered rows can silently produce scientifically incorrect means and covariance. Please require a unique observation coordinate here and enforce exact alignment with the model observations at the builder boundary.

openghg_inversions/rhime/co2/outer_regions.py
```diff
@@ -0,0 +1,579 @@
...
+
+@dataclass(frozen=True)
+class OuterRegionTreatment:
+    """Mutually exclusive outer-region mean, covariance, or sampled state.
+
+    Fixed and inferred treatments retain the ordinary sensitivity/state-vector
+    contract. A marginalized treatment instead exposes its Gaussian mean
+    contribution and observation-space factor, without sampling outer states.
+
+    Marginalization is outer-scoped here because these states describe flux
```
brendan-m-murphy, line 55: It might be worth saying that this is needed for the "PARIS verification games". I'm not sure if it's useful more widely. In the verification games, observations are generated from a known flux and transport model, so we can actually "score" the posterior flux against a known truth.

openghg_inversions/rhime/co2/co2_model.py
```diff
@@ -152,10 +230,22 @@ def build_co2_rhime_model(
...
+    _validate_outer_observation_alignment(observations, outer_treatment)
+    _reject_outer_state_double_counting(flux_sensitivity, outer_treatment)
+    aggregation_error = (
+        aggregation_error
+        if outer_treatment is None
+        else add_outer_observation_covariance(aggregation_error, outer_treatment)
+    )
 
     with registered_model() as model:
         h = add_model_data(flux_sensitivity.transpose("nmeasure", ...), "hx")
```
brendan-m-murphy: Please replace this manual state/data/`pt.dot` sequence with the shared prepared-linear path introduced in PR #632. Prepare `flux_sensitivity` with `prepare_linear_sensitivity` before entering the model, resolve activity from `prepared.removed`, construct the correlated full state, and apply the retained sensitivity with `add_linked_linear_component` (or `apply_linear_sensitivity`). After composing the CO2, boundary, and outer linear contributions, use `add_coherent_affine_component(..., output_name="modelled_concentration")` for the fixed intercept. PR #632 explicitly left applying that affine component to the CO2-only model for follow-up, and this PR already changes this exact composition boundary.

Please add one structural-zero regression proving that the public state remains complete, the backend sensitivity retains only nonzero columns, and the resulting CO2 contribution still equals the original full sensitivity multiplied by the full state. This is component consistency, not compatibility preservation; the model is unreleased, so its current `x`/`hx`/`mu*` names should not constrain the refactor.

openghg_inversions/rhime/co2/outer_regions.py
```diff
@@ -0,0 +1,585 @@
...
+    output_name: str = "mu_outer",
+    observation_dim: str = "nmeasure",
+) -> Any:
+    """Add a fixed or inferred outer state through the ordinary linear path."""
+    if treatment.mode not in ("fixed", "inferred"):
+        raise ValueError("Only fixed and inferred outer treatments retain a state vector.")
+    if treatment.sensitivity is None or treatment.resolved_activity is None:
+        raise ValueError(f"Outer-region treatment {treatment.mode!r} is missing its state contract.")
+    sensitivity = treatment.sensitivity.transpose(observation_dim, ...)
+    state_dim = _matrix_state_dim(sensitivity, observation_dim)
```
brendan-m-murphy: [P2] Carry the resolved dimension roles into the outer component

Preparation already validates `observation_dim` and resolves the one `state_dim`, but `OuterRegionTreatment` drops both; this path then accepts a second observation-dimension authority and reruns `_matrix_state_dim`. Please have the treatment retain the prepared sensitivity contract (if the new #632 path is applied here too), or minimally retain `observation_dim` and `state_dim`, and consume that information here and in the observation/covariance paths. Keeping sole-other-dimension discovery inside the strict 2-D preparation boundary is fine; repeating it during graph construction is the ad hoc inference discussed in #638. This is complementary to the prepared-linear refactor requested on `co2_model.py`: once prepared, the component should not rediscover its roles.

brendan-m-murphy (commented):
One additional dimension-contract point after reviewing the new prepared-linear and naming comments.

openghg_inversions/rhime/co2/outer_regions.py
brendan-m-murphy: Agreed. I would make the Verification Games motivation explicit without presenting this as a generally recommended outer-state policy. Suggested wording: “This mode is currently required for the PARIS Verification Games, where observations are generated from a known flux and transport model and posterior flux can therefore be scored against known truth. Its usefulness outside that setting has not yet been established.” The existing distinction immediately below—general weak-state marginalization belongs in a shared state-disposition API—should remain.

openghg_inversions/rhime/co2/outer_regions.py
brendan-m-murphy: Deliberately narrowed in 099a4d7 after maintainer direction. We do not validate duplicate or unlabeled custom arrays. The builder performs one cheap composition check only when both observations and the prepared outer sensitivity carry explicit indexes; conflicting values, order, or index-level names are rejected. Canonical gathered inputs already establish unique per-site times, while unlabeled direct callers own positional correctness. The reordered-explicit-index regression remains.

openghg_inversions/rhime/co2/outer_regions.py
brendan-m-murphy: Documented in 099a4d7. The public treatment docs now state that this API is motivated by the PARIS Verification Games, whose synthetic observations have known flux and transport truth and therefore permit posterior-flux scoring. The wording also records that current evidence uses inferred outer states, that marginalized mode is not an established production/default choice, and that general weak-state marginalization belongs in a shared state-disposition API.

openghg_inversions/rhime/co2/outer_regions.py
brendan-m-murphy: Addressed in f0d2420. OuterRegionTreatment now retains PreparedLinearSensitivity as the sole observation/state-role authority. Fixed and inferred graph construction, marginalized covariance composition, activity resolution, and namespacing all consume that prepared contract; add_outer_state_component no longer accepts a second observation-dimension authority. A custom-output-dimension regression covers the non-default path.

openghg_inversions/rhime/co2/co2_model.py
brendan-m-murphy: Addressed in f0d2420. CO2 prepares the sensitivity once, resolves activity from the prepared full-state mapping, builds the complete flux_scaling state, applies the pruned backend through apply_linear_sensitivity, and composes the complete modelled_concentration with add_coherent_affine_component. The structural-zero regression proves the public state stays complete, co2_sensitivity drops only the zero column, the fixed value is one, and co2_flux_contribution equals the original full sensitivity times the full state. apply_linear_sensitivity is intentional for CO2-only; the linked helper remains limited to CO2/O2 sharing one flux state.

openghg_inversions/rhime/co2/outer_regions.py
brendan-m-murphy: Follow-up in c7571cf: the outer-specific explicit-index helper has now been deleted entirely. The existing xarray exact-alignment and CoordRegistry paths reject conflicting labeled indexes in fixed, marginalized, and inferred modes; the regression exercises those shared paths. Unlabeled custom arrays remain caller-owned by design.

brendan-m-murphy (commented):
Final review at `9ce4f753`: no remaining blocking findings.

The follow-up commits address the review points: outer observation composition now relies on the documented shared xarray/`CoordRegistry` alignment policy; CO2 and outer sensitivities use the prepared-linear contract with full public states and structurally pruned backend data; the coherent affine component owns the complete model mean; dimension roles are retained rather than rediscovered; and CO2/CO2-O2 use the agreed `*_sensitivity`, `flux_scaling`, contribution, and `modelled_concentration` vocabulary.

Validation reviewed:
- focused CO2, outer-region, CO2/O2 preparation/model/runner suite: 58 passed locally;
- Ruff, borrowed-reference typing, Python 3.10 tests against OpenGHG 0.18.0, 0.19.0, and OpenGHG devel: all passed;
- CI Gate: passed;
- `git diff --check`: passed.

From the code and scientific-boundary perspective, this is ready to merge.

openghg_inversions/rhime/co2/co2_model.py
```diff
@@ -12,36 +12,43 @@
...
-    """Resolve activity once in the retained-state coordinate order."""
-    return resolve_state_activity(
-        detect_zero_sensitivity(flux_sensitivity),
-        state_activity,
-    )
+    """Resolve activity once from the prepared full-state mapping."""
+    return resolve_state_activity(prepared_sensitivity.removed, state_activity)
 
 
 def co2_prior_forward_mean(
```
brendan-m-murphy: This function is only called in tests. What is its purpose? (Also, it does not have a full google-style docstring)

openghg_inversions/rhime/co2/co2_model.py
```diff
@@ -98,7 +105,7 @@ def _add_co2_retained_state(
     return add_correlated_lognormal_state_with_activity(
```
brendan-m-murphy: In the co2_o2 model, this is done in the model building function. Adopt that pattern.

openghg_inversions/rhime/co2/co2_model.py
```diff
@@ -124,6 +131,39 @@ def _fixed_mismatch_array(
...
+    if "basis_group" not in outer_metadata:
+        raise ValueError("outer_treatment requires basis_group metadata proving every state is outer.")
+    outer_groups = outer_metadata["basis_group"]
+    if outer_groups.ndim != 1 or bool(outer_groups.isnull().any().compute().item()):
+        raise ValueError("outer_treatment basis_group metadata must be complete and one-dimensional.")
+    if not bool((outer_groups == "outer").all().compute().item()):
+        raise ValueError("outer_treatment basis_group metadata must label every state 'outer'.")
+
+
 def build_co2_rhime_model(
```
brendan-m-murphy: This needs a full google-style docstring.

openghg_inversions/rhime/co2/co2_model.py
```diff
@@ -12,36 +12,43 @@
...
 from openghg_inversions.sigma import SigmaAlignment
 
+from .outer_regions import (
+    OuterRegionTreatment,
+    add_outer_state_component,
+    add_outer_observation_covariance,
+)
+
 
 def _resolve_co2_state_activity(
```
brendan-m-murphy: This wrapper adds nothing. Remove it.

openghg_inversions/rhime/co2/co2_model.py
```diff
@@ -74,16 +82,15 @@ def co2_prior_forward_mean(
 
 
 def _add_co2_retained_state(
```
brendan-m-murphy: This is done in the model building function for the co2_o2 model. Also, the covariance checks should be scrutinized; they are not present in the co2_o2 model.

openghg_inversions/rhime/co2/co2_model.py
```diff
@@ -124,6 +131,39 @@ def _fixed_mismatch_array(
     ).rename("fixed_model_mismatch")
 
 
+def _reject_outer_state_double_counting(
```
brendan-m-murphy: This looks like a data prep concern, not a modelling concern. Also, is it is checking something that is guaranteed by our own data prep? If so, remove it unless a failure would be totally inscrutable. I do not want to over-engineer custom data validation.

brendan-m-murphy (commented):
The co2_o2 model when through considerable review. Consider it's patterns as the standard to match. 

Also, if `_fixed_mistmatch_array` is meant to be extracted into a model component later, that should be documented in its docstring.

PR #641, opened 2026-08-24
Add configurable additive-sigma Gaussian likelihood

brendan-m-murphy:
## Summary of changes

- add one complete reusable additive-sigma Gaussian likelihood whose independent variance is `observation_error**2 + fixed_model_mismatch**2 + sigma**2`
- retain `min_error` as an optional floor on total marginal standard deviation; the standard configuration defaults it to `0.0`
- use `SigmaAlignment | None` as the complete component's inferred-sigma switch, with a HalfNormal prior as the installed adapter's default
- avoid a separate error-state constructor and store observations only under the canonical observed variable `y`
- reuse the likelihood in the CO2 and CO2/O2 recipes introduced by #635 while preserving their structured aggregation covariance
- let `run_hbmcmc.py` select the likelihood with `likelihood = "additive_sigma"`, explicitly forcing aggregation error off
- add the preferred `additive_sigma_prior` INI key, with `sigprior` as a compatibility fallback
- accept labelled HalfNormal scales such as `{"MHD": 5.0, "TAC": 2.0}` and align them to retained sites independently of dictionary order
- document the recommended HalfNormal-without-floor starting point and `E[sigma] = prior_scale * sqrt(2 / pi)`

This addresses the additive-likelihood replacement investigated in #637 and is related to #639. Unlike #639's original wording, this version deliberately retains the optional minimum-error floor following user feedback.

- [x] Rebased onto the CO2-family changes from #635
- [x] Tests added and passing
- [x] No new requirements

## Validation

- `uv run pytest tests/test_run_hbmcmc_shim.py tests/models/test_likelihoods.py` — 47 passed
- broader RHIME, shim, likelihood, and CO2 selection — 315 passed; three unrelated environment/data-stack failures (`h5netcdf` dimension-scale read and two OpenGHG `long_name` expectations)
- Ruff on all changed Python paths — passed
- `git diff --check` — passed
- BP1 `ch4_hbmcmc_additive_sigma_2sites` against commit `93f7e0b29e08fd9a09b6bf169f84948ada61be8f` — SLURM job `18693385`, exit 0, run status `OK`, convergence output `Passed`, one divergence
  - HalfNormal prior scale: 5.0 ppb; `min_error = 0.0`
  - posterior sigma MHD: mean 4.76 ppb, median 4.75 ppb, 5–95% interval 4.21–5.36 ppb
  - posterior sigma TAC: mean 1.35 ppb, median 1.28 ppb, 5–95% interval 0.10–2.84 ppb
- direct fixed-seed CH4 prototype against commit `2636fd87ffead0a6f90e9937acfec1db74c21b3d` reconstructed the 260-observation/259-state model from the completed real-data run without using the test_inversions registry:
  - default diagonal NumPyro: zero divergences, 186 mean steps, max R-hat 1.020, minimum bulk ESS 393
  - dense mass: zero divergences but every draw hit depth 10 (1,023 steps), 6x slower, max R-hat 1.035, minimum bulk ESS 106
  - diagonal `target_accept=0.75`: clean convergence but 222 mean steps, so it does not shorten trajectories
  - diagonal `target_accept=0.70`: 148 mean steps but max R-hat 1.193 and minimum bulk ESS 14; `0.60` produced 52 divergences
  - conclusion: retain NumPyro's default diagonal sampler; the long trajectories reflect the sharp, correlated flux/BC posterior rather than an additive-sigma implementation defect

### DIFF
--- a/docs/usage/getting_started.rst
+++ b/docs/usage/getting_started.rst
@@ -379,7 +379,10 @@ The following file, ``my_hbmcmc_inputs.ini`` can be used to run an
 
    xprior   = {"pdf":"lognormal", "stdev":1}  ; lognormal with mean = 1, stdev = 1
    bcprior  = {"pdf":"truncatednormal", "mu":1.0, "sigma":0.02}  ; truncated normal with mean = 1, stdev 0.02
-   sigprior = {"pdf":"uniform", "lower":0.5, "upper":10}
+   sigprior = {"pdf":"halfnormal", "sigma":5.0}  ; additive sigma, in observation units
+   ; For the historical pollution-event likelihood instead, omit `likelihood`
+   ; below and use a dimensionless fractional prior such as:
+   ; sigprior = {"pdf":"uniform", "lower":0.0, "upper":0.2}
    ;add_offset = False
    ;offsetprior = {"pdf": "normal"}
    ;offset_args = {"drop_first": False}  ;; set to True if you want first site to have offset 0.0
@@ -413,6 +416,9 @@ The following file, ``my_hbmcmc_inputs.ini`` can be used to run an
    nchain = 4
 
    [MCMC.OPTIONS]
+   ; Select response-independent absolute model error. Omit this option to use
+   ; the historical pollution-event likelihood.
+   likelihood = "additive_sigma"
    ; averaging_error (bool): Add variability in averaging period to the measurement error (Note: currently this
    ;                         doesn't work correctly)
    ; min_error: numeric lower bound for model-measurement mismatch, or "residual"/"percentile"
@@ -438,12 +444,12 @@ The following file, ``my_hbmcmc_inputs.ini`` can be used to run an
    ; sampler_kwargs (dict): Kwargs to pass to the sampler (e.g. sampler_kwargs = {'target_accept': 0.99})
 
    averaging_error = True
-   min_error = "residual"
+   min_error = 0.0
    fix_basis_outer_regions = False
    use_bc = True
    nuts_sampler = "numpyro"
    save_trace = True
-   min_error_options = {"by_site": True}  ; mapping; only supported key is boolean by_site (residual only)
+   ;min_error_options = {"by_site": True}  ; mapping; only supported key is boolean by_site (residual only)
    pollution_events_from_obs = True
    no_model_error = False
    reparameterise_log_normal = False
@@ -460,6 +466,48 @@ The following file, ``my_hbmcmc_inputs.ini`` can be used to run an
 
    outputpath = '/user/work/ab12345/my_inversions'  ; (required)
    outputname = 'ch4_TAC_test'  ; (required)
+
+Additive-sigma likelihood
+-------------------------
+
+The ``run_hbmcmc.py`` INI route can select a response-independent additive
+Gaussian model-error term with::
+
+   [MCMC.PDF]
+   sigprior = {"pdf": "halfnormal", "sigma": 5.0}
+
+   [MCMC.BC_SPLIT]
+   sigma_freq = "monthly"
+   sigma_per_site = True
+
+   [MCMC.OPTIONS]
+   likelihood = "additive_sigma"
+   min_error = 0.0
+
+This gives the independent likelihood variance
+:math:`s_y^2 + \sigma_{site,period}^2`. It does not use the pollution
+enhancement and does not add aggregation error. ``min_error`` works as before
+as an optional floor on the total standard deviation; it defaults to ``0.0``.
+We recommend first trying the half-normal model with ``min_error = 0.0``.
+``aggregation_error_mode`` is unavailable for this ``run_hbmcmc.py`` option.
+
+A half-normal prior is the recommended starting family. Its ``sigma`` input is
+the half-normal *scale*, and
+
+.. math::
+
+   E[\sigma_{site,period}] = \sigma_{prior}\sqrt{2/\pi},
+   \qquad
+   \sigma_{prior} = E[\sigma_{site,period}]\sqrt{\pi/2}.
+
+Thus the scale which exactly matches a desired prior mean is about 1.25 times
+that mean. If the empirical "average model error" is only a rough upper
+starting point, choosing a somewhat smaller scale gives a correspondingly
+smaller prior mean. This scale is in the same physical units and on the same
+numerical scale as the observations (for example, ppt). It is not the
+dimensionless fractional ``sigma`` used by the pollution-event scaling
+options. Consequently, ``sigprior`` has different units and meaning depending
+on the selected likelihood.
 
 Description of HBMCMC output file
 ---------------------------------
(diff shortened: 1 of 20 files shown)


docs/usage/getting_started.rst
```diff
@@ -379,7 +379,7 @@ The following file, ``my_hbmcmc_inputs.ini`` can be used to run an
 
    xprior   = {"pdf":"lognormal", "stdev":1}  ; lognormal with mean = 1, stdev = 1
    bcprior  = {"pdf":"truncatednormal", "mu":1.0, "sigma":0.02}  ; truncated normal with mean = 1, stdev 0.02
-   sigprior = {"pdf":"uniform", "lower":0.5, "upper":10}
```
brendan-m-murphy, line 382: This old uniform option is very bad... it should be more like lower: 0, upper: 0.2. Also, sigprior has different meaning now depending on the likelihood option, so we need to explain that and probably keep a commented out version around for the multiplicative sigma. (Maybe we add new terms that are less ambiguous like `additive_sigma_prior` and `multiplicative_sigma_prior` and just interpret `sigprior` as one of those, depending on what likelihood is used)

openghg_inversions/models/additive_sigma.py
```diff
@@ -109,9 +112,15 @@ def build_additive_sigma_error(
...
                 "Additive-sigma likelihood requires `sigma_alignment` when model error is enabled."
             )
-    observed = add_model_data(observations.transpose(output_dim), "Y")
-    reported_error = add_model_data(observation_error.transpose(output_dim), "error")
-    minimum_error_data = add_model_data(minimum_error.transpose(output_dim), "min_error")
+        if sigma_prior is None:
+            raise ValueError(
+                "Additive-sigma likelihood requires `sigma_prior` when model error is enabled."
+            )
+    observed = add_model_data(observations.transpose(output_dim), observations_name)
```
brendan-m-murphy: We have an issue open about this: registering the observed data in an "error constructor" is bad. It should be registered in the likelihood. I don't think the additive sigma likelihood needs a separate error constructor

openghg_inversions/models/additive_sigma.py
```diff
@@ -163,14 +168,15 @@ def add_additive_sigma_gaussian_likelihood(
...
     fixed_model_mismatch: xr.DataArray | None = None,
     mean: TensorVariable,
-    sigma_alignment: SigmaAlignment | None,
-    sigma_prior: Mapping[str, Any],
-    no_model_error: bool,
+    sigma_alignment: SigmaAlignment | None = None,
+    sigma_prior: Mapping[str, Any] | None = None,
+    no_model_error: bool = False,
     output_dim: str = "nmeasure",
+    observations_name: str = "Y",
```
brendan-m-murphy: Get rid of this (as per the previous comment)

brendan-m-murphy (commented):
One correctness finding.

openghg_inversions/hbmcmc/run_hbmcmc.py
```diff
@@ -224,6 +238,52 @@ def fixedbasis_params_to_rhime(params: dict[str, Any]) -> dict[str, Any]:
...
+            "run_hbmcmc supports only likelihood='additive_sigma'; omit `likelihood` "
+            "to retain the historical pollution-event likelihood."
+        )
+    if rhime_params.get("aggregation_error_mode", "none") != "none":
+        raise ValueError(
+            "run_hbmcmc additive_sigma does not support `aggregation_error_mode`; "
+            "remove it or set it to 'none'."
+        )
+    rhime_params["aggregation_error_mode"] = "none"
+    options = {
```
brendan-m-murphy, line 268: **[P1] Preserve the sigma-period anchor**

The INI adapter copies `sigma_freq_anchor` only if it already exists in `rhime_params`, but `run_hbmcmc` supplies `start_date`; RHIME derives the anchor from it later. Consequently, fixed-duration frequencies such as `8D` reach the likelihood with `anchor_time=None` and are silently anchored to the first retained observation, potentially changing sigma group assignments after filtering. Forward `start_date` as `sigma_freq_anchor` and add a regression where the first observation occurs after the inversion start.

openghg_inversions/rhime/co2/co2_o2_model.py
```diff
@@ -125,23 +121,14 @@ def _add_co2_o2_fixed_error_likelihood(
...
-    )
-    registered_aggregation_error = add_aggregation_error_data(
-        aggregation_error,
-        observations,
-        output_dim=output_dim,
-    )
-    add_gaussian_observation_likelihood(
-        observed=observed,
+    """Add the reusable fixed-error form of the additive-sigma likelihood."""
+    add_additive_sigma_gaussian_likelihood(
```
brendan-m-murphy: `_add_co2_o2_fixed_error_likelihood` now only renames the observation-error data and forwards to the shared component. Please call `add_additive_sigma_gaussian_likelihood` directly from `build_co2_o2_model` and remove this wrapper. That makes the fixed-error likelihood visibly the no-`sigma_alignment`, no-`minimum_error` specialization of the same likelihood used by the CO2 recipe.

brendan-m-murphy (commented):
One small structural simplification for the CO2/O2 likelihood.

openghg_inversions/rhime/co2/co2_model.py
```diff
@@ -217,9 +220,8 @@ def build_co2_model(
             aggregation_error=aggregation_error,
             fixed_model_mismatch=fixed_mismatch,
             mean=modelled_mean,
-            sigma_alignment=sigma_alignment,
-            sigma_prior=sigma_prior,
-            no_model_error=no_model_error,
+            sigma_alignment=None if no_model_error else sigma_alignment,
```
brendan-m-murphy: The shared likelihood now defines `sigma_alignment=None` as the complete fixed-error switch. Please remove `no_model_error` from `build_co2_model` and resolve the runner's documented debugging flag before this call by passing no alignment (and no sigma prior). That leaves one model-level control instead of two equivalent disabling paths.

brendan-m-murphy (commented):
Two additional CO2-family cleanup points are not on changed diff lines:

- Since these builder APIs are unreleased, please make `build_co2_model` accept `retained_prior: CorrelatedLognormalPrior`, matching `build_co2_o2_model`, and construct it from `alpha_prior_mean` / `alpha_prior_covariance` at the runner or preparation boundary. `parse_prior` creates backend distributions and should not own the coherent reduced mean/covariance.
- In the current embedded-ratio CO2/O2 path, `add_linked_linear_component` receives the unchanged shared state and therefore suggests a transformation that is not present. Please use `apply_linear_sensitivity` directly. OPE-118 owns the later explicit ratio-free/fixed/latent sector-and-species-resolved coupling; it should add an aligned multiplier or transformed O2 state rather than rely on this alias as placeholder scaffolding.

The larger outer-region/grouped-state cleanup will be tracked in a separate Linear issue blocked by this PR.

PR #646, opened 2026-08-25
OPE-119: Simplify grouped CO2 state construction

brendan-m-murphy:
## Summary of changes

- replace the separate CO2 outer-region subsystem with one labelled retained state and ordinary `StateActivity`
- preserve `basis_group`, `basis_partition`, and `region_in_partition` metadata; reconstruct group contributions from the stored full sensitivity and state during reporting
- pass a prepared `CorrelatedLognormalPrior` into the CO2 builder and apply the current CO2/O2 sensitivity directly
- derive offset site coding from the observations' labelled `site` coordinate instead of forwarding a separate builder input
- remove Gaussian outer marginalization, outer-only sector collapse, retired exports/docs, and their mode tests
- add fixed/inferred parity, ordering, cross-covariance, activity-composition, runner-debugging, and reporting-closure evidence

## Dependency

PR #641 has merged. This branch contains its exact final tested head, `51329e808178ad804f1a57f10692ca6f4359c1e5`; no newer #641 commits need to be incorporated.

The OPE-119 diff does not modify or reimplement the shared likelihood, aggregation-error, or observation-error behavior delivered by #641.

## Please check if the PR fulfills these requirements

- [x] Tracked by Linear OPE-119; there is no GitHub issue
- [x] Tests added and passing
- [x] Documentation and tutorials updated
- [x] Added a Towncrier news fragment in `newsfragments/`
- [x] No new requirements

## Validation

- `uv run pytest tests/test_co2_rhime.py tests/test_co2_outer_regions.py tests/test_rhime_co2_o2.py tests/test_rhime_co2_o2_runner.py tests/models/test_state_activity.py` — 69 passed
- `uv run pytest tests/test_preview_docs.py tests/test_grouped_basis_docs.py` — 16 passed
- `uv run pytest tests/models/test_components.py tests/test_co2_rhime.py tests/test_rhime_co2_o2_runner.py` — 41 passed
- Ruff on every changed Python path — passed
- `git diff --check` — passed
- removal search finds no production/public references to `OuterRegionTreatment`, `OuterRegionMode`, `prepare_outer_region_treatment`, `collapse_outer_sectors`, or the outer `"marginalized"` mode

PR #648, opened 2026-08-25
Add labelled fixed-OU mismatch covariance

### LINKED ISSUE #567: Compose fixed aggregation covariance with sampled site OU mismatch
## Goal

Add a reusable RHIME residual-covariance component that composes fixed aggregation covariance with sampled, temporally correlated site mismatch.

The target family is:

`C(sigma) = A + D_obs + blockdiag_s sigma_s^2 T_s(tau)`

where `A` may be dense or low-rank-plus-diagonal, `D_obs` is independent observation variance, and `T_s` is a fixed within-site OU correlation kernel.

## Requirements

- Build OU blocks from labelled site/time observations and a fixed common or site-specific `tau`.
- Align sampled absolute site/period amplitudes through existing sigma preparation.
- Support fixed dense and low-rank-plus-diagonal aggregation covariance.
- Provide normalized log probability and random generation for joint posterior predictive draws.
- Permit non-negative diagonal tails and exact zero modes when positive sampled sigma makes the complete covariance proper.
- Distinguish observation-independent structural-null support from residual-specific improper behaviour at the sigma=0 boundary.
- Persist covariance-component provenance and units.

## Acceptance criteria

- Exact dense and low-rank implementations agree on small fixtures.
- `tau -> 0` agrees with the independent absolute-sigma likelihood.
- Unequal site observation counts and reordered site/time labels are covered.
- Gradient checks cover site amplitudes.
- Predictive draws reproduce the full marginal covariance.
- Existing diagonal absolute-sigma and pollution-event likelihoods remain unchanged.

## Non-goals

- Inferring or selecting `tau` in the first PR.
- Encoding experiment-specific 3 h or 8 h defaults.
- Discrete-tau sampling.

brendan-m-murphy:
* **Summary of changes**

  - Add a labelled fixed-tau, within-site Ornstein–Uhlenbeck mismatch likelihood for
    `A + D_obs + blockdiag_site(sigma_site² T_site(tau))`.
  - Preserve first-occurrence site ordering, exact elapsed-time lags, zero cross-site
    OU covariance, and fixed or inferred per-site amplitudes.
  - Evaluate low-rank aggregation covariance with the verification-games
    generalized-eigen/Woodbury algorithm. The dynamic Cholesky is rank-by-rank;
    the low-rank path never forms an observation-by-observation covariance.
  - Support existing none, diagonal, low-rank-plus-diagonal, and explicitly dense
    aggregation-error inputs; reject singular complete covariance without hidden
    jitter.
  - Expose the component through the standard and multisector RHIME likelihood seam,
    persist component/tau/amplitude provenance and units, and document usage.
  - Keep sampled tau and the accepted-state cached CompoundStep sampler out of scope;
    the latter remains OPE-115.

  **Verification-games parity and performance**

  Frozen WUR25 structure: 4,767 rows, rank 512, 25 sites, ordered-row hash
  `949f69926976e61b2e810354fb35ef6ecd46ad2659f8a934fb5c044da936420d`.

  With GCC loaded and a writable PyTensor compiledir, median warm value+gradient
  evaluation was 0.162349 s in OGI versus 0.139508 s in VG (1.164×). Logp differed
  by 1.46e-11 and the maximum absolute gradient difference was 5.82e-11. Peak RSS
  was 726.3 MiB, with no approximately 173 MiB dense observation covariance.

  **Validation**

  - 83 focused OU and neighboring regression tests passed
  - 11 existing additive-sigma and likelihood-seam cases passed; one provenance
    case needed an immediate retry after an intermittent NetCDF/HDF write error
  - Focused `ruff check` — passed
  - `git diff --check` — passed

- [x] Closes #567
- [x] Tests added and passing

### DIFF
--- a/docs/usage/customising_rhime.rst
+++ b/docs/usage/customising_rhime.rst
@@ -121,6 +121,33 @@ labelled observation ``site`` and ``time`` coordinates. Optional
 independent variance. The prepared ``minimum_error`` retains its historical
 meaning as an optional floor on total standard deviation.
 
+The installed ``rhime.likelihoods.fixed_ou_likelihood_builder`` adds a fixed-
+timescale, within-site Ornstein--Uhlenbeck mismatch covariance. For example::
+
+   from openghg_inversions.rhime import fixed_ou_likelihood_builder, run_rhime
+
+   result = run_rhime(
+       ...,
+       likelihood_builder=fixed_ou_likelihood_builder,
+       likelihood_kwargs={
+           "tau_hours": 5.0,
+           "site_amplitude_prior": {"pdf": "halfnormal", "sigma": 0.75},
+       },
+   )
+
+``tau_hours`` may instead be an exact mapping from retained site labels to
+fixed positive timescales. Pass ``fixed_site_amplitudes`` as a scalar or exact
+site mapping to use known amplitudes instead of the inferred prior. Amplitudes
+have the observation units and tau has units of hours. Observation rows may be
+interleaved or nonmonotonic in time; the component preserves their order and
+sets cross-site OU covariance exactly to zero.
+
+The component applies ``min_error`` to the fixed aggregation-plus-reported-
+error marginal before adding OU variance. It evaluates low-rank aggregation
+covariance with the fixed-OU generalized-eigen and Woodbury method, without
+forming a dense observation covariance. Sampled tau and the cached blocked
+sampler are separate extensions.
+
 Built-in aggregation covariance relies on the guarantees of its construction
 pipeline. A custom pipeline that assembles its own covariance may optionally
 call
--- a/openghg_inversions/models/fixed_ou.py
+++ b/openghg_inversions/models/fixed_ou.py
@@ -1,0 +1,513 @@
+"""Fixed within-site OU covariance under a diagonal-plus-low-rank base."""
+
+from __future__ import annotations
+
+from collections.abc import Mapping, Sequence
+from dataclasses import dataclass, field
+import math
+from typing import Any, cast
+
+import numpy as np
+from numpy.typing import ArrayLike, NDArray
+import pymc as pm
+import pytensor.tensor as pt
+from pytensor.gradient import DisconnectedType, grad_not_implemented
+from pytensor.graph.basic import Apply, Variable
+from pytensor.graph.op import Op
+from pytensor.tensor.variable import TensorVariable
+from scipy.linalg import cho_solve, eigh, solve_triangular
+from scipy.sparse import csr_matrix
+
+
+FloatArray = NDArray[np.float64]
+IntArray = NDArray[np.int64]
+_HOURS_PER_NANOSECOND = 1.0 / 3.6e12
+
+
+@dataclass(frozen=True)
+class FixedOuSiteBlock:
+    """Precomputed generalized eigenbasis for one site's fixed OU template."""
+
+    site: int
+    observation_indices: IntArray
+    mode_slice: slice
+    correlation: FloatArray
+    correlation_cholesky: FloatArray
+    transform: FloatArray
+
+
+@dataclass(frozen=True)
+class FixedOuLikelihoodEvaluation:
+    """One exact likelihood value and its physical-space gradients."""
+
+    log_likelihood: float
+    gradient_residual: FloatArray
+    gradient_site_amplitude: FloatArray
+
+
+def _rejection_gradient_amplitude(amplitude: FloatArray) -> FloatArray:
+    """Return a transform-safe gradient for an invalid scale proposal."""
+    gradient = np.full(amplitude.shape, -1.0, dtype=np.float64)
+    finite_large = np.isfinite(amplitude) & (amplitude > 1.0)
+    gradient[finite_large] = -np.reciprocal(amplitude[finite_large])
+    return gradient
+
+
+class _FixedOuLogpOp(Op):
+    """Instance-local PyTensor bridge to one prepared NumPy/SciPy target."""
+
+    def __init__(self, target: FixedOuLowRank) -> None:
+        self.target = target
+
+    def make_node(self, residual: Any, site_amplitude: Any) -> Apply:
+        residual_variable = pt.as_tensor_variable(residual)
+        amplitude_variable = pt.as_tensor_variable(site_amplitude)
+        if residual_variable.ndim != 1:
+            raise TypeError("residual must be a one-dimensional PyTensor variable.")
+        if amplitude_variable.ndim != 1:
+            raise TypeError("site_amplitude must be a one-dimensional PyTensor variable.")
+        return Apply(
+            self,
+            [residual_variable, amplitude_variable],
+            [pt.dscalar(), pt.dvector(), pt.dvector()],
+        )
+
+    def perform(
+        self,
+        node: Apply,
+        inputs: list[NDArray[Any]],
+        output_storage: list[list[NDArray[Any] | None]],
+    ) -> None:
+        del node
+        evaluation = self.target.evaluate(inputs[0], inputs[1])
+        output_storage[0][0] = np.asarray(evaluation.log_likelihood, dtype=np.float64)
+        output_storage[1][0] = np.asarray(evaluation.gradient_residual, dtype=np.float64)
+        output_storage[2][0] = np.asarray(
+            evaluation.gradient_site_amplitude, dtype=np.float64
+        )
+
+    def L_op(
+        self,
+        inputs: list[Variable],
+        outputs: list[Variable],
+        output_grads: list[Variable],
+    ) -> list[Variable]:
+        """Reuse analytic outputs when differentiating the likelihood value."""
+        if not isinstance(output_grads[1].type, DisconnectedType) or not isinstance(
+            output_grads[2].type, DisconnectedType
+        ):
+            return [
+                grad_not_implemented(self, 0, inputs[0]),
+                grad_not_implemented(self, 1, inputs[1]),
+            ]
+        return [output_grads[0] * outputs[1], output_grads[0] * outputs[2]]
+
+    def infer_shape(
+        self,
+        fgraph: Any,
+        node: Apply,
+        input_shapes: list[tuple[Any, ...]],
+    ) -> list[tuple[Any, ...]]:
+        del fgraph, node
+        return [(), input_shapes[0], input_shapes[1]]
+
+
+@dataclass(frozen=True)
+class FixedOuLowRank:
+    """Exact fixed-OU target with a diagonal-plus-low-rank fixed covariance.
+
+    The eager constructor precomputes one generalized eigenbasis per site.
+    :meth:`logp` then evaluates changing site amplitudes through an ``r x r``
+    Woodbury system, where ``r`` is the aggregation-covariance rank.
+    """
+
+    factor: FloatArray
+    diagonal_variance: FloatArray
+    observation_time_hours: FloatArray
+    site_index: IntArray
+    site_labels: tuple[str, ...]
+    tau_hours_by_site: FloatArray
+    site_blocks: tuple[FixedOuSiteBlock, ...]
+    mode_eigenvalues: FloatArray
+    mode_site_index: IntArray
+    transformed_factor: FloatArray
+    mode_transform: csr_matrix
+    base_logdet: float
+    _logp_op: _FixedOuLogpOp = field(init=False, repr=False, compare=False)
+
+    def __post_init__(self) -> None:
+        object.__setattr__(self, "_logp_op", _FixedOuLogpOp(self))
+
+    @property
+    def n_observation(self) -> int:
+        """Number of observation rows."""
+        return int(self.diagonal_variance.size)
+
+    @property
+    def n_site(self) -> int:
+        """Number of independently scaled sites."""
+        return int(self.tau_hours_by_site.size)
+
+    @property
+    def rank(self) -> int:
+        """Rank of the fixed coherent covariance factor."""
+        return int(self.factor.shape[1])
+
+    def logp(
+        self,
+        value: TensorVariable,
+        mean: TensorVariable,
+        site_amplitude: TensorVariable,
+    ) -> TensorVariable:
+        """Return the normalized Gaussian log density as a differentiable Op."""
+        residual = pt.cast(value - mean, "float64")
+        amplitude = pt.cast(pt.atleast_1d(site_amplitude), "float64")
+        return cast(TensorVariable, self._logp_op(residual, amplitude)[0])
+
+    def evaluate(
+        self,
+        residual: ArrayLike,
+        site_amplitude: ArrayLike | float,
+    ) -> FixedOuLikelihoodEvaluation:
+        """Evaluate the exact log density and analytic physical gradients."""
+        residual_value = np.asarray(residual, dtype=np.float64)
+        amplitude = np.atleast_1d(np.asarray(site_amplitude, dtype=np.float64))
+        if residual_value.shape != (self.n_observation,):
+            raise ValueError(
+                f"residual has shape {residual_value.shape}, "
+                f"expected {(self.n_observation,)}."
+            )
+        if amplitude.shape != (self.n_site,):
+            raise ValueError(
+                f"site_amplitude has shape {amplitude.shape}, expected {(self.n_site,)}."
+            )
+        with np.errstate(over="ignore", invalid="ignore"):
+            amplitude_squared = np.square(amplitude)
+        invalid_amplitude = bool(
+            not np.isfinite(amplitude).all()
+            or np.any(amplitude < 0.0)
+            or not np.isfinite(amplitude_squared).all()
+        )
+        if invalid_amplitude:
+            return FixedOuLikelihoodEvaluation(
+                log_likelihood=-np.inf,
+                gradient_residual=np.zeros(self.n_observation, dtype=np.float64),
+                gradient_site_amplitude=_rejection_gradient_amplitude(amplitude),
+            )
+        if not np.isfinite(residual_value).all():
+            gradient_residual = np.zeros(self.n_observation, dtype=np.float64)
+            infinite = np.isinf(residual_value)
+            gradient_residual[infinite] = -np.sign(residual_value[infinite])
+            return FixedOuLikelihoodEvaluation(
+                log_likelihood=-np.inf,
+                gradient_residual=gradient_residual,
+                gradient_site_amplitude=np.zeros(self.n_site, dtype=np.float64),
+            )
+
+        mode_variance = self.mode_eigenvalues + amplitude_squared[self.mode_site_index]
+        if not np.isfinite(mode_variance).all() or np.any(mode_variance <= 0.0):
+            return FixedOuLikelihoodEvaluation(
+                log_likelihood=-np.inf,
+                gradient_residual=np.zeros(self.n_observation, dtype=np.float64),
+                gradient_site_amplitude=_rejection_gradient_amplitude(amplitude),
+            )
+        weights = np.reciprocal(mode_variance)
+        transformed_residual = cast(FloatArray, self.mode_transform @ residual_value)
+        square_root_weights = np.sqrt(weights)
+        whitened_residual = transformed_residual * square_root_weights
+
+        if self.rank:
+            whitened_factor = self.transformed_factor * square_root_weights[:, None]
+            core = np.eye(self.rank, dtype=np.float64) + (
+                whitened_factor.T @ whitened_factor
+            )
+            core = (core + core.T) * 0.5
+            cholesky = np.linalg.cholesky(core)
+            projected = whitened_factor.T @ whitened_residual
+            latent_mode = cho_solve((cholesky, True), projected, check_finite=False)
+            conditional_whitened = whitened_residual - whitened_factor @ latent_mode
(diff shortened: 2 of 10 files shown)


brendan-m-murphy:
Rebased onto the current PR #646 head (`98c9bf17`). The only conflict was the shared `rhime/likelihoods.py`; the resolution retains #646’s labelled additive-sigma prior behavior alongside the fixed-OU builder.

Post-rebase validation:
- 83 focused OU and neighboring regression tests passed
- 6 existing additive-sigma and likelihood-seam tests passed
- focused Ruff and `git diff --check` passed

Updated OPE-22 head: `83d52087`.

openghg_inversions/rhime/likelihoods.py
```diff
@@ -133,4 +202,174 @@ def additive_sigma_likelihood_builder(
...
+    add_model_data(
+        xr.DataArray(
+            prepared.tau_hours_by_site,
+            dims=("ou_site",),
+            coords={"ou_site": np.asarray(site_labels, dtype=object)},
+            name="ou_tau_hours",
+        )
+    )
+
+    if fixed_site_amplitudes is None:
```
brendan-m-murphy, line 319: **[P1] Reject the improper zero-amplitude posterior.** This inferred branch applies a HalfNormal prior without detecting structurally null rows whose residual is also zero. For a single zero-residual row with no fixed covariance, the likelihood behaves as `1 / amplitude` near zero; because the HalfNormal density remains nonzero there, the posterior is non-integrable and NUTS can drift down the transformed tail until floating-point underflow supplies an artificial boundary. I reproduced an effectively flat transformed logp (`-0.857088` at `log(a)=-5`, `-0.857049` at `-50`). Issue #567 explicitly requires distinguishing structural-null support from residual-specific improper behavior, and Verification Games already has that diagnostic. Please port the boundary check or otherwise reject/remove/regularize this configuration before sampling.

brendan-m-murphy (commented):
One blocking correctness issue found in the inferred-amplitude boundary.

openghg_inversions/rhime/likelihoods.py
```diff
@@ -133,4 +202,174 @@ def additive_sigma_likelihood_builder(
...
-__all__ = ["DEFAULT_ADDITIVE_SIGMA_PRIOR", "additive_sigma_likelihood_builder"]
+def fixed_ou_likelihood_builder(
+    *,
+    observations: xr.DataArray,
+    observation_error: xr.DataArray,
+    minimum_error: xr.DataArray,
+    aggregation_error: AggregationError,
+    mean: TensorVariable,
+    pollution_mean: TensorVariable,
+    pollution_event_baseline: TensorVariable | None,
```
brendan-m-murphy, line 213: **[P1] Rebase onto the likelihood-ownership prerequisite.** A general likelihood should not accept PEFO-only scientific inputs and immediately delete them. [OPE-124](https://linear.app/openghg-inversions/issue/OPE-124/tidy-rhime-likelihood-ownership-before-fixed-ou-lands) removes `pollution_mean` and `pollution_event_baseline` from the common contract, keeps them only on the PEFO branches that consume them, and moves model-context construction under `models`. It also preserves `run_hbmcmc.py` unchanged; `rhime.likelihoods` may remain only as the smallest compatibility import shim if that unchanged file requires it. Please land that prerequisite and rebase rather than adding another implementation to this seam.

openghg_inversions/rhime/likelihoods.py
```diff
@@ -133,4 +202,174 @@ def additive_sigma_likelihood_builder(
...
+            for inferred site amplitudes. Mutually exclusive with fixed values.
+
+    Returns:
+        The observed Gaussian variable named ``y``. The component also creates
+        canonical ``epsilon`` and labelled ``ou_site_amplitude`` variables.
+
+    Raises:
+        ValueError: If labels, configuration, or covariance inputs are invalid.
+    """
+    del pollution_mean, pollution_event_baseline
```
brendan-m-murphy, line 257: **[P1] Put the complete fixed-OU component under `models`.** This function is not a thin adapter: it validates labels, registers PyMC data and coordinates, chooses the minimum-error policy, converts covariance representations, prepares the generalized eigensystem, constructs the amplitude prior, and creates `epsilon` and `pm.CustomDist`. Add a readable `add_fixed_ou_gaussian_likelihood` beside the numerical code in `models.fixed_ou`, in scientific construction order. Nothing in `rhime.likelihoods` should need an active `pm.Model` context.

openghg_inversions/rhime/likelihoods.py
```diff
@@ -133,4 +202,174 @@ def additive_sigma_likelihood_builder(
...
+    observed = add_model_data(observations.transpose(output_dim), "Y")
+    add_model_data(observation_error.transpose(output_dim), "error")
+    add_model_data(minimum_error.transpose(output_dim), "min_error")
+    add_aggregation_error_data(aggregation_error, observations, output_dim=output_dim)
+    add_model_data(site_index.rename("ou_site_index"), "ou_site_index")
+
+    observation_variance = np.square(np.asarray(observation_error.values, dtype=float))
+    minimum_variance = np.square(np.asarray(minimum_error.values, dtype=float))
+    fixed_marginal = observation_variance + aggregation_error.marginal_variance
+    floor_variance = np.maximum(minimum_variance - fixed_marginal, 0.0)
```
brendan-m-murphy, line 300: **[P1] Do not silently add `min_error` to the Verification Games model.** The reviewed VG candidate uses `R = A + D_obs + blockdiag_site(a_site² T_site)` and explicitly adds no positive minimum-error floor. These lines define a different covariance, while the PR summary claims the VG equation. Please remove the floor, its model data, and the builder argument. If a minimum-error OU variant is scientifically wanted later, it needs an explicit option/equation and separate provenance rather than arriving through the generic seam.

openghg_inversions/rhime/likelihoods.py
```diff
@@ -48,6 +62,61 @@ def _resolve_site_sigma_prior(
     return resolved
 
 
+DEFAULT_OU_SITE_AMPLITUDE_PRIOR: dict[str, Any] = {
```
brendan-m-murphy, line 65: **[P1] Make the amplitude prior unit-safe.** `0.75` is the Verification Games CO₂ scale in ppm, but this public builder interprets it in whatever concentration units the retained observations already use; RHIME does not convert those inputs to ppm here. That produces radically different priors for ppm, ppb, or mol/mol data. Require `site_amplitude_prior` explicitly, perform a real unit conversion, or constrain the component to a documented canonical unit. Calling this merely “observation units” does not make the frozen number portable.

openghg_inversions/rhime/likelihoods.py
```diff
@@ -48,6 +62,61 @@ def _resolve_site_sigma_prior(
...
+
+def _aggregation_low_rank_factor(
+    aggregation_error: AggregationError,
+) -> tuple[np.ndarray, np.ndarray]:
+    """Return an exact factor-plus-diagonal representation of aggregation error."""
+    n_observation = aggregation_error.marginal_variance.size
+    if aggregation_error.mode == "dense":
+        assert aggregation_error.covariance is not None
+        covariance = np.asarray(aggregation_error.covariance.values, dtype=float)
+        eigenvalues, eigenvectors = np.linalg.eigh((covariance + covariance.T) * 0.5)
```
brendan-m-murphy, line 79: **[P1] Give `AggregationError` one explicit factor-plus-diagonal conversion.** This is not literal duplication of an existing converter, but it is representation ownership in the RHIME layer. Move it to `observation_error` as an explicit materializing method/function (not a property). The dense branch also accepts eigenvalues within the existing PSD tolerance and then silently drops every non-positive mode while calling the result “exact”; the covariance recorded as model data and the marginal used for the floor can therefore differ from the CustomDist target. Use one PSD/projection policy and document it. Also make clear that dense input generally becomes rank `n`, so this fallback is full-rank/O(n³), not the WUR low-rank performance path.

brendan-m-murphy (commented):
This draft is now blocked on [OPE-124](https://linear.app/openghg-inversions/issue/OPE-124/tidy-rhime-likelihood-ownership-before-fixed-ou-lands), which removes the inherited universal likelihood seam, restores clear model ownership, and preserves the current guarantee that `run_hbmcmc.py` does not change. Please land that prerequisite first and rebase this PR.

After the rebase, this PR still needs the OU-specific cleanup in the inline comments: match or explicitly distinguish the Verification Games covariance, make the prior unit-safe, move the complete PyMC component under `models`, give `AggregationError` one explicit LRPD conversion, document the scientific and numerical model, reject unsupported sampler backends at the configuration boundary, and make provenance explicit.

The intended sampling path in this PR is ordinary configured PyMC NUTS through `RhimeSampler`; users should not have to provide the Verification Games cached `CompoundStep`. OPE-115 owns that later optimization. OPE-114/115/116 do not own the cleanup requested here.

The independent algorithm review found the generalized-eigen/Woodbury value and gradients consistent with the Verification Games derivation, aside from the already-posted improper zero-amplitude boundary and the representation/provenance issues called out below.

docs/usage/customising_rhime.rst
```diff
@@ -121,6 +121,33 @@ labelled observation ``site`` and ``time`` coordinates. Optional
...
+fixed positive timescales. Pass ``fixed_site_amplitudes`` as a scalar or exact
+site mapping to use known amplitudes instead of the inferred prior. Amplitudes
+have the observation units and tau has units of hours. Observation rows may be
+interleaved or nonmonotonic in time; the component preserves their order and
+sets cross-site OU covariance exactly to zero.
+
+The component applies ``min_error`` to the fixed aggregation-plus-reported-
+error marginal before adding OU variance. It evaluates low-rank aggregation
+covariance with the fixed-OU generalized-eigen and Woodbury method, without
+forming a dense observation covariance. Sampled tau and the cached blocked
```
brendan-m-murphy, line 148: **[P1] Add scientific and numerical documentation, not just an invocation.** A user needs the purpose—temporally autocorrelated within-site model–data mismatch—and the full residual model. Please use a conventional residual-covariance symbol such as `R`, define `R = F Fᵀ + diag(d) + ⊕_s a_s² T_s(τ_s)`, state `T_s[i,j] = exp(-|t_i-t_j|/τ_s)`, and explain that the direct sum `⊕` means block-diagonal composition with zero cross-site covariance. Define units and assumptions and give pinned VG provenance. A linked technical page should then derive the LRPD transform, per-site generalized eigensystem, determinant correction, Woodbury core, stable quadratic, analytic gradients, complexity, float64/eager SciPy boundary, duplicate-time restriction, and no-jitter policy.

openghg_inversions/models/fixed_ou.py
```diff
@@ -0,0 +1,513 @@
+"""Fixed within-site OU covariance under a diagonal-plus-low-rank base."""
```
brendan-m-murphy, line 1: **[P2] Replace the stub documentation surface.** This module docstring should introduce the scientific residual model and the implementation strategy. Public dataclasses need attribute shapes, units, invariants, and construction routes; `FixedOuLowRank.logp/evaluate/random` and `prepare_fixed_ou_low_rank` need full Google-style `Args`, `Returns`, `Raises`, and numerical notes; non-obvious boundary helpers such as `_time_hours`, `_group_index`, and `_site_values` need semantic docstrings. Please also add the missing API-reference page/toctree entry and document the instance-local PyTensor `Op`, its gradient outputs/`L_op`, unsupported higher derivatives, and the `CustomDist` signature/random contract.

tests/models/test_fixed_ou_rhime.py
```diff
@@ -0,0 +1,358 @@
...
+            observation_error=data["mf_error"],
+            minimum_error=data["min_error"],
+            aggregation_error=resolve_aggregation_error(data, "low_rank"),
+            mean=mean,
+            pollution_mean=mean,
+            pollution_event_baseline=None,
+            output_dim="nmeasure",
+            tau_hours=5.0,
+        )
+        trace = pm.sample(
```
brendan-m-murphy, line 176: **[P1] Enforce the sampler contract at the configuration boundary.** This proves the desired path: ordinary `pm.sample` selects stock PyMC NUTS and jointly samples the state and site amplitudes, so users do not need a custom step specification. But `RhimeSampler` still advertises `pymc`, `nutpie`, `numpyro`, and `blackjax`, while this SciPy-backed custom Op supplies no JAX/Numba lowering and the component does not reject the external choices. Add runner/config tests that unsupported backends fail early with an actionable message, and document PyMC NUTS as the supported path. OPE-115 can later install the cached CompoundStep without exposing that machinery in INI configuration.

openghg_inversions/rhime/outputs.py
```diff
@@ -45,6 +47,44 @@ class RhimeResult:
     model_build_result: RhimeModelBuildResult | None = None
 
 
+def annotate_likelihood_trace(
+    idata: az.InferenceData,
+    *,
+    builder_identity: dict[str, str],
+    likelihood_kwargs: Mapping[str, Any] | None,
+    concentration_units: str | None,
+) -> None:
```
brendan-m-murphy, line 56: **[P2] Do not infer scientific provenance from a Python function name.** Wrapping, moving, or renaming the builder changes this result, while an unrelated callable with the same suffix is misidentified. The fixed-OU component/model build should provide explicit metadata or variable attributes; generic output handling should serialize that declared metadata rather than recognize one likelihood by `qualname`.

openghg_inversions/models/fixed_ou.py
```diff
@@ -0,0 +1,513 @@
...
+@dataclass(frozen=True)
+class FixedOuSiteBlock:
+    """Precomputed generalized eigenbasis for one site's fixed OU template."""
+
+    site: int
+    observation_indices: IntArray
+    mode_slice: slice
+    correlation: FloatArray
+    correlation_cholesky: FloatArray
+    transform: FloatArray
```
brendan-m-murphy, line 36: **[P3] Remove retained state that the evaluator does not use.** `mode_slice` and `transform` are populated but never read. `correlation` is retained only for the small dense reference method even though it can be reconstructed from `correlation_cholesky`; keeping both stores two full site blocks. Delete the dead fields and retain one representation.

tests/models/test_fixed_ou.py
```diff
@@ -0,0 +1,380 @@
...
+    np.testing.assert_allclose(
+        amplitude_gradient, expected_amplitude, rtol=2e-8, atol=2e-8
+    )
+    np.testing.assert_allclose(evaluation.gradient_residual, expected_value, rtol=2e-8)
+    np.testing.assert_allclose(
+        evaluation.gradient_site_amplitude, expected_amplitude, rtol=2e-8
+    )
+
+
+def test_logp_matches_frozen_verification_games_fixture() -> None:
```
brendan-m-murphy, line 194: **[P2] Pin what “Verification Games parity” means.** A frozen numerical vector without the VG repository commit, source path, input identity, model equation, and deliberate deviations is not reproducible provenance. It also cannot catch builder policies such as `min_error`, prior units, or sampler orchestration because this test starts below those layers. Add a pinned provenance note plus an explicit parity/deviation table, and keep a builder-level parity test for the complete covariance contract.

brendan-m-murphy:
Rebased and updated through `7b3607b4` (current `devel` is an ancestor; GitHub conflict resolved).

Completed focused cleanup:
- moved the active OU construction under `models`, removed the undocumented `min_error` floor, and require an explicit amplitude prior in observation units;
- made aggregation covariance conversion an explicit `observation_error` operation with dense/LRPD/diagonal/none regression coverage;
- made fixed-OU trace provenance declared component metadata rather than function-name inference;
- reject non-PyMC NUTS backends before building a run that uses this SciPy-backed component;
- updated the usage equation/API and ran Ruff plus 49 focused aggregation/OU/RHIME tests.

Remaining draft blockers:
1. The residual-specific improper zero-amplitude posterior diagnostic still needs the reviewed Verification Games boundary policy; the current generic PD guard rejects structurally singular fixed bases but does not encode that residual-dependent policy.
2. Pinned Verification Games provenance/parity must name the authoritative VG commit, source path, and input identity; this repository contains only a frozen numerical fixture, not those identifiers.
3. The requested full numerical/API documentation pass (derivation, Op gradient/random contract, API reference) remains.

No review requested while these P1 items remain.

brendan-m-murphy:
CI follow-up: all three matrix failures at `7b3607b4` were the same regression in `tests/integration/test_cookiecutter_rhime_consumer.py::test_consumer_runs_public_acquisition_to_supported_output`. The test deliberately supplies `setup.sampler = object()`; the early fixed-OU backend check unconditionally accessed `.nuts_sampler`. `8a4b01bd` preserves that consumer seam by defaulting such sampler-like stubs to the ordinary PyMC backend, while explicit incompatible fixed-OU backend choices still fail early.

Local evidence: exact failing test + focused OU/aggregation coverage: 32 passed; complete consumer integration file: 4 passed; Ruff and `git diff --check` passed. Draft status and the existing residual scientific checklist are unchanged.

brendan-m-murphy:
Addressed the three requested functional issues in d7cbf931, with the structural-null precision follow-up in 0c792267:

- fixed-amplitude construction now validates the complete covariance, so a positive OU term may rescue a singular fixed base;
- zero-amplitude evaluation now includes the low-rank factor through the complete-covariance Schur path, including exact structural null modes that LAPACK may otherwise return as tiny positive eigenvalues;
- likelihood provenance now passes custom options through the existing structured-metadata conversion before JSON serialization, so NumPy arrays are supported.

Focused local validation: 58 tests passed across fixed OU, RHIME builder, observation-error, and consumer integration coverage; Ruff and git diff checks passed. The PR remains draft and the existing residual scientific checklist is unchanged.

openghg_inversions/models/fixed_ou.py
```diff
@@ -0,0 +1,704 @@
...
+
+        mode_variance = self.mode_eigenvalues + amplitude_squared[self.mode_site_index]
+        if not np.isfinite(mode_variance).all():
+            return FixedOuLikelihoodEvaluation(
+                log_likelihood=-np.inf,
+                gradient_residual=np.zeros(self.n_observation, dtype=np.float64),
+                gradient_site_amplitude=_rejection_gradient_amplitude(amplitude),
+            )
+        transformed_residual = cast(FloatArray, self.mode_transform @ residual_value)
+        if np.any(mode_variance <= 0.0):
```
brendan-m-murphy: Given the clarified operational invariant that real observations always have positive reported uncertainty, this zero-mode branch is unreachable in real inversions. Positive OU amplitudes already rescue zero-error synthetic cases through the ordinary `mode_variance > 0` path. I recommend deleting this Schur branch and restoring the simple `mode_variance <= 0` rejection, with a clear build-time error for fixed zero amplitudes. The current branch is about 100 lines for an unusual boundary, and its Cholesky test can accept an exactly singular Gram matrix through floating-point roundoff. Add an explicit opt-in or dense fallback only if a real need for factor-rescued zero modes appears.

openghg_inversions/rhime/outputs.py
```diff
@@ -45,8 +47,49 @@ class RhimeResult:
...
+    component_metadata: Mapping[str, str] | None = None,
+) -> None:
+    """Persist likelihood provenance and variable metadata in place.
+
+    Array-valued likelihood options are converted to JSON-compatible values;
+    labelled arrays cross an explicit eager serialization boundary.
+    """
+    idata.attrs["rhime_likelihood_builder"] = json.dumps(builder_identity, sort_keys=True)
+    idata.attrs["rhime_likelihood_kwargs"] = json.dumps(
+        _structured_metadata(dict(likelihood_kwargs or {})), sort_keys=True
```
brendan-m-murphy: `_structured_metadata` fixes numeric ndarrays, but it does not make every array-valued option JSON-compatible. For a `datetime64` array, `.tolist()` yields Python `date`/`datetime` objects and `json.dumps` here still raises `TypeError` after sampling. Please serialize datetime/timedelta values explicitly (for example as ISO strings), or narrow and validate the supported provenance contract before sampling.

brendan-m-murphy (commented):
Re-review at 0c792267. The earlier positive-OU covariance issue and numeric-array provenance issue are addressed. With the clarified invariant that real observations have strictly positive reported uncertainty, exact zero generalized modes are an unusual synthetic boundary; the simplest safe treatment is to reject that boundary rather than maintain a second factorization algorithm. I left inline comments on that simplification and on the remaining datetime-array serialization gap. Validation: 54 focused tests and 20 neighboring likelihood/provenance tests passed; Ruff and git diff checks passed; 795 randomized positive-definite structural-null cases agreed with dense calculations.

openghg_inversions/rhime/outputs.py
brendan-m-murphy: Addressed in 0868be3b. Structured metadata now handles NumPy datetime64/timedelta64 scalars before generic conversion, recursively preserves array shape, and emits JSON-safe strings. The regression covers datetime64[m] and timedelta64[m] arrays; the user documentation records the serialization contract.

openghg_inversions/models/fixed_ou.py
brendan-m-murphy: Addressed in 0868be3b. The zero-mode Schur path has been deleted; evaluate now rejects any non-positive generalized base-plus-OU mode before the low-rank update. Exact structural nullity is still restored after the eigensolve so roundoff cannot bypass that rejection. Fixed-amplitude construction reports the generalized-mode requirement, while the positive-OU rescue case remains covered.

PR #651, opened 2026-08-26
OPE-87 bound multisource sensitivity memory

brendan-m-murphy:
* **Summary of changes**

Project multisource footprint-times-flux sensitivities one source at a time before gathering the state axis. This avoids the state-by-grid temporary array that exhausted memory in the 25-site OPE-87 prepared-input build while preserving labels and numerical results. The legacy float32 comparison now has an explicit 2e-5 absolute tolerance for the changed reduction order (observed maximum delta 1.33e-5).

* **Please check if the PR fulfills these requirements**

- [ ] Closes #xxxx (tracked by Linear OPE-87; no GitHub issue)
- [x] Tests added and passing (`125 passed` across basis function/operator suites)
- [ ] Documentation and tutorials updated/added (no public API change; method docstring updated)
- [x] Added a Towncrier news fragment in `newsfragments/` for user-visible changes
- [ ] Added any new requirements to `requirements.txt` (none)

PR #654, opened 2026-08-26
OPE-121 warn and fuse gathered source-level operations

brendan-m-murphy:
* **Summary of changes**

Closes #653. Implements the useful parts of Linear OPE-121 as a draft PR stacked on #651.

- adds `iter_multi_index_level_slices`, which validates exact unique-level coverage and groups gathered positions in one O(states) pass without broadcasting the dense operand;
- adds a metadata-only warning before `align_to_multi_index_level_values` performs an explicit broadcast when both the projected logical size is at least 1 GiB and the expansion is at least 10x;
- fuses source selection with `MultiSourceBucketBasisOperator.native_prolongation()` and performs one sparse block-diagonal contraction for multisource sensitivity;
- preserves the existing weighted-interpolation path because profiling rejects the proposed source-wise replacement;
- adds missing/extra/duplicate/reordered label regressions, exact grid-order coverage, sparse-contraction structural coverage, and source-independent controls;
- chooses a collision-free temporary source dimension and preserves source-dependent auxiliary coordinates on dimensions that survive the contraction;
- includes reproducible reduced and WUR25/SLURM profiling harnesses.

Related: #653, #651, [Linear OPE-121](https://linear.app/openghg-inversions/issue/OPE-121/profile-multiindex-level-operations-in-multisource-sensitivity-and).

#### Cheap broadcast warning: adopt

The warning uses metadata only:

`projected bytes = da.nbytes / da.sizes[array_dim] * multi_index.sizes[multi_dim]`

It does not compute lazy data. The two-part threshold avoids noise for small arrays and low-ratio reshaping. For the WUR25 sensitivity shape, it predicts the forbidden 168.062 GiB `(state, time, lat, lon)` representation before vectorized selection builds it. The safe iterator is unaffected.

#### Sparse selector/native prolongation: adopt for sensitivity

The source selector is fused with the sparse basis first. The dense input is contracted as `(time, source × grid) @ sparse(source × grid, state)`; no dense `(source, grid, state)` or `(state, grid, time)` block is constructed. Production uses the existing `native_prolongation()` object, with a distinct `native_source` dimension to avoid collision with the state MultiIndex level.

Reduced Dask fixture (float32; four sources; 80×100 grid; 24 times):

| sensitivity | wall | tasks | peak RSS |
|---|---:|---:|---:|
| #651 source-wise loop | 5.57–5.69 s | 189 | 393–398 MiB |
| sparse native contraction | 3.82–4.11 s | 105 | about 381 MiB |

The largest observed native payloads were 0.0049 MiB dense and 0.275 MiB sparse. Coordinates matched the broadcast reference; max float32 difference was `3.0518e-5`, caused by reduction order.

#### Source-wise weighted interpolation: reject

| weighted interpolation | eager wall / peak RSS | Dask wall / peak RSS / tasks |
|---|---:|---:|
| current weight expansion | 0.029 s / 482.1 MiB | 0.933 s / 343.1 MiB / 1,309 |
| proposed source-wise | 0.040 s / 413.7 MiB | 4.623 s / 393.5 MiB / 149 |

The proposed interpolation saved 68.4 MiB eager but was 36% slower; in the production-relevant lazy mode it was 5.0× slower and used 50.4 MiB more peak RSS. Outputs and coordinates were identical under Dask. It is not adopted.

### WUR25 production evidence

Immutable revisions: OGI `3a5234cadf53ab96ec5fd295994c23bd284e9ad6` (stacked on #651 `3c7deceea3006d724b0b9131094da1cdfd690660`); VG `0fe7a2ecec976fcc4bb50d9802f419e9ba0bfc7f`; reduced harness SHA-256 `b786bc75c5a3eb92b0a3c87a30bff3e64f9f47d654877c090cdab32697196c34`; WUR25 harness SHA-256 `09b9b316186c4dee24767a40f40218ec59ae994bbc0c8ce879e40c94084dc59b`.

- PR651 reference gate `18730242`: real `fp_x_flux(4, 293, 391, 743)`, basis `(530, 293, 391)`, 10.652 s and 71 tasks.
- Sparse-native immutable gate `18730624`: completed in 4m42s with scheduler MaxRSS 36,618,292 KiB. The production contraction returned `(530, 743)` in 7.182 s with 36 tasks. Every coordinate, including exact `region` and `time` indexes, matched PR651. Max absolute/relative float32 differences were `1.6689e-5` / `4.3365e-5`. The largest observed payloads were 1.502 MiB dense (the output) and 15.733 MiB sparse.
- Full 25-site build `18730778`: completed in 3m56s with scheduler MaxRSS 37,786,232 KiB. All 25 calls completed in 0.903–6.690 s (mean 1.393 s), using 17–36 tasks. The reopened DataTree contained 5,899 observations, 530 active states, four sources, and 25 sites. Identity, labels/order, units, inner/outer coverage, NESW reconstruction, state activity, and fixed-one closure passed; maximum closure delta was 0.0 ppm at every site.

### Validation

- [x] `137 passed` across `tests/test_array_ops.py` and `tests/test_basis_functions.py`
- [x] Ruff passed on every changed Python path
- [x] `bash -n scripts/slurm_profile_ope121_wur25.sh`
- [x] `git diff --check`
- [x] Towncrier fragment added
- [x] No new dependency

This remains draft while stacked on #651. Repository PR workflows target `main`/`devel`, so the stacked base does not trigger them; retarget after #651 lands.

### DIFF
--- a/openghg_inversions/basis/operators.py
+++ b/openghg_inversions/basis/operators.py
@@ -60,6 +60,7 @@ from openghg_inversions.array_ops import (
     concat_gather_data_arrays,
     force_align,
     get_xr_dummies,
+    iter_multi_index_level_slices,
 )
 from openghg_inversions.basis.layout import (
     BasisStateMetadata,
@@ -1016,22 +1017,30 @@ class MultiSourceBucketBasisOperator(BasisOperator):
         mat_aligned = mat_aligned.transpose(*self.meta.grid_dims, ...)
 
         if self.source_dim in fp_x_flux.dims:
-            state_sources = np.asarray(mat_aligned[self.source_dim].values)
             pieces = []
-            for source in self.source_labels:
-                source_flux = fp_x_flux.sel({self.source_dim: source}, drop=True)
+            positions = []
+            for _, source_positions, source_flux in iter_multi_index_level_slices(
+                fp_x_flux,
+                multi_index=mat_aligned[self.meta.state_dim],
+                multi_dim=self.meta.state_dim,
+                level=self.source_dim,
+                array_dim=self.source_dim,
+            ):
                 if fillna:
                     source_flux = source_flux.fillna(0.0)
+                positions.append(source_positions)
                 pieces.append(
                     xr.dot(
                         source_flux,
-                        mat_aligned.isel(
-                            {self.meta.state_dim: np.flatnonzero(state_sources == source)}
-                        ),
+                        mat_aligned.isel({self.meta.state_dim: source_positions}),
                         dim=list(self.meta.grid_dims),
                     )
                 )
-            h = xr.concat(pieces, dim=self.meta.state_dim).as_numpy()
+            gathered_positions = np.concatenate(positions)
+            h = xr.concat(pieces, dim=self.meta.state_dim)
+            if not np.array_equal(gathered_positions, np.arange(mat_aligned.sizes[self.meta.state_dim])):
+                h = h.isel({self.meta.state_dim: np.argsort(gathered_positions)})
+            h = h.as_numpy()
         elif fillna:
             h = xr.dot(fp_x_flux.fillna(0.0), mat_aligned, dim=list(self.meta.grid_dims)).as_numpy()
         else:
--- a/openghg_inversions/basis/operators.py
+++ b/openghg_inversions/basis/operators.py
@@ -1015,17 +1015,45 @@ class MultiSourceBucketBasisOperator(BasisOperator):
         every retained state while keeping the source pairing sparse.
         """
         if self.source_dim in fp_x_flux.dims:
+            state_coordinate = self.basis_matrix[self.meta.state_dim]
             source_order = [
                 source
                 for source, _, _ in iter_multi_index_level_slices(
                     fp_x_flux,
-                    multi_index=self.basis_matrix[self.meta.state_dim],
+                    multi_index=state_coordinate,
                     multi_dim=self.meta.state_dim,
                     level=self.source_dim,
                     array_dim=self.source_dim,
                 )
             ]
-            native_source_dim = f"native_{self.source_dim}"
+            surviving_source_coordinates = {
+                name: align_to_multi_index_level_values(
+                    coordinate,
+                    multi_index=state_coordinate,
+                    multi_dim=self.meta.state_dim,
+                    level=self.source_dim,
+                    other_dim=self.source_dim,
+                )
+                for name, coordinate in fp_x_flux.coords.items()
+                if name != self.source_dim
+                and self.source_dim in coordinate.dims
+                and not set(coordinate.dims).intersection(self.meta.grid_dims)
+            }
+            occupied_names = {
+                name
+                for name in (
+                    *fp_x_flux.dims,
+                    *fp_x_flux.coords,
+                    *self.basis_matrix.dims,
+                    *self.basis_matrix.coords,
+                )
+            }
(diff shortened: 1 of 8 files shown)


openghg_inversions/basis/operators.py
```diff
@@ -1009,32 +1010,46 @@ def _align_source_like_state(self, other: xr.DataArray) -> xr.DataArray:
...
+                    multi_index=self.basis_matrix[self.meta.state_dim],
+                    multi_dim=self.meta.state_dim,
+                    level=self.source_dim,
+                    array_dim=self.source_dim,
                 )
-            h = xr.concat(pieces, dim=self.meta.state_dim).as_numpy()
-        elif fillna:
-            h = xr.dot(fp_x_flux.fillna(0.0), mat_aligned, dim=list(self.meta.grid_dims)).as_numpy()
+            ]
+            native_source_dim = f"native_{self.source_dim}"
```
brendan-m-murphy: The generated dimension name is not guaranteed to be private. A valid operator configured with `state_dim="native_source"` reaches `native_prolongation()` with the native and retained dimensions sharing a name and fails with an xarray `AlignmentError`; likewise, an input with an otherwise valid extra `native_source` dimension fails here during `rename()`. Since `sensitivity()` documents support for extra input dimensions and `state_dim` is configurable, please choose a collision-free temporary dimension name and cover both cases.

openghg_inversions/basis/operators.py
```diff
@@ -1009,32 +1010,46 @@ def _align_source_like_state(self, other: xr.DataArray) -> xr.DataArray:
...
+                self.basis_matrix,
+                dims=list(self.meta.grid_dims),
+            )
+            if fillna:
+                fp_native = fp_native.fillna(0.0)
+            prolongation = self.native_prolongation(
+                fp_native,
+                native_dims=(native_source_dim, *self.meta.grid_dims),
+            )
+            h = xr.dot(
```
brendan-m-murphy, line 1071: Contracting the newly renamed source dimension silently drops auxiliary coordinates that depend on source plus a surviving dimension. For example, with `calibration(source, time)`, the #651 source-wise path returns `calibration(state, time)`, while this fused path omits `calibration` entirely. That breaks the stated exact-coordinate preservation goal for a valid labelled input. Please map these coordinates onto the gathered state axis before/after contraction, or reject them explicitly rather than losing them.

brendan-m-murphy (commented):
The sparse-native approach resolves the large broadcast and the focused validation is strong. I found two edge-case regressions in the new contraction path: the synthetic native-source dimension is not collision-safe, and source-dependent auxiliary coordinates can be dropped. I left inline comments with reproductions and suggested coverage.

openghg_inversions/basis/operators.py
brendan-m-murphy: Fixed in d2a4c850. The temporary native source dimension is now chosen against every input/basis dimension and coordinate name, with a deterministic numeric suffix on collision. Added regressions for both cases raised here: `state_dim="native_source"` and a valid extra input `native_source` dimension.

openghg_inversions/basis/operators.py
brendan-m-murphy: Fixed in d2a4c850. Source-dependent auxiliary coordinates whose other dimensions survive the spatial contraction are now gathered onto the retained state axis and reattached after the sparse dot. Grid-dependent coordinates remain contracted away, matching PR651. Added an exact `calibration(source, time) -> calibration(state, time)` regression, including index, values, and attrs.

PR #659, opened 2026-08-26
OPE-124 Clarify RHIME likelihood ownership

brendan-m-murphy:
## Summary
- represent built-in RHIME likelihoods with typed, likelihood-specific settings and a small named `ForwardModelTerms` value
- keep standard and multisector recipes readable in scientific order, then dispatch one explicit built-in or custom likelihood
- require ordinary runner calls to select a likelihood explicitly; the shipped config template explicitly selects pollution-event for template users
- materialize `min_error` only for likelihoods which own a minimum-error floor
- preserve the two historical `run_hbmcmc` no-model-error behaviors and frozen additive provenance at the compatibility boundary

## Review response
- `RhimeModelSpec.likelihood` is the serializable selector; concrete recipes have no pollution-event default or callable-identity selection
- `PollutionEventSettings`, `AdditiveSigmaSettings`, and `FixedErrorSettings` own only their valid options and prepared-input requirements
- the generic template leaves `sigma_prior=None`, so changing its selector to additive sigma uses the component's HalfNormal default
- custom likelihoods receive the completed mean contract, use the active PyMC context, and do not require `min_error`
- configured wrappers translate each settings value into explicit direct-component arguments; the dispatcher never receives the whole prepared dataset
- unsupported and no-op option combinations fail during parameter resolution
- no public `rhime.likelihoods`, `rhime.likelihood_seam`, or custom-seam submodule remains; only the frozen legacy provenance string is retained

## Validation
- affected compatibility matrix: **407 passed, 2 deselected**, with one executed known OpenGHG observation-fixture `long_name` mismatch also present on `devel`
- final focused review cases: **10 passed**
- Ruff passes on all changed Python files
- Python compilation and `git diff --check` pass

Linear: https://linear.app/openghg-inversions/issue/OPE-124/tidy-rhime-likelihood-ownership-before-fixed-ou-lands

Closes OPE-124


openghg_inversions/models/additive_sigma.py
```diff
@@ -123,14 +129,20 @@ def add_additive_sigma_gaussian_likelihood(
             FIXED_MODEL_MISMATCH,
         )
         independent_variance = independent_variance + fixed_mismatch_data**2
-    if sigma_alignment is not None:
-        assert sigma_prior is not None
-        sigma = add_sigma_component(sigma_alignment, prior_args=dict(sigma_prior))
-        independent_variance = independent_variance + sigma**2
+    if additive_scale_alignment is not None:
```
brendan-m-murphy: Why not use `additive_sigma_alignment`? "scale" implies "scaling" to me, which is the alternative to "additive" in this context

openghg_inversions/models/additive_sigma.py
```diff
@@ -123,14 +129,20 @@ def add_additive_sigma_gaussian_likelihood(
...
+        assert additive_scale_prior is not None
```
brendan-m-murphy: Add message after assert?

openghg_inversions/rhime/_custom_likelihood_seam.py
```diff
@@ -0,0 +1,80 @@
...
+        )
+    return likelihood
+
+
+def validate_likelihood_kwargs(
+    likelihood_builder: object | None,
+    likelihood_kwargs: object | None,
+) -> dict[str, Any] | None:
+    """Copy and validate options owned by a custom likelihood."""
+    if likelihood_builder is not None and not callable(likelihood_builder):
```
brendan-m-murphy: Just make the type in the signature `Callable | None`; we're not doing runtime type checking.

openghg_inversions/rhime/_custom_likelihood_seam.py
```diff
@@ -0,0 +1,80 @@
...
+        raise TypeError(
+            f"`likelihood_builder` must be callable or None; got {type(likelihood_builder).__name__}."
+        )
+    if likelihood_kwargs is None:
+        return None
+    if not isinstance(likelihood_kwargs, Mapping):
```
brendan-m-murphy: Again, just use `Mapping[str, Any]` in the signature

openghg_inversions/rhime/_custom_likelihood_seam.py
```diff
@@ -0,0 +1,80 @@
...
+    missing_names = sorted({"y", "epsilon"} - set(model.named_vars))
+    if missing_names:
+        raise ValueError(
+            "A RHIME likelihood builder did not create the canonical variables required by "
+            f"sampling and outputs: {missing_names!r}."
+        )
+    return likelihood
+
+
+def validate_likelihood_kwargs(
```
brendan-m-murphy: This public function needs a full google docstring... if we're keeping it. It looks way over-engineered. It is essentially checking that the kwargs can be JSON round tripped? I would kind of expect this function to inspect the builder signature and determine if the kwargs can be used. (`openghg` uses its `split_function_inputs` helper to do this, and you could warn about unused inputs.)

openghg_inversions/rhime/_custom_likelihood_seam.py
```diff
@@ -0,0 +1,80 @@
...
+        observation_error: xr.DataArray,
+        aggregation_error: AggregationError,
+        mean: TensorVariable,
+        output_dim: str,
+    ) -> TensorVariable:
+        """Add canonical ``y`` and ``epsilon`` variables to the active model."""
+        ...
+
+
+def validate_custom_likelihood_result(
```
brendan-m-murphy: This seems like a reasonable helper but it is not listed under `__all__`, and is "public" in a "private" submodule, so I'm unclear if this is part of the API or not. It should have a full google-style docstring if it is.

openghg_inversions/rhime/_model_building.py
```diff
@@ -17,41 +22,67 @@
...
+    observations: xr.DataArray,
+    *,
+    output_dim: str,
+    site_indicator: xr.DataArray | None = None,
+    sigma_prior: Mapping[str, Any] | None = None,
+    sigma_freq: str | None = None,
+    sigma_per_site: bool = True,
+    sigma_freq_anchor: DatetimeLike | None = None,
+    no_model_error: bool = False,
+) -> tuple[SigmaAlignment | None, dict[str, Any] | None]:
```
brendan-m-murphy: use the `PriorArgs` type alias instead of `dict`

openghg_inversions/rhime/multisector.py
```diff
@@ -321,7 +338,9 @@ def build_multisector_rhime_model(
             either shared-basis or gathered source-specific state layout.
         observations: Observed mole fractions.
         observation_error: Reported observation-error standard deviations.
-        minimum_error: Minimum total-error standard deviations.
+        minimum_error: Optional minimum total-error standard deviations. The
+            pollution-event mismatch requires it; additive sigma uses it only
+            when the resolved model spec opts into the historical floor.
```
brendan-m-murphy: What does this mean? We have moved away from the model spec emphasis. I don't think mentioning it adds anything for the user, since this function does accept a "model spec"

openghg_inversions/rhime/multisector.py
```diff
@@ -355,13 +379,16 @@ def build_multisector_rhime_model(
         TypeError: If a custom likelihood returns the wrong result type.
     """
     likelihood_kwargs = validate_likelihood_kwargs(likelihood_builder, likelihood_kwargs)
+    if mismatch_model not in {"pollution_event", "additive_sigma"}:
+        raise ValueError(f"Unsupported built-in mismatch model {mismatch_model!r}.")
+    if likelihood_builder is not None and mismatch_model != "pollution_event":
```
brendan-m-murphy: This check is using the default value for mismatch_model as if it were `None`. What if the user supplies `pollution_event` explicitly? That is the same error as supplying `additive_sigma` explicitly, but we do not flag it.

openghg_inversions/rhime/multisector.py
```diff
@@ -306,6 +320,9 @@ def build_multisector_rhime_model(
     power: PriorArgs | float = 1.99,
     state_activity: StateActivity | None = None,
     bc_state_activity: StateActivity | None = None,
+    mismatch_model: MismatchModel = "pollution_event",
```
brendan-m-murphy: We're falling back into the "many option arguments + many flags/branches" trap of `fixedbasisMCMC`.

Scientifically, the model-building code reads fine if you use a generic likelihood builder in place of the "default" pollution-event likelihood. In fact, this is probably preferable to all the branch and argument wrangling. I think it is reasonable to say we're supplying a likelihood builder and its arguments, and then building the likelihood in the model, passing it the inputs from the forward model; separate, here is the list of likelihoods you can use, and the inputs they accept.

To some extent the issue is the pollution-event likelihoods requiring different parts of the forward model, and we have no canonical object that describes the forward model (the prototype model-configs branch had such an object).

docs/plans/issue_528_semantic_model_adr.md
```diff
@@ -1076,8 +1076,8 @@ linked-observation-model implementations.
 - `openghg_inversions/basis/prior_uncertainty.py` documents and implements the
   independent-grid-cell special case.
 - `openghg_inversions/observation_error.py` and
-  `openghg_inversions/models/likelihoods.py` contain the current fixed dense,
-  diagonal, and LRPD covariance consumers.
+  `openghg_inversions/models/_gaussian_observation.py` contain the current fixed
```
brendan-m-murphy: This ADR was withdrawn, you should not be referencing it.

brendan-m-murphy (commented):
I don't want a mess of parameter options and branching code in the model builders. Refactor this as I suggest.

The only invariant here is that the behaviour of `run_hbmcmc.py` does not change (I suppose it could be extended if it does not invalidate existing configs, slurm scripts, and outputs)

brendan-m-murphy:
Implemented the latest simplification review in `b81335a4`.

The concrete standard and multisector recipes now have one generic likelihood call and no pollution-event/additive selection branch. Runner/config resolution owns the backwards-compatible pollution-event choice, while `RhimeModelSpec` itself defaults to no mismatch selection. `min_error` is selected only by a built-in component that needs it, and custom prepared runs are covered without that input.

I also removed the custom seam/validator module, fixed the custom-runner examples for explicit config defaults, preserved `run_hbmcmc` construction and saved provenance, and added standard/multisector collision and no-floor coverage.

Validation: 421 affected tests passed (3 known unrelated tests deselected), Ruff passed on all changed Python files, and `git diff --check` passed. I have left the review threads open for reviewer verification.

brendan-m-murphy:
Implemented the recommended likelihood-settings refactor in `0c01a498`.

Key changes:
- `RhimeModelSpec` now carries a typed `PollutionEventSettings | AdditiveSigmaSettings | None` value instead of flat likelihood fields.
- Standard and multisector recipes expose `ForwardModelTerms(total, pollution, baseline)` and share one explicit built-in dispatcher.
- Sigma alignment is derived from `mf.coords["site"]`; likelihoods no longer select or materialize `site_indicator`.
- Prepared-input requirements come from the selected settings, so custom and unfloored additive likelihoods do not require `min_error`.
- Likelihood-specific configuration is rejected during option resolution (for example additive sigma + `power`).
- The runner/config boundary still defaults ordinary runs to pollution-event settings; concrete recipes have no implicit PEFO default.
- The private legacy boundary-only PEFO baseline and frozen `run_hbmcmc` provenance/graph behavior are preserved.

Validation:
- 447 broader RHIME/CO2/prepared/compatibility tests passed with 3 known environment-sensitive tests deselected.
- Four float32 stochastic tolerance flakes from that combined run all passed on immediate focused rerun.
- Ruff passed on every changed Python path.
- `git diff --check` passed.

openghg_inversions/rhime/params.py
```diff
@@ -695,20 +753,23 @@ def make_rhime_runner_setup(
...
     if multisector:
         validate_multisector_x_prior(x_prior)
     offset_args = normalise_optional_mapping(remaining.get("offset_args"))
 
     use_bc = remaining.get("use_bc", True)
-    sigma_per_site = remaining.get("sigma_per_site", True)
-    sigma_freq = remaining.pop("sigma_freq", None)
+    mismatch_model = cast(
+        MismatchModel | None,
+        remaining.pop("mismatch_model", "pollution_event"),
```
brendan-m-murphy: Please remove the implicit `"pollution_event"` fallback for ordinary RHIME. Omission should fail or otherwise require an explicit modern likelihood selection; `run_hbmcmc.py` already injects `"pollution_event"` for its compatibility path, so preserving legacy behaviour does not require making it the new runner default.

openghg_inversions/config/templates/rhime_template.ini
```diff
@@ -69,6 +69,8 @@ chains = 4
 averaging_error = True
 min_error = 0.0
 fix_basis_outer_regions = False
+; Built-in mismatch component. Use "additive_sigma" for absolute mismatch variance.
+mismatch_model = "pollution_event"
```
brendan-m-murphy: Switching this line to `"additive_sigma"` still carries the global uniform `sigma_prior` from line 53 into the additive model, bypassing its HalfNormal default. Please make the generic template prior neutral (`None`) or otherwise ensure the selected likelihood owns its default, so this one-line switch is scientifically complete. The wider namespaced-config refactor can remain a follow-up.

openghg_inversions/rhime/_model_building.py
```diff
@@ -48,12 +144,131 @@ def validate_custom_likelihood_result(
...
+
+    if isinstance(settings, AdditiveSigmaSettings):
+        alignment, prior = prepare_additive_sigma_inputs(
+            observations,
+            sigma_alignment=sigma_alignment,
+            sigma_prior=settings.sigma_prior,
+            sigma_freq=settings.sigma_freq,
+            sigma_per_site=settings.sigma_per_site,
+            sigma_freq_anchor=settings.sigma_freq_anchor,
+            no_model_error=settings.no_model_error,
```
brendan-m-murphy: Please make the absence of inferred model error a separate built-in likelihood/settings type (for example `FixedErrorSettings`) instead of a flag carried by every likelihood. Otherwise sigma-specific options become accepted no-ops. Keep the two historical `run_hbmcmc` cases in its compatibility adapter/tests: pollution-event mode retains an unused `sigma` and ignores `min_error`, while additive mode omits `sigma` but retains its minimum-error floor.

openghg_inversions/rhime/standard.py
```diff
@@ -279,39 +261,27 @@ def build_standard_rhime_model(
...
-            )
-            validate_custom_likelihood_result(model, likelihood)
 
+        forward = ForwardModelTerms(
+            total=modelled_mean,
+            pollution=pollution_mean,
+            baseline=baseline_mean,
+        )
+        add_rhime_likelihood(
+            model,
```
brendan-m-murphy: Please drop the explicit `model` argument. These component calls already run inside `registered_model()`, and the only consumer is the custom-likelihood `named_vars` validation; that can obtain the active model with `pm.modelcontext(None)`. This keeps `add_*` components consistent with the rest of the PyMC-context-based API. Please make the equivalent multisector change.

openghg_inversions/rhime/_model_building.py
```diff
@@ -48,12 +144,131 @@ def validate_custom_likelihood_result(
...
 
 
+def add_builtin_likelihood(
+    settings: LikelihoodSettings,
+    *,
+    forward: ForwardModelTerms,
+    observations: xr.DataArray,
+    observation_error: xr.DataArray,
+    aggregation_error: AggregationError,
+    model_inputs: xr.Dataset,
```
brendan-m-murphy: Please avoid passing the entire materialized `model_inputs` dataset through the generic likelihood dispatcher merely to extract `min_error`. Pass `minimum_error` explicitly. A small, distinctly named configured wrapper per likelihood (for example `add_configured_additive_sigma_likelihood(settings, ...)`) can own settings translation while leaving the direct PyMC component explicit; no overload or wider component framework is needed.

brendan-m-murphy (commented):
I’ve kept this review bounded to OPE-124: the inline comments cover the changes needed for a truthful, local likelihood-selection seam, without requiring the wider config-parser refactor here.

Please also rebase onto current `devel` before the next pass. This head is based on `92d4448a` and is missing the East Asia expansion and Central Asia mappings added in `96d16429`.

brendan-m-murphy:
Addressed the five active review threads in `8ea6eb42` after rebasing onto current `devel`:

- removed ordinary RHIME's implicit pollution-event fallback
- made the template sigma prior neutral
- introduced `FixedErrorSettings` and kept both historical `run_hbmcmc` no-model-error behaviors in the compatibility adapter
- removed explicit PyMC model arguments from likelihood construction
- replaced whole-dataset likelihood forwarding with explicit `minimum_error`

Validation: 407 affected tests passed; the only executed failure is the known OpenGHG fixture `long_name` drift also present on `devel`. Final focused review cases pass 10/10, and Ruff, Python compilation, and `git diff --check` pass.

openghg_inversions/config/templates/rhime_template.ini
```diff
@@ -50,7 +50,8 @@ x_prior = {"pdf": "lognormal", "mean": 1.0, "stdev": 1.0}
 ; Leave this as None to apply the shared x_prior to every sector.
 sector_priors = None
 bc_prior = {"pdf": "truncatednormal", "mu": 1.0, "sigma": 0.05, "lower": 0.0}
-sigma_prior = {"pdf": "uniform", "lower": 0.1, "upper": 3.0}
+; Leave neutral so each selected likelihood applies its own scientific default.
+sigma_prior = None
```
brendan-m-murphy: This probably needs a comment to explain the options, or explain where you can find the options.

